PR #97, opened 2023-05-17
React to container chain assignment

tmpolaczyk:
This PR implements a mechanism that allows collators to automatically detect changes to their assignment.

Before this PR:

* The container chain is selected using the `--para-id` CLI argument. Starting the node without the extra CLI arguments results in a Tanssi collator.
* The container chain starts immediately
* The container chain never stops
* There is only one tanssi client and one container chain client

After this PR:

* The container chain is selected automatically by using a runtime API. We still need the extra CLI arguments to start the `ContainerChainSpawner`, but we don't need to know the para id in advance. Starting the node without the extra CLI arguments results in a Tanssi collator, same as before. But if a node started with the extra CLI arguments is assigned to collate on tanssi, it will do so, and it will rotate to a container chain if it is assigned to one in the future.
* The container chain starts after importing the first Tanssi block that indicates that assignment. This introduces a small delay in zombienet tests.
* The node reacts to new assignments: if the node is assigned to tanssi, it will stop the previous container chain. And if the node is assigned to a new container chain, it will start a new client for that container chain.
* There is only one tanssi client, but there can be multiple container chain clients active at a time. Only the most recently started container chain will collate. The tanssi client can restart collation at any time. The websockets RPC port is only usable for the first container chain, as any new ones will use a different port, and there is a bug in cumulus related to CORS that prevents us to connect to the non-default port (actually if the first container chain stops before the second container chain starts, it can reuse the port)


Cargo.toml
```diff
@@ -118,33 +118,34 @@ substrate-wasm-builder = { git = "https://github.com/paritytech/substrate", bran
 try-runtime-cli = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40" }
 
 # Polkadot (wasm)
-polkadot-parachain = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.40", default-features = false }
-polkadot-runtime-common = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.40", default-features = false }
+polkadot-parachain = { git = "https://github.com/tmpolaczyk/polkadot", branch = "tomasz-reinitialize-9-40", default-features = false }
+polkadot-runtime-common = { git = "https://github.com/tmpolaczyk/polkadot", branch = "tomasz-reinitialize-9-40", default-features = false }
```
girazoki, line 122: Can you use our fork? https://github.com/moondance-labs/polkadot. You can put polkadot-v0.9.40 as the branch or tanssi-polkadot-v0.9.40

node/src/container_chain_spawner.rs
```diff
@@ -0,0 +1,276 @@
...
+
+            let mut container_chain_cli_config = sc_cli::SubstrateCli::create_configuration(
+                &container_chain_cli,
+                &container_chain_cli,
+                tokio_handle.clone(),
+            )
+            .map_err(|err| format!("Container chain argument error: {}", err))?;
+
+            // Change database path to make it depend on container chain para id
+            // So instead of the usual "db/full" we have "db/full-container-2000"
```
girazoki, line 156: Love this

node/src/command.rs
```diff
@@ -380,6 +382,7 @@ pub fn run() -> Result<()> {
                 )),
                 #[cfg(feature = "runtime-benchmarks")]
                 BenchmarkCmd::Storage(cmd) => runner.sync_run(|config| {
+                    // TODO: should this be new_partial or new_partial_orchestrator?
```
girazoki, line 385: New partial orchestrator

node/src/container_chain_spawner.rs
```diff
@@ -0,0 +1,277 @@
...
+        // This closure is used to emulate a try block, it enables using the `?` operator inside
+        let try_closure = move || async move {
+            // Preload genesis data from orchestrator chain storage.
+            // The preload must finish before calling create_configuration, so any async operations
+            // need to be awaited.
+
+            // TODO: the orchestrator chain node may not be fully synced yet,
+            // in that case we will be reading an old state.
+            let orchestrator_chain_info = orchestrator_client.chain_info();
+            log::info!(
```
girazoki, line 106: This is an important TODO, we should add an issue for this

node/src/service.rs
```diff
@@ -298,7 +457,15 @@ async fn start_node_impl(
         .map(|e| e.relay_chain.clone())
         .ok_or_else(|| "Could not find relay_chain extension in chain-spec.")?;
 
-    let params = new_partial(&parachain_config)?;
+    // Channel to send messages to start/stop container chains
+    let (cc_spawn_tx, cc_spawn_rx) = unbounded_channel();
+    // If this node was started without a `container_chain_config`, we don't support collation
+    // on container chains, so there is no need to detect changes to assignment
+    let params = if container_chain_config.is_some() {
```
girazoki, line 464: In this case, what happens if I start as a Tanssi-collator, then I get moved to a container-chain? what happens if I start without the container_chain_config, because I am a tanssi-collator, but then I get re-assigned?

tmpolaczyk: If you start without the `container_chain_config`, the collator will always try to collate on Tanssi.

It shouldn't be too hard to fix it (we just need to create a default config if it's missing, or even panic and ask the user to provide it), but I would prefer to leave it for a future PR.

For now if you want I can start all the collators with a `container_chain_config`, by modifying the args to `-- --tmp`.

girazoki: I see perfect. Yeah I agree we can leave it for future PR, let's also open an issue for this. For me it should be that:
- If parachain_config.is_authority() is true, then we should always spawn new_partial_orchestrator, because you might get re-assigned just because you are a collator

test/suites/para/test_tanssi_containers.ts
```diff
@@ -159,5 +167,165 @@ describeSuite({
...
+        const header2002 = await getHeaderFromRelay(relayApi, 2002);
+        expect(header2002.number.toNumber()).to.be.equal(0);
+        const registered1 = (await paraApi.query.registrar.registeredParaIds());
+        expect(registered1.toJSON().includes(2002)).to.be.false;
+
+        const tx = paraApi.tx.registrar.register(2002, containerChainGenesisDataFromRpc[1]);
+        await paraApi.tx.sudo.sudo(tx).signAndSend(alice);
+        
+        const tanssiBlockNum = (await paraApi.rpc.chain.getBlock()).block.header.number.toNumber();
+        // TODO: this should wait 2 sessions. We are waiting 10 + 20 blocks
```
girazoki, line 240: Should not this change after we reduce it to 5?

test/suites/para/test_tanssi_containers.ts
```diff
@@ -159,5 +167,165 @@ describeSuite({
...
+        const container2002Network = container2002Api.consts.system.version.specName.toString();
+        const paraId2002 = (await container2002Api.query.parachainInfo.parachainId()).toString();
+        expect(container2002Network, "Container2002 API incorrect").to.contain("container-chain-template");
+        expect(paraId2002, "Container2002 API incorrect").to.be.equal("2002");
+
+        // Check authors of tanssi blocks
+        // Should be 2 different keys because 2002 has been registered
+        await countUniqueBlockAuthors(context, paraApi, 4, 2);
+
+        let blockNum = (await container2002Api.rpc.chain.getBlock()).block.header.number.toNumber();
```
girazoki, line 262: Isnt this essentially the same as waitForBlock?

test/suites/para/test_tanssi_containers.ts
tmpolaczyk: Yes, setting it to 5 didn't work at first because the transaction took a few blocks more to get included, but I am working on a fix for that.

test/suites/para/test_tanssi_containers.ts
tmpolaczyk: Ideally yes, but:

    // Cannot use context.waitBlock because the container2002Api is not part of moonwall

I need to see if it is possible to add a network to the moonwall context while running a test, or some other workaround.

test/suites/para/test_tanssi_containers.ts
```diff
@@ -177,5 +180,169 @@ describeSuite({
...
+        expect(header2002.number.toNumber()).to.be.equal(0);
+        const registered1 = (await paraApi.query.registrar.registeredParaIds());
+        expect(registered1.toJSON().includes(2002)).to.be.false;
+
+        const tx = paraApi.tx.registrar.register(2002, containerChainGenesisDataFromRpc[1]);
+        await signAndSendAndInclude(paraApi.tx.sudo.sudo(tx), alice);
+        const session1 = (await paraApi.query.session.currentIndex()).toNumber();
+        await jumpSessions(context, 2);
+        const session2 = (await paraApi.query.session.currentIndex()).toNumber();
+        console.log("Session before/after: ", session1, "/", session2);
```
tmpolaczyk, line 255: Now the tests are failing because in CI `await jumpSessions(context, 2);` only waits for 1 session. Locally it is working as expected (prints 1/3), not sure why CI prints (1/2).

node/src/service.rs
```diff
@@ -195,6 +199,161 @@ pub fn new_partial(
...
+    initial_assigned_para_id: Arc<Mutex<Option<ParaId>>>,
+    client_set_aside_for_cidp: Arc<Client>,
+    block_hash: H256,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>>
+where
+    Client: ProvideRuntimeApi<Block>,
+    <Client as ProvideRuntimeApi<Block>>::Api: CollatorAssignmentApi<Block, AccountId, ParaId>,
+{
+    let nimbus_keys = SyncCryptoStore::keys(&*sync_keystore, NIMBUS_KEY_ID).unwrap();
+    let collator_account_id =
```
girazoki, line 328: This should not convert to collator_account, it should **always** use the nimbus keys to detect the assignment. Collator accountIds can be very different to their nimbus keys. We should check the assignation with the nimbus Id

test/suites/para/test_tanssi_containers.ts
```diff
@@ -177,5 +183,224 @@ describeSuite({
...
+        let blockNumber = sessionPeriod;
+        // Before 2002 registration: 4 authors
+        await countUniqueBlockAuthors(paraApi, blockNumber, blockNumber2002Start-1, 4);
+
+        // While 2002 is live: 2 authors (the other 2 went to container chain 2002)
+        // FIXME: there is a delay between a node detecting a change in assignment, and changing the
+        // collation para id, so at the beginning nodes that should be creating blocks in 2002 may
+        // still create some blocks for the orchestrator chain. This will be fixed when we implement
+        // validation, once that works remove this delay.
+        //let bugDelay = sessionPeriod;
```
tmpolaczyk, line 364: Remove this comment, no longer applies

node/src/rpc.rs
```diff
@@ -90,3 +95,34 @@ where
...
 }
+
+/// Utils API implementation.
+pub struct Utils {}
+
+impl Utils {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
```
nanocryk, line 106:
```suggestion
pub struct Utils;
```

node/src/service.rs
```diff
@@ -195,6 +198,86 @@ pub fn new_partial(
...
+/// and start/stop container chains on demand. The check runs on every new block.
+pub fn build_check_assigned_para_id(
+    client: Arc<ParachainClient>,
+    sync_keystore: SyncCryptoStorePtr,
+    cc_spawn_tx: UnboundedSender<CcSpawnMsg>,
+    spawner: impl SpawnEssentialNamed,
+) {
+    // Subscribe to new blocks in order to react to para id assignment
+    let mut import_notifications = client.import_notification_stream();
+    let client_set_aside_for_cidp = client.clone();
```
nanocryk, line 211: Why clone `client` as it is not used in this scope?

nanocryk (approved):
LGTM outside of small nitpicking :p

PR #103, opened 2023-05-23
Modify development_config for pallet-configuration

Agusrodri:
This PR changes the default dev_config of the orchestrator chain (Tanssi), for Alice to be the only one capable of producing blocks.

### Changes
* Changed `configuration` parameter in `testnet_genesis()` function within `chain_spec` file. Instead of default, now we use a custom `HostConfiguration` to allow only 1 collator in the orchestrator chain.


girazoki:
Hi @Agusrodri, thanks for the PR. I was more thinking on just changing the development_config chain-spec [development_config](https://github.com/moondance-labs/tanssi/blob/5733249cbb9560be25e0fb3276d6d103f76fdb07/node/src/chain_spec.rs#LL146C8-L146C26) to actually match what we expect for development. Right now we are passing in all cases the default (with which we were happy having a 2 by default):

https://github.com/moondance-labs/tanssi/blob/5733249cbb9560be25e0fb3276d6d103f76fdb07/node/src/chain_spec.rs#L302

test/suites/dev/session-keys/test_session_keys.ts
```diff
@@ -33,7 +33,6 @@ describeSuite({
             // Check authorities are correct
             const authorities = (await polkadotJs.query.aura.authorities());
             expect(u8aToHex(authorities[0])).to.be.eq(u8aToHex(alice.publicKey));
```
tmpolaczyk, line 35: I recently made a PR to use to.deep.equal in cases like these, did I miss this line or did you forget to merge master?

```suggestion
            expect(authorities.toJSON()).to.deep.equal([u8aToHex(alice.publicKey)]);
```

tmpolaczyk (approved):

girazoki (approved):

PR #106, opened 2023-05-26
evm client non tracing

girazoki:
It adds all necessary components for the frontier-template node to be able to support basic ethereum support, e.g., ethers metamask and ganache. Tracing is not included in this PR, and will come in a separate PR.

Most of the configuration comes from Moonbeam, although it has been tailored to be a bit simpler. We should discuss in the future whether we should try to gather most of the primitives from there.

I added some basic tests to the template in the para-tests, but I think in the future we should implement manual-seal to this template, and test more thoroughly. I am expecting this to be a separate PR though


test/util/ethereum.ts
```diff
@@ -0,0 +1,214 @@
+
+
+import { Signer, ethers } from "ethers";
```
girazoki, line 3: This should probably go away once we have a way to do in from Moonwall directly

container-chains/templates/frontier/node/src/service.rs
```diff
@@ -35,42 +38,78 @@ use {
...
+        .as_ref()
+        .map(|base_path| base_path.config_dir(config.chain_spec.id()))
+        .unwrap_or_else(|| {
+            BasePath::from_project("", "", "moonbeam").config_dir(config.chain_spec.id())
+        });
+    config_dir.join("frontier").join(path)
+}
+
+// TODO This is copied from frontier. It should be imported instead after
+// https://github.com/paritytech/frontier/issues/333 is solved
```
tmpolaczyk, line 75: That issue is still open but now that function is publicly available as `fc_db::Backend::open()`, so you can do something like

```rust
pub fn open_frontier_backend<C>(
    client: Arc<C>,
    config: &Configuration,
) -> Result<Arc<fc_db::Backend<Block>>, String>
where
    C: sp_blockchain::HeaderBackend<Block>,
{
    let config_dir = config
        .base_path
        .as_ref()
        .map(|base_path| base_path.config_dir(config.chain_spec.id()))
        .unwrap_or_else(|| {
            BasePath::from_project("", "", "moonbeam").config_dir(config.chain_spec.id())
        });
    let backend = fc_db::Backend::open(client.clone(), &config.database, &config_dir)?;

    Ok(Arc::new(backend))
}
```

test/configs/zombieTanssi.json
```diff
@@ -83,7 +83,8 @@
         "collators":  [
           {
             "name": "DummyCollator-2001",
-            "command": "../target/release/container-chain-template-frontier-node"
+            "command": "../target/release/container-chain-template-frontier-node",
+            "ws_port": 9951
```
tmpolaczyk, line 87: I used the same port number in my PR, could you change it?

girazoki: OFC

test/suites/para/test_tanssi_containers.ts
```diff
@@ -177,5 +190,21 @@ describeSuite({
...
+      title: "Transactions can be made with ethers",
+      timeout: 30000,
+      test: async function () {
+        const randomAccount = generateKeyringPair();
+        let tx = await createTransfer(context, randomAccount.address, 1_000_000_000_000, { gasPrice: MIN_GAS_PRICE });
+        console.log("here 2")
+        await customWeb3Request(context.web3(), "eth_sendRawTransaction", [
+          tx,
+        ]); 
+        await context.waitBlock(2, "Container2001");
```
tmpolaczyk, line 205: I guess ideally this should wait until the transaction is included in a block? I made a helper function for that in my other PR, but it requires you to use the polkadot js api (signAndSend), so not sure if it can be used here.

https://github.com/moondance-labs/tanssi/blob/313a0a03652ae065af8bfe89e2fb2aa0688877ac/test/util/block.ts#LL64C1-L64C1

container-chains/templates/frontier/node/src/service.rs
girazoki: I did a similar thing to moonbeam where we have a different path depending on whether we have a rocksDb or a Parity db database. I think we should stay with that...

test/util/ethereum.ts
```diff
@@ -211,4 +209,19 @@ export const createTransaction = async (
...
+/// const txHash = await tx.signAndSend(alice);
+/// // We don't know if the transaction has been included in a block or not
+/// const { txHash, blockHash } = await signAndSendAndInclude(tx, alice);
+/// // We know the blockHash of the block that includes this transaction
+/// ```
+export async function waitUntilIncluded(promise, web3, txHash)  {
+  while ((await customWeb3Request(web3, "eth_getTransactionByHash", [
+    txHash.result,
+  ])).result.blockNumber== null) {  
+    await promise;
```
tmpolaczyk, line 225: In Rust this would only wait the first time, not sure how it works in javascript

container-chains/templates/frontier/node/src/command.rs
```diff
@@ -225,21 +245,22 @@ pub fn run() -> Result<()> {
                 cmd.run(&*spec)
             })
         }
+        Some(Subcommand::FrontierDb(_)) => todo!(),
```
tgmichel, line 248: This command is likely to be deprecated. The original idea was to be able to explore the secondary rocks db using the cli, but we are probably moving to sql now that is merged upstream, and we obviously don't need this for it.

container-chains/templates/frontier/node/src/rpc/mod.rs
```diff
@@ -0,0 +1,330 @@
...
+            params.pubsub_notification_sinks.clone(),
+        )
+        .for_each(|()| futures::future::ready(())),
+    );
+
+    // Frontier `EthFilterApi` maintenance.
+    // Manages the pool of user-created Filters.
+    if let Some(filter_pool) = params.filter_pool {
+        // Each filter is allowed to stay in the pool for 100 blocks.
+        const FILTER_RETAIN_THRESHOLD: u64 = 100;
```
tgmichel, line 271: Makes me wonder how this (and other block-based on-memory caches) will work with parathreads. Are those retained in memory for an undetermined amount of time for X blocks?

container-chains/templates/frontier/node/src/service.rs
```diff
@@ -202,6 +269,27 @@ async fn start_node_impl(
         })
         .await?;
 
+    let overrides = crate::rpc::overrides_handle(client.clone());
```
tgmichel, line 272: Just a comment on storage overrides. If you want to disable it and use runtime api, if you want to mantain a performant server you will need to increase `runtime_cache_size` to a value as high as possible to avoid compiling `VersionedRuntime`s onthefly. This shouldn't be an issuein the beginning but I guess something to take into consideration.

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -742,6 +773,158 @@ impl_runtime_apis! {
...
+            data: Vec<u8>,
+            value: U256,
+            gas_limit: U256,
+            max_fee_per_gas: Option<U256>,
+            max_priority_fee_per_gas: Option<U256>,
+            nonce: Option<U256>,
+            estimate: bool,
+            access_list: Option<Vec<(H160, Vec<H256>)>>,
+        ) -> Result<pallet_evm::CallInfo, sp_runtime::DispatchError> {
+            let config = if estimate {
```
tgmichel, line 818: You likely want to just remove this if you are compiling with `rpc-binary-search-estimate` feature, and use the evm without `estimate` mode (it's buggy and useless for the most part). I will also remove it in moonbeam.

container-chains/templates/frontier/node/src/rpc/mod.rs
girazoki: good question. I guess blocks makes more sense

girazoki: let's review later, will add TODO for this

tmpolaczyk (approved):

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -545,6 +574,8 @@ impl pallet_evm::Config for Runtime {
     type OnChargeTransaction = ();
     type OnCreate = ();
     type FindAuthor = FindAuthorTruncated<Aura>;
+    type Timestamp = Timestamp;
+    type WeightInfo = ();
```
tgmichel, line 578: fyi WeightInfo for pallet-evm has been recently added to frontier, so you don't forget to update this once you pin a more recent version.

tgmichel (approved):
Added a comment, everything LGTM!

PR #108, opened 2023-05-29
dedicated primitives folder for container chains

nanocryk:
Move `authorities-noting-inherent` primitive crate inside the `container-chains` folder as it is used only by them, and makes it stand out from `author-noting-inherent` which is specific to the orchestrator chain.


Cargo.toml
```diff
@@ -32,13 +33,13 @@ pallet-registrar-runtime-api = { path = "pallets/registrar/rpc/runtime-api", def
 
 container-chain-template-runtime = { path = "container-chain-template/runtime", default-features = false }
 pallet-cc-authorities-noting = { path = "container-chains/pallets/authorities-noting", default-features = false }
+tp-authorities-noting-inherent = { path = "container-chains/primitives/authorities-noting-inherent", default-features = false }
```
girazoki, line 36: I would then rename this as ccp-authorities-noting-inherent, to make it consistent with the naming conventions we are following

girazoki:
Do not worry about the test, this is something to be fixed on a separate PR that I will open shortly

PR #109, opened 2023-05-31
Girazoki evm manual seal

girazoki:


container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -321,7 +330,7 @@ pub const MILLIUNIT: Balance = 1_000_000_000;
 pub const MICROUNIT: Balance = 1_000_000;
 
 /// The existential deposit. Set to 1/10 of the Connected Relay Chain.
```
tmpolaczyk, line 332:
```suggestion
/// The existential deposit. Set to 0 because this is an ethereum-like chain
```

girazoki: indeed

container-chains/templates/frontier/node/src/service.rs
```diff
@@ -413,3 +439,325 @@ pub async fn start_parachain_node(
...
+    }
+
+    log::info!("Development Service Ready");
+
+    network_starter.start_network();
+    Ok(task_manager)
+}
+
+/// Block authoring scheme to be used by the dev service.
+#[derive(Debug, Copy, Clone)]
```
girazoki, line 738: Potential to move it somewhere else

client/consensus/src/manual_seal.rs
```diff
@@ -134,3 +134,98 @@ where
...
+            .expect("Timestamp is always present; qed");
+
+        // we always calculate the new slot number based on the current time-stamp and the slot
+        // duration.
+        // TODO: we need to add the nimbus digest here
+        let slot = Slot::from_timestamp(timestamp, self.slot_duration);
+        let aura_digest_item =
+            <DigestItem as CompatibleDigestItem<NimbusSignature>>::aura_pre_digest(slot);
+
+        // Fetch the authorities for the orchestrator chain
```
girazoki, line 196: remove

test/suites/dev-frontier-template/test-eth-fee/test-eth-fee-history.ts
```diff
@@ -0,0 +1,174 @@
...
+      reward_percentiles: number[],
+      priority_fees: number[],
+      max_fee_per_gas: string
+    ) {
+      let nonce = await context
+        .viemClient("public")
+        .getTransactionCount({ address: ALITH_ADDRESS });
+      const contractData = getCompiled("MultiplyBy7");
+      for (var b = 0; b < block_count; b++) {
+        for (var p = 0; p < priority_fees.length; p++) {
```
tmpolaczyk, line 36: Please always use let instead of var.

test/moonwall.config.json
```diff
@@ -22,6 +22,25 @@
...
+    {
+      "name": "dev_frontier_template",
+      "testFileDir": ["suites/dev-frontier-template"],
+      "multiThreads": true,
+      "html": false,
+      "foundation": {
+        "type": "dev",
+        "launchSpec": [
+          {
+            "name": "tanssi",
```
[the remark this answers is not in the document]
girazoki: ah good catch. I guess not that important

tmpolaczyk (approved):

container-chains/templates/frontier/runtime/src/impl_on_charge_evm_transaction.rs
```diff
@@ -0,0 +1,64 @@
...
+				<EVMCurrencyAdapter<<T as pallet_evm::Config>::Currency, OU> as OnChargeEVMTransactionT<
+					T,
+				>>::correct_and_deposit_fee(who, corrected_fee, base_fee, already_withdrawn)
+			}
+
+			fn pay_priority_fee(tip: Self::LiquidityInfo) {
+				if let Some(tip) = tip {
+					OU::on_unbalanced(tip);
+				}
+			}
```
tgmichel, line 61: Just a note to remember that this overrides the original `pallet-evm` impl by removing block author tipping. We do it like this in Moonbeam because we already have other incentives in place like staking rewards. We still ration the tip (`on_unbalanced`) but this is something that you guys need to keep in mind if you want block author tipping to be part of your tokenomics.

tgmichel (approved):
Just a simple comment, LGTM!

container-chains/templates/frontier/runtime/src/impl_on_charge_evm_transaction.rs
girazoki: there is no author here that we can reward (or at least we shouldnt since authors will not have a 20 byte key), so we probably should by defualt burn. But it is true that we should add a test for this, will add it

PR #110, opened 2023-05-31
Add isEthereum ChainSpec property

tmpolaczyk:
Add `isEthereum` field to ChainSpec properties. In the future this will allow polkadot.js and similar apps to detect whether a parachain is using ethereum-style addresses, but for now it does nothing. The default value is `true` for the frontier template, and `false` for the simple template and for the orchestrator chain.

### DIFF
--- a/primitives/container-chain-genesis-data/src/lib.rs
+++ b/primitives/container-chain-genesis-data/src/lib.rs
@@ -88,6 +88,7 @@ pub struct TokenMetadata {
     pub token_symbol: BoundedVec<u8, MaxLengthTokenSymbol>,
     pub ss58_format: u32,
     pub token_decimals: u32,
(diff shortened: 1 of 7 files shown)


primitives/container-chain-genesis-data/src/lib.rs
```diff
@@ -97,6 +98,7 @@ impl Default for TokenMetadata {
             token_symbol: BoundedVec::truncate_from(b"UNIT".to_vec()),
             ss58_format: 42,
             token_decimals: 12,
+            is_ethereum: false,
```
girazoki, line 101: Since this is not token metadata per se, I would make a new struct of the form:

```
struct Properties {
token_metadata: TokenMetadata
is_ethereum: bool
}
```

PR #111, opened 2023-06-01
Add token deposit to registrar pallet, and add markValidForCollating extrinsic

tmpolaczyk:
Now the registrar::register extrinsic can be called by any signed origin, as long as they provide a deposit. The registered para ids are not considered by collators until a root origin calls markValidForCollating.


pallets/registrar/src/lib.rs
```diff
@@ -159,37 +173,52 @@ pub mod pallet {
...
+                Ok(_) => return Err(Error::<T>::ParaIdAlreadyRegistered.into()),
+                Err(index) => {
+                    pending_verification
+                        .try_insert(index, para_id)
+                        .map_err(|_e| Error::<T>::ParaIdListFull)?;
+                }
+            }
 
-            // TODO: while the registration takes place on the next session, the genesis data
+            // While the registration takes place on the next session, the genesis data
```
tmpolaczyk, line 221: This comment is a bit outdated, registration happens on the next session after someone else calls `mark_valid_for_collating`, I will reword it.

test/suites/dev/registar/test_registrar_register.ts
```diff
@@ -52,8 +52,10 @@ describeSuite({
         const containerChainGenesisData = emptyGenesisData();
 
         const tx = polkadotJs.tx.registrar.register(2002, containerChainGenesisData);
-        await polkadotJs.tx.sudo.sudo(tx).signAndSend(alice);
-
+        await tx.signAndSend(alice);
+        await context.createBlock();
+        const tx2 = polkadotJs.tx.registrar.markValidForCollating(2002);
+        await polkadotJs.tx.sudo.sudo(tx2).signAndSend(alice);
         await context.createBlock();
```
tmpolaczyk, line 59: This can be rewritten into using only one create block like:

```rust
          const { result } = await context.createBlock(
            tx, tx2
          );
```

girazoki (approved):

PR #112, opened 2023-06-02
Switch to proving NimbusId assignements in container chains

nanocryk:
To check blocks are authored by the proper collator we need to know which NimbusIds assigned collators are using. Knowing their orchestrator AccountId however is not useful. For that reason this PR changes the proof to provide assignment data over NimbusId. Since the storage used depends on the session id, we include the session id storage in the proof.


tmpolaczyk:
Should we change the `authority_assignment_for_session` storage so that it doesn't depend on the session?

girazoki:
I think if we decided to go this route its fine. Demonstrating an additional storage item does not add that much overhead, the only thing potentially worrying is that this adds more dependency with pallet indices and storages.

It also gives us more flexibility in case we want to store more than 2 session keys.

girazoki (approved):
I think it's good, once we fix the tests for zombienet. You probably only have to change the accountId for the nimbusId in the verification

girazoki (approved):

girazoki (approved):

girazoki (approved):

girazoki (approved):

test/suites/para/test_tanssi_containers.ts
```diff
@@ -82,7 +82,7 @@ describeSuite({
       title: "Test assignation did not change",
       test: async function () {
         const currentSession = (await paraApi.query.session.currentIndex()).toNumber();
-        expect(currentSession).to.be.equal(0);
+        // expect(currentSession).to.be.equal(0);
```
girazoki, line 85:
```suggestion
```

test/suites/para/test_tanssi_containers.ts
```diff
@@ -69,7 +69,7 @@ describeSuite({
       title: "Test Tanssi assignation is correct",
       test: async function () {
         const currentSession = (await paraApi.query.session.currentIndex()).toNumber();
-        expect(currentSession).to.be.equal(0);
+        // expect(currentSession).to.be.equal(0);
```
girazoki, line 72:
```suggestion
```

PR #114, opened 2023-06-05
Girazoki start benchmarking efforts

girazoki:


pallets/registrar/src/benchmarks.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    register {
+        // We make it dependent on the size of the runtime
+        let x in 5..3_000_000;
+        // ..and on the nummber of parachains already registered
+        let y in 1..50;
+
+        let caller: T::AccountId = account("account id", 0u32, 0u32);
+        let storage =  ContainerChainGenesisData {
+            storage: vec![(vec![], vec![0; x as usize]).into()],
```
tmpolaczyk, line 38: There is a TODO in pallet_registrar that says that this may have a higher cost (n empty keys instead of 1 key with n bytes):
 
```suggestion
            storage: vec![(vec![], vec![]).into(); x as usize],
```

We probably want a separate benchmark for that? Not sure what's the best practice here.

benchmarking/frame-weight-template.hbs
```diff
@@ -0,0 +1,136 @@
+// Copyright 2019-2022 PureStake Inc.
```
tmpolaczyk, line 1: This copyright is wrong, although I fixed it in #115

tmpolaczyk: Makes sense, I'll add you as a coauthor :P

girazoki:
Closing in favor of #115

PR #115, opened 2023-06-06
Benchmark pallet-configuration

tmpolaczyk:


pallets/configuration/src/benchmarks.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn new_test_ext() -> TestExternalities {
+        let t = frame_system::GenesisConfig::default()
+            .build_storage::<Test>()
+            .unwrap();
+        TestExternalities::new(t)
+    }
+
+    #[test]
+    fn bench_register() {
+        new_test_ext().execute_with(|| {
```
girazoki, line 50: this looks like wrong ;)

tmpolaczyk: It was intentional to check if CI checks that, and it doesn't. It is supposed to check that these tests compile but not run them, right?

girazoki: We can evaluate later on if we should run tests with runtime-benchmarks

tmpolaczyk:
```
running 10 tests
test mock::__construct_runtime_integrity_test::runtime_integrity_tests ... ok
test tests::config_sets_values_from_genesis ... ok
test tests::config_set_value ... ok
test benchmarks::bench_set_config_with_u32 ... ok
test tests::config_sets_default_values ... ok
test tests::config_set_many_values_different_sessions ... ok
test tests::config_set_many_values_same_block ... ok
test tests::config_set_many_values_different_blocks ... ok
test benchmarks::benchmark_tests::test_benchmarks ... ok
test benchmarks::tests::bench_set_config_with_u32 ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

What's the difference between `benchmarks::bench_set_config_with_u32`, `benchmarks::benchmark_tests::test_benchmarks` and `benchmarks::tests::bench_set_config_with_u32`?

runtime/test/src/lib.rs
```diff
@@ -669,6 +669,103 @@ impl_runtime_apis! {
...
+                        add_benchmark, BenchmarkBatch, Benchmarking, TrackedStorageKey,
+                    };
+
+                    use frame_system_benchmarking::Pallet as SystemBench;
+                    impl frame_system_benchmarking::Config for Runtime {}
+                    use pallet_configuration::Pallet as PalletConfigurationBench;
+
+                    let whitelist: Vec<TrackedStorageKey> = vec![
+                        // Block Number
+                        hex_literal::hex!(  "26aa394eea5630e07c48ae0c9558cef7"
```
girazoki, line 708: We should remove any whitelisted key or at least make sure they match the key in our runtimes.

girazoki (approved):

PR #116, opened 2023-06-08
Girazoki add pallet registrar benchmarks

girazoki:
Benchmarks pallet-registrar extrinsics

### DIFF
--- a/pallets/registrar/src/benchmarks.rs
+++ b/pallets/registrar/src/benchmarks.rs
@@ -50,11 +50,10 @@ mod benchmarks {
 
     #[benchmark]
     fn register(x: Linear<5, 3_000_000>, y: Linear<1, 50>, z: Linear<1, 10>) {
-        
         let mut data = vec![];
         // Number of keys
         for _i in 1..z {
-            data.push((b"code".to_vec(), vec![1; (x/z) as usize]).into())
+            data.push((b"code".to_vec(), vec![1; (x / z) as usize]).into())
         }
 
         let storage = ContainerChainGenesisData {


pallets/registrar/src/lib.rs
```diff
@@ -267,12 +278,15 @@ pub mod pallet {
...
+        /// Deregister container-chain.
         ///
-        /// If a parachain is registered but not marked as valid_for_collating, this will remove it
+        /// If a container-chain is registered but not marked as valid_for_collating, this will remove it
         /// from `PendingVerification` as well.
         #[pallet::call_index(1)]
-        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(2,2).ref_time())]
+        #[pallet::weight(T::WeightInfo::deregister(
+            T::MaxGenesisDataSize::get(),
+            T::MaxLengthParaIds::get()
```
girazoki, line 288: We could make this a weight hint but I think it would over-complicate things, also in terms of UI. But I am open to suggestions. We can also add it in future PR

tmpolaczyk: We can add a separate `deregister_with_hints` extrinsic in the future, if the weight becomes a problem.

pallets/registrar/src/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+        let x in 5..3_000_000;
+        // ..and on the nummber of parachains already registered
+        let y in 1..50;
+
+        // Worst case scenario is when the parachain is not yet valid for collation
+        let caller: T::AccountId = account("account id", 0u32, 0u32);
+        let storage =  ContainerChainGenesisData {
+            // Runtime would go under "code" key, so we mimic
+            // with 4 byte key
+            storage: vec![(vec![1; 4], vec![1; x as usize]).into()],
```
tmpolaczyk, line 97: You can just set this to code

```suggestion
            storage: vec![(b"code".to_vec(), vec![1; x as usize]).into()],
```

pallets/registrar/src/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+    T::Currency::issue(total);
+    (user, total)
+}
+
+benchmarks! {
+    where_clause { where T::SessionIndex: From<u32> }
+
+    register {
+        // We make it dependent on the size of the runtime
+        let x in 5..3_000_000;
```
notlesh, line 52: I assume this is poking at proof size, but the resulting benchmarks don't appear to show any proof-size-per-x. Am I missing something? Maybe `verify` should show that the expected runtime blob exists in storage with the expected size.

notlesh (commented):
I have a few thoughts:

1. None of the benchmarks seem to capture the proof_size scaling for the code size. Maybe I'm missing something, though.
2. I would suggest the "new" benchmark style seen [here](https://github.com/paritytech/substrate/blob/31b57a1884b00ed979ba85343e308ed73756cb41/frame/benchmarking/src/lib.rs#L77). Sorry I didn't mention this before these were written :sweat_smile:
3. Some sort of test against these would be nice -- in particular, it would be good to show the very important proof_size aspect works. Things like "a big proof_size costs a lot" and "only one max-size proof fits in a block", etc.

Another thing to consider is that `pallet-transaction-payment` allows for specifying a length fee, and this would stack on top of any proof_size-related fee. That's probably a good thing, but the fee could be obscenely high as a result...

pallets/registrar/src/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+        // ..and on the nummber of parachains already registered
+        let y in 1..50;
+
+        // Worst case scenario is when the parachain is not yet valid for collation
+        let caller: T::AccountId = account("account id", 0u32, 0u32);
+        let storage =  ContainerChainGenesisData {
+            // Runtime would go under "code" key, so we mimic
+            // with 4 byte key
+            storage: vec![(b"code".to_vec(), vec![1; x as usize]).into()],
```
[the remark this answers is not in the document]
girazoki: we have a storage in pallet-registrar where we store genesis-state for cotnainer-chains (in the form of key-value pairs). Tipically the key-value that would be more problematic is the code, and for reference, I injected a very big value in the "code" key

girazoki:
Thank you Stephen. A couple of thoughts on the proof size:

- I am not expecting the proof size to be shown in `register`, as we know here the long code size would be injected through an extrinsic, and it's probably not accounted in this case.

- I have doubts with de-register. Here the only thing we are doing is **removing** a previously injected large code. We do not need to read this value whatsoever. So it might be that it does not reflect the proof size because of this..

girazoki:
@notlesh I just checked and essentially, the reason why the proof size does not increase linearly with the code injected is the ones I mentioned above. As soon as I introduce a read (e..g, in de-register) then the proof size linear component appears.

pallets/registrar/src/lib.rs
notlesh: hints are such an unpleasant UX/DX, I'd be in favor of avoiding them if possible.

Maybe we could have both: one extrinsic that assumes worst-case and refunds and another that takes a weight hint. That way we can at least offer users to choose their flavor of shit

pallets/registrar/src/benchmarks.rs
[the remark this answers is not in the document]
notlesh: I wondered if you were trying to avoid a whitelisted storage key -- I think `"code"` is commonly whitelisted.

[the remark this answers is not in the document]
girazoki: this would not be code though, it would be` registrar->genesistate->map("code", value)`

notlesh:
> @notlesh I just checked and essentially, the reason why the proof size does not increase linearly with the code injected is the ones I mentioned above. As soon as I introduce a read (e..g, in de-register) then the proof size linear component appears.

That's a good way to test the behavior, nice idea.

I can pretty clearly see that we only ever (1) `insert` this data and (2) `remove` it, neither of which impact proof size. So the results make sense.

I find it interesting that both benchmarks associate a tiny amount of `ref_time` with the size, maybe that's just the SCALE encoding/decoding of a Vec...

Anyway, I'm content with the results now, thanks for being thorough with me about it.

notlesh (approved):

pallets/registrar/src/benchmarks.rs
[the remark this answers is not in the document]
notlesh: Thanks, I'm clear on that now :)

pallets/registrar/src/lib.rs
girazoki: I think this is a nice idea yes

tmpolaczyk (approved):

tmpolaczyk:
Run `cargo fmt` and also remove the comment that says

    // TODO: for benchmarks

PR #120, opened 2023-06-12
[Do not merge] Remove individual container nodes

girazoki:
Removes individual container-chain nodes in favour of a subcommand in the tanssi-node called `run-container-command`, e.g.:

- `container-chain-frontier-template args`

for

- `tanssi-node run-container-command ethereum args`

and 
- `container-chain-simple-template args`

for

- `tanssi-node run-container-command simple args`

This ensures that it is not necessary to compile two additional nodes and the tanssi-node can manage directly container-chain commands.

Zombienet tests not working because there is an incompatibility with moonwall, which checks the entire "command" is executable


node/src/cli.rs
```diff
@@ -351,3 +378,36 @@ impl ContainerChainCli {
         Ok(())
     }
 }
+
+#[derive(Debug, clap::Parser)]
+#[group(skip)]
+pub struct EthRunCmd {
```
girazoki, line 384: remove

girazoki:
Will leave it in stale mode for now. We should use each of the different clients separately

PR #121, opened 2023-06-13
Add RPC method to convert genesis data into chain spec JSON

tmpolaczyk:


girazoki:
Let's add a test for this. It can be in dev-mode, as you dont need other chains for this

test/suites/dev/registar/test_utils_rpc.ts
```diff
@@ -0,0 +1,111 @@
...
+      alice = keyring.addFromUri('//Alice', { name: 'Alice default' });
+      bob = keyring.addFromUri('//Bob', { name: 'Bob default' });
+      polkadotJs = context.polkadotJs();
+    });
+
+    it({
+        id: "E01",
+        title: "Read a ChainSpec, convert it to ContainerChainGenesisData, and back to the same ChainSpec",
+        test: async function () {
+            debugger;
```
tmpolaczyk, line 28:
```suggestion
```

tmpolaczyk:
I added a test that reads a chain spec file, converts it into the onchain version, and then back into a chain spec file using the new RPC method. But that test doesn't work in CI because the chain spec files do not exist in the dev_tanssi environment (we need the template binaries to generate them). So I see two options here, add the binaries to the github environment, or change the test so that it uses a mocked chain spec. What do you think is better?

test/suites/dev/registar/test_utils_rpc.ts
```diff
@@ -0,0 +1,128 @@
...
+              Properties: {
+                token_metadata: "TokenMetadata",
+                is_ethereum: "bool",
+              },
+              TokenMetadata: {
+                // TODO: this is actually a Vec<u8>, but that doesn't work because polkadot.js converts the
+                // Vec<u8> into hex bytes, while the Rust code doesn't work with hex bytes because this is
+                // actually a BoundedVec, and there is no easy way to serialize a BoundedVec as hex bytes.
+                // Ideally this should simply be a string, because this is a token name like "UNIT".
+                token_symbol: "Vec<u16>",
```
girazoki, line 49: Cant this be bytes?

tmpolaczyk: Also I tried `Bytes` and `Vec<u8>` to see if one of them works, but they are exactly the same, both give the same error.

girazoki: For now let's leave it like this then, lets try to remember this for the future

girazoki: we can also see check how polkadotjs handles boundedVecs

primitives/container-chain-genesis-data/src/json.rs
```diff
@@ -200,3 +200,74 @@ pub fn properties_to_map(
...
+            }
+        }"#
+    }
+
+    #[test]
+    fn test_serde_serialize() {
+        let x = expected_container_chain_genesis_data();
+        let xv = serde_json::to_value(&x).unwrap();
+        // Regenerate expected string using
+        //println!("{}", serde_json::to_string_pretty(&x).unwrap());
```
girazoki, line 260: remove

tmpolaczyk: This is intentional, in case we want to update this test with new values in the future.

girazoki:
let's check how to make this rpc method insecure

girazoki (approved):
This should be changed in a follow up pr to be a script, however, since there is a lot of common code between both prs, will approve this to re-use work from this pr

PR #124, opened 2023-06-14
Benchmark author-noting pallet

tmpolaczyk:
This is missing a way to create storage proof from the benchmark itself, which is not possible because the sproof builder doesn't work in no-std. Currently I am using a test to generate the proof, save that into the `mock_proof.rs` file, and import that file from benchmarks.


pallets/author-noting/src/lib.rs
```diff
@@ -141,7 +185,11 @@ pub mod pallet {
                 GenericStateProof::new(relay_storage_root, relay_storage_proof)
                     .expect("Invalid relay chain state proof");
 
-            for para_id in T::ContainerChains::current_container_chains() {
+            let registered_para_ids = T::ContainerChains::current_container_chains();
+            let total_weight =
+                T::WeightInfo::set_latest_author_data(registered_para_ids.len() as u32);
+            for para_id in registered_para_ids {
+                log::info!("fetch_author_slot_from_proof {}", u32::from(para_id));
```
tmpolaczyk, line 192:
```suggestion
```

pallets/author-noting/src/lib.rs
```diff
@@ -54,15 +55,60 @@ mod mock;
...
+#[cfg(any(test, feature = "runtime-benchmarks"))]
+mod benchmarks;
+
+mod mock_proof;
 
 pub use pallet::*;
 
+use crate::weights::WeightInfo;
+
+// TODO: we can remove this once cumulus_pallet_parachain_system::RelaychainStateProvider has a
```
tmpolaczyk, line 69: Cherrypick that commit to our fork of cumulus

pallets/author-noting/src/test.rs
```diff
@@ -363,3 +363,44 @@ fn test_not_inserting_inherent() {
             assert!(AuthorNoting::latest_author(ParaId::from(1001)).is_none());
         });
 }
+
+#[test]
+fn encode_proof_for_benchmarks() {
```
tmpolaczyk, line 368:
```suggestion
#[test]
#[ignore]
fn encode_proof_for_benchmarks() {
```

We don't need to run this test all the time

pallets/author-noting/Cargo.toml
```diff
@@ -9,8 +9,10 @@ version = "0.1.0"
 [package.metadata.docs.rs]
 targets = [ "x86_64-unknown-linux-gnu" ]
 [dependencies]
+frame-benchmarking = { workspace = true, optional = true }
 frame-support = { workspace = true }
 frame-system = { workspace = true }
+hex = { workspace = true, features = [ "alloc" ] }
```
girazoki, line 15: this probably can be optional, we only use it in runtime-benchmarks right?¿

tmpolaczyk: Yes, I only use it for creating and reading the mock_proof.rs. So it is needed for `any(runtime-benchmarks, test)`.

girazoki: ok make it optional then and we import it with std and runtime-benchmarks

girazoki (approved):
LGTM, we will have to evaluate in the future how to clean up the mock_proof.rs, but overall it's good

PR #126, opened 2023-06-16
Benchmark pallet-cc-authorities-noting

tmpolaczyk:


tmpolaczyk:
I don't understand why is CI failing, I run the same command locally and it works:

```
cargo check --features=runtime-benchmarks --release --all
```

tmpolaczyk:
Thanks @girazoki, not sure what happened here. It was working for me locally because I had modified the files in the `.cargo` dir, but not sure why the Cargo.lock had changes.

container-chains/pallets/authorities-noting/src/benchmarks.rs
```diff
@@ -0,0 +1,112 @@
...
+            orchestrator_chain_state: orchestrator_proof,
+        };
+
+        T::RelayChainStateProvider::set_current_relay_chain_state(cumulus_pallet_parachain_system::RelayChainState {
+            state_root: relay_root,
+            number: 0,
+        });
+        #[cfg(test)]
+        {
+            // set_current_relay_chain_state doesn't work in tests, we need to write the MOCK_RELAY_ROOT_KEY.
```
girazoki, line 87: Why wouldnt it? can we make it such that the `RelayChainStateProvider` set_current_relay_chain_state matches what you are doing here?

tmpolaczyk: We would need to modify the `MockRelayChainStateProvider`.

PR #127, opened 2023-06-19
start preparing build-runtime job

girazoki:
Adds a CI job able to publish the runtime release on request, giving the previous tag name and the new one


tmpolaczyk:
In the `tests/` dir we don't commit `package-lock.json`, we commit the `pnpm-lock.yaml` created by `pnpm`. So maybe it is worth it to also use pnpm here?

girazoki:
I tried pnpm but failed. Will try in a future pr, for now I will leave yarn

.github/workflows/publish-runtime.yml
```diff
@@ -0,0 +1,166 @@
...
+        required: true
+      to:
+        description: tag (ex. runtime-155) to generate release note and srtool runtimes from
+        required: true
+
+jobs:
+  ####### Build runtimes with srtool #######
+
+  setup-scripts:
+    runs-on: self-hosted
```
crystalin, line 18: changing to use label instead of generic self-hosted will be easier for maintenance

.github/workflows/publish-runtime.yml
```diff
@@ -0,0 +1,166 @@
...
+    steps:
+      - uses: actions/checkout@v3
```
crystalin, line 20: If you want more comments about this, it is to ensure the script are executed from the current branch and not from the tags being checked out later in the process

.github/workflows/publish-runtime.yml
```diff
@@ -0,0 +1,166 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: original-scripts
+          path: original-scripts
+      - name: Build & Push purestake/srtool image
+        if: github.repository == 'moondance-labs/tanssi'
+        run: |
+          docker pull "${{ matrix.srtool_image }}:${{ matrix.srtool_image_tag }}" && image_exists=true || image_exists=false
+
+          if [[ $image_exists = "false" ]]; then
```
crystalin, line 56: You might want to do something else than try to build it, as you won't be able to push to moonbeam docker hub from here.
I think simply making the script fail is enouigh

crystalin (approved):

.github/workflows/publish-runtime.yml
girazoki: yes, we have an issue for that already. But will do a PR changing all those at once.

PR #128, opened 2023-06-19
Point custom branches with stable compilation

girazoki:
Points at custom branches for all repos and cherry-picks the stable compilation changes for all of them

We were also importing two different cumulus repos before, which now should be fixed


tmpolaczyk:
Can confirm that we no longer depend on two versions of cumulus.

PR #129, opened 2023-06-19
Remove utils RPC in favor of typescript script

tmpolaczyk:
To test locally:

* Start tanssi node using `pnpm moonwall run zombie_tanssi`
* Run this command to save an on-chain chain spec to a file

```
pnpm run download-chain-spec --url ws://127.0.0.1:9948 --para-id 2001 --output /tmp/spec2001a.json --chain-type Local --relay-chain rococo-local
```

* Run this command to register a new parachain (but you still need to manually "markValidForCollating" using sudo)

```
pnpm run register-para --url ws://127.0.0.1:9948 --account //Alice --chain ../specs/template-container-2002.json
```


girazoki:
Can you provide a description on what the pr does?

girazoki:
Can you additionally create a simple tool (that we can run with either npm or sh) using the new typescript functions? I think this would be good for operational purposes

girazoki:
Ideally this script should take a ws-provider as an argument, and should be able to generate to a file the chain spec

tmpolaczyk:
Working on a separate typescript script, any suggestions for name/usage? This is what I am thinking:

```
# register para from chain spec
pnpm run register-para --url wss://rpc.example.com --chain ../specs/template-container-2000.json

# create chain spec from registered para
pnpm run download-chain-spec --url wss://rpc.example.com --para-id 2000 --chain-type foo --relay-chain kusama
```

girazoki:
I think the script names look good, I would keep it like that

girazoki (approved):

tmpolaczyk:
I agree that this would fit better in `scripts/` instead of `test/scripts`, but that may not be easy to change because this script depends on a file from `test/utils`, and not sure if you can import a single file from outside the project root. So for this PR I am going to leave the scripts here if you don't mind.

girazoki:
no worries yes, we will check later on how to move them somewhere where it makes more sense

PR #131, opened 2023-06-20
Cancel CI job

girazoki:
Allows to cancel previously running job if new commit is pushed


girazoki:
Can you additionally create a simple tool (that we can run with either npm or sh) using the new typescript functions? I think this would be good for operational purposes

librelois (approved):

PR #132, opened 2023-06-20
[MD-27] use OrchestratorParaId as storage item

nbaztec:
makes `OrchestratorParaId` in `pallet-cc-authorities-noting` as a storage item.


container-chains/pallets/authorities-noting/src/lib.rs
```diff
@@ -124,6 +122,31 @@ pub mod pallet {
         }
     }
 
+    #[pallet::storage]
+    #[pallet::getter(fn orchestrator_para_id)]
+    pub type OrchestratorParaId<T: Config> = StorageValue<_, ParaId, ValueQuery>;
```
girazoki, line 127:
```suggestion
    pub type OrchestratorParaId = StorageValue<_, ParaId, ValueQuery>;
```

container-chains/pallets/authorities-noting/src/lib.rs
```diff
@@ -149,7 +172,7 @@ pub mod pallet {
             let relay_storage_root =
                 T::RelayChainStateProvider::current_relay_chain_state().state_root;
 
-            let para_id = T::OrchestratorParaId::get();
+            let para_id = OrchestratorParaId::<T>::get();
```
girazoki, line 175: We probably should re-run benchmarks after this

container-chains/pallets/authorities-noting/src/lib.rs
```diff
@@ -124,6 +122,31 @@ pub mod pallet {
         }
     }
 
+    #[pallet::storage]
```
girazoki, line 125: It would be good to have a root-based extrinsic that can change this? Just in case for whatever reason, this gets updated

girazoki:
If we add the extrinsic, we can also add a typescript test for it under `dev-frontier-template` suite

girazoki (approved):

tmpolaczyk (approved):

PR #133, opened 2023-06-20
[MD-99] add sudo and utility pallets

nbaztec:
adds sudo and utility pallets


container-chains/templates/simple/node/src/chain_spec.rs
```diff
@@ -170,6 +171,7 @@ pub fn local_testnet_config(para_id: ParaId, seeds: Option<Vec<String>>) -> Chai
                     .collect(),
                 default_funded_accounts.clone(),
                 para_id.into(),
+                collator_accounts[0].clone(),
```
tmpolaczyk, line 174: The code looks correct, but for some reason when running zombienet the sudo pallet is not using Alice key. As a result, I can't use the sudo pallet because I don't have the sudo key.

This is the sudo key of container 2000:

5F9QPCe6FYdgeStu4MgXHDAstu32HVohzDnXsfekVXBJV5Cu

This is Alice key in container 2000:

5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY

In the orchestrator chain, the sudo key matches this Alice key.

Same happens with the frontier template.

girazoki: the sudo key is set to be the first of the collators. Since with Zombienet the collators are overriden, I am assuming it's what we call DummyCollator.

@nbaztec can we hardcode it to Alice? I think its better because in the future there will not be any collators that we set through genesis for the container-chains

girazoki: Or maybe we can set it to `default_funded_accounts[0]` instead of `collator_accounts[0]`

tmpolaczyk: Oh, I see. It's because of the `build-spec.sh` script:

```
./target/release/container-chain-template-simple-node build-spec --disable-default-bootnode --parachain-id 2000 --seeds "Collator2000-01,Collator2000-02" --raw > specs/template-container-2000.json
```

Why was the "seeds" argument needed there? Maybe we can just remove it?

girazoki: it was because there was a verification of the collators made before in the container-chains, and we needed the collators to match. I think this is not done anymore, so we could try removing them

tmpolaczyk: The key is correct after removing the `--seeds` argument, so let's try that. @nbaztec can you try to remove that argument from `test/scripts/build-spec.sh` and `test/scripts/build-spec-ci.sh`?

girazoki: I think we still need to put `default_funded_accounts[0]`. As said, in the inminent future there will not be collators to push in the genesis-spec of the containers, as these will come from tanssi

nbaztec: I feel hardcoding might be the appropriate approach here, given the n-different variations that could occur in n-different settings - that is if we absolutely want the same accounts to be always root. I feel `default_funded_accounts[0]` could be an option but given that I'm not entirely sure that this parameter wouldn't be used/overridden in future, I might just wanna avoid any future surprises.

girazoki:
We can leave the removal of the --seed parameter for another pr @tmpolaczyk.

girazoki (approved):

tmpolaczyk:
Agree, and we could maybe even remove it from the struct definition to turn it into an error, because it is going to be ignored anyway.

tmpolaczyk (approved):

PR #134, opened 2023-06-21
Node build jobs

girazoki:
Node build release CI jobs


.github/workflows/publish-binary.yml
```diff
@@ -0,0 +1,161 @@
...
+        id: create-release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.event.inputs.to }}
+          release_name: Tanssi ${{ github.event.inputs.to }}
+          body_path: body.md
+          draft: true
+      - name: Upload tanssi
```
crystalin, line 108: we can probably matrix thise part, I didn't try on Moonbeam I think

crystalin (approved):

.github/workflows/publish-binary.yml
girazoki: I can try yes

PR #136, opened 2023-06-22
upgrade moonwall

girazoki:
Upgrades moonwall to latest version


test/suites/para/test_tanssi_containers.ts
```diff
@@ -140,8 +129,9 @@ describeSuite({
       test: async function () {
         const currentSession = (await paraApi.query.session.currentIndex()).toNumber();
         const paraId = (await container2000Api.query.parachainInfo.parachainId()).toString();
-        const containerChainCollators = (await paraApi.query.authorityAssignment.collatorContainerChain(currentSession))
-          .toJSON().containerChains[paraId];
+        const containerChainCollators = (
+          await paraApi.query.authorityAssignment.collatorContainerChain(currentSession)
```
timbrinded, line 133: `toJSON()` and `toHuman()` should almost never be used under normal circumstance. I realise its 100% neccesary here, because no types are being injected in, so we should add //TODO comments to reminds us to eventually remove it

tmpolaczyk: Yeah, probably my fault for using it everywhere. As you say the problem is that we don't have type definitions, so we cannot decode this type. But for some reason we can convert it to JSON and that's exactly what we need for testing.

girazoki: added TODOs

PR #137, opened 2023-06-22
Update dependencies to 0.9.43

tmpolaczyk:
Peer discovery bug doesn't affect us anymore thanks to #142, and also I have cherry-picked a fix from upstream here: https://github.com/moondance-labs/substrate/commit/11f2f27ba5af2d4e8eb1f73ae0056106be5bcfb2


tmpolaczyk:
This PR now includes #142 , should fix the peer discovery bug

runtime/dancebox/src/lib.rs
```diff
@@ -352,9 +352,13 @@ impl pallet_balances::Config for Runtime {
     type DustRemoval = ();
     type ExistentialDeposit = ExistentialDeposit;
     type AccountStore = System;
-    type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
     type MaxReserves = ConstU32<50>;
     type ReserveIdentifier = [u8; 8];
+    type FreezeIdentifier = ();
+    type MaxFreezes = ();
+    type HoldIdentifier = ();
+    type MaxHolds = ();
```
tmpolaczyk, line 360: It is ok to use the default values for these?

client/consensus/src/consensus_orchestrator.rs
```diff
@@ -683,6 +681,13 @@ where
 
         let authorities_len = self.authorities_len(&aux_data);
 
+        info!(
+            target: logging_target,
+            "sync_oracle offline? {}, authorities_len: {:?}",
+            self.sync_oracle().is_offline(),
+            authorities_len,
+        );
```
tmpolaczyk, line 689: remove

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -951,10 +968,23 @@ impl_runtime_apis! {
         }
 
         fn elasticity() -> Option<Permill> {
-            None
+            Some(pallet_base_fee::Elasticity::<Runtime>::get())
```
tmpolaczyk, line 971: This isn't needed for the upgrade, but I noticed it was set like that in the frontier template, and this compiles. So need someone to confirm whether we want this to Some or None.

container-chains/templates/frontier/runtime/Cargo.toml
```diff
@@ -30,7 +30,10 @@ frame-support = { workspace = true }
 frame-system = { workspace = true }
 frame-system-rpc-runtime-api = { workspace = true }
 frame-try-runtime = { workspace = true, optional = true }
-pallet-balances = { workspace = true }
+# TODO: we enable this feature here, but this feature should not be enabled when compiling
+# the other substrate-based runtimes. Check if cargo does enabled it automatically because
+# it probably does (otherwise it would have to compile this crate twice)
+pallet-balances = { workspace = true, features = ["insecure_zero_ed"] }
```
tmpolaczyk, line 36: I verified this, and it depends on how we run `cargo build`. With our current CI setup, both templates are built at the same time using one cargo command, so both templates have this feature enabled.

Building them separately like this, the feature is only enabled in the frontier template:

```
cargo build --release --features=runtime-benchmarks -p container-chain-template-simple-node
cargo build --release --features=runtime-benchmarks -p container-chain-template-frontier-node
```

But I don't think this is an issue, because the only differences when compiling with this feature are:
* Disable an assert that the existential deposit is not zero
* Change the impl of have_providers_or_no_zero_ed from always returning true to doing some computation (storage read?)

```rust
        /// Returns `true` when `who` has some providers or `insecure_zero_ed` feature is disnabled.
        /// Returns `false` otherwise.
        #[cfg(not(feature = "insecure_zero_ed"))]
        fn have_providers_or_no_zero_ed(_: &T::AccountId) -> bool {
            true
        }

        /// Returns `true` when `who` has some providers or `insecure_zero_ed` feature is disnabled.
        /// Returns `false` otherwise.
        #[cfg(feature = "insecure_zero_ed")]
        fn have_providers_or_no_zero_ed(who: &T::AccountId) -> bool {
            frame_system::Pallet::<T>::providers(who) > 0
        }
```

container-chains/templates/frontier/runtime/src/lib.rs
girazoki: it really depends. But as long as base-fee is there you can leave it like this

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -371,6 +375,7 @@ parameter_types! {
 impl pallet_sudo::Config for Runtime {
     type RuntimeCall = RuntimeCall;
     type RuntimeEvent = RuntimeEvent;
+    type WeightInfo = ();
```
girazoki, line 378: we should probably bring the weights calculated by substrate

node/src/service.rs
```diff
@@ -15,6 +15,10 @@
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
+use sc_executor::HeapAllocStrategy;
```
girazoki, line 18: FMT (and others above)

runtime/dancebox/src/lib.rs
girazoki: yes for now yes

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -562,6 +567,7 @@ impl pallet_evm::Config for Runtime {
     type OnChargeTransaction = OnChargeEVMTransaction<()>;
     type OnCreate = ();
     type FindAuthor = ();
+    type GasLimitPovSizeRatio = ();
```
girazoki, line 570: Can you put a TODO here to remember to update it in the future? we really need to account pov on evm, I just dont want to delay this (also if tyou can create a ticket for this that would be great)

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -951,10 +968,23 @@ impl_runtime_apis! {
         }
 
         fn elasticity() -> Option<Permill> {
-            None
+            Some(pallet_base_fee::Elasticity::<Runtime>::get())
         }
 
         fn gas_limit_multiplier_support() {}
+
+        fn pending_block(xts: Vec<<Block as sp_api::BlockT>::Extrinsic>) -> (Option<pallet_ethereum::Block>, Option<sp_std::prelude::Vec<TransactionStatus>>) {
```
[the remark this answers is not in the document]
girazoki: Ok I see its part of a new runtime-api. maybe it should be worth testing. Maybe we should add a test covering the rpc case as explained here, or at least again write a todo: https://github.com/paritytech/frontier/pull/1048

container-chains/templates/simple/runtime/src/lib.rs
tmpolaczyk: I thought that `()` is equivalent to the default weights, which are the ones calculated by substrate?

girazoki: I dont know for sure, I always thought it does not apply any weight.

container-chains/templates/frontier/runtime/src/lib.rs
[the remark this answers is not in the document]
tmpolaczyk: Added a test (copied from upstream)

nanocryk (approved):

girazoki (approved):

container-chains/templates/simple/runtime/src/lib.rs
tmpolaczyk: They are equivalent, but it says that `()` is for backwards compatibility and tests, so maybe we should change it in the future

PR #139, opened 2023-06-23
Check author in container chains

nanocryk:
Check that the block author is the selected one for this slot among the authorities fetched from the orchestrator chain.

Changes:
- Adds `pallet_author_inherent` to perform the verification. Adds multiple structs to impl the `Config` trait.
- Change `OnTimestampSet` to a custom trait to perform slot/timestamp verification. It is in preparation of the removal of the Aura pallets.
- Various changes in the mocks to add the authorities and the block author.

TODO:
- Improve mock to remove hardcoded values.
- Work on tests


client/consensus/src/manual_seal.rs
```diff
@@ -135,6 +137,14 @@ where
     }
 }
 
+/// Helper function to generate a crypto pair from seed
+pub fn get_aura_id_from_seed(seed: &str) -> NimbusId {
```
girazoki, line 141: Probably should move it to client primitives

test/moonwall.config.json
```diff
@@ -35,7 +35,8 @@
             "binPath": "../target/release/container-chain-template-frontier-node",
             "options": [
                 "--dev",
-                "--sealing=manual"
+                "--sealing=manual",
+                "--para-id 2001"
```
girazoki, line 39:
```suggestion
                "--para-id=2001"
```

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -408,6 +428,47 @@ impl pallet_cc_authorities_noting::Config for Runtime {
     type WeightInfo = ();
 }
 
+pub struct IdentityLookup;
```
girazoki, line 431: this, since its repeated in both templates, should also be move to a primitives section probably

container-chains/primitives/authorities-noting-inherent/src/mock.rs
```diff
@@ -68,7 +78,11 @@ impl InherentDataProvider for MockAuthoritiesNotingInherentDataProvider {
         let assignment = AuthorityAssignmentSproofBuilder::<NimbusId> {
             authority_assignment: AssignedCollators {
                 orchestrator_chain: vec![],
-                container_chains: Default::default(),
+                container_chains: {
```
girazoki, line 81: We should use whatever comes from the `MockAuthoritiesNotingInherentDataProvider`. Additionally we should add own-para-id to `MockAuthoritiesNotingInherentDataProvider`

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -642,6 +645,29 @@ impl pallet_hotfix_sufficients::Config for Runtime {
     type WeightInfo = pallet_hotfix_sufficients::weights::SubstrateWeight<Runtime>;
 }
 
+pub struct CanAuthor;
```
girazoki, line 648: I want to put this into` ccp-author-verifiability,` but need to think how to do it since I dont want to depend on pallet_author_inherent..

tmpolaczyk: Why would moving this to `ccp-author-verifiability` make it depend on `pallet_authorities_noting`? It only depends on `nimbus_primitives`, right?

container-chains/primitives/authorities-noting-inherent/src/mock.rs
```diff
@@ -68,7 +71,11 @@ impl InherentDataProvider for MockAuthoritiesNotingInherentDataProvider {
...
+                    let mut chains = BTreeMap::new();
+                    chains.insert(self.container_para_id, self.authorities.clone());
+                    chains
```
tmpolaczyk, line 77: This can be done as a one-liner using `BTreeMap::from_iter([(self.container_para_id, self.authorities.clone())])`, but I prefer it to be declared above using `let container_chains = ...`

container-chains/templates/frontier/runtime/src/lib.rs
girazoki: you need to get the authorities from somewhere. In the case of containers, this will be `pallet_authorities_noting`. In the case of tanssi, this will be `pallet_authorities_assignment`. I am thinking actually most of these traits will have to go to somewhere into the tanssi primitives, and not container-chain primitives, since Tanssi will use most of them too

Cargo.lock
```diff
@@ -6110,7 +6113,7 @@ dependencies = [
 [[package]]
 name = "pallet-author-inherent"
 version = "0.9.0"
-source = "git+https://github.com/moondance-labs/nimbus?branch=tanssi-polkadot-v0.9.40#007fb6d37c8f65c505931fccd2494276865fd99e"
+source = "git+https://github.com/moondance-labs/nimbus?branch=tanssi-polkadot-v0.9.40#24d282222894a564d0cb98217de7d9ec7bc063c9"
```
tmpolaczyk, line 6116: Was this intentional? There is another commit after this: https://github.com/moondance-labs/nimbus/commits/tanssi-polkadot-v0.9.40

girazoki: yes we can point to the new commit to make sure we dont need the commit I reverted

girazoki (approved):
I have worked on cleaning up a bit the PR, and now its in approval state for me.

tmpolaczyk (approved):

PR #142, opened 2023-06-26
Add boot nodes to pallet_registrar

tmpolaczyk:
This PR adds a new storage item to pallet_registrar:

```rust
    pub type BootNodes<T: Config> = StorageMap<
        _,
        Blake2_128Concat,
        ParaId,
        BoundedVec<BoundedVec<u8, T::MaxBootNodeUrlLen>, T::MaxBootNodes>,
        ValueQuery,
    >;
```

This is a list of boot nodes for each registered container chain, stored in a format that's basically `Vec<String>` but with size bounds. It is read by the collator when it detects a new container chain assignment, and the bootnodes present on-chain are included in the generated chain spec.

These bootnodes can only be set by root.

Also adds a new CLI param to build-spec command: `--add-bootnode`, used when generating the container chain chain spec files in CI.

The full nodes that run in CI now must have a fixed p2p port, aside from having a fixed ws port. Also they must not change its node key, but that was already the case before.


tmpolaczyk:
Update build-spec-ci.sh

node/src/chain_spec.rs
```diff
@@ -146,13 +147,21 @@ pub fn development_config(
...
     mock_container_chains: Vec<ParaId>,
+    boot_nodes: Vec<String>,
 ) -> ChainSpec {
     // Give your base currency a unit name and decimal places
     let mut properties = sc_chain_spec::Properties::new();
     properties.insert("tokenSymbol".into(), "UNIT".into());
     properties.insert("tokenDecimals".into(), 12.into());
     properties.insert("ss58Format".into(), 42.into());
     properties.insert("isEthereum".into(), false.into());
+    let boot_nodes: Vec<MultiaddrWithPeerId> = boot_nodes
```
girazoki, line 158: Why do we need this in the orchestrator?  I understand that for containers you need to generate a chain-spec with specific bootnodes on it to be able to register it properly in the orchestrator, but I dont see a reason why this would be needed in the orchestrator

pallets/registrar/src/lib.rs
```diff
@@ -375,6 +396,26 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Set boot_nodes for this para id
+        #[pallet::call_index(3)]
+        #[pallet::weight(T::WeightInfo::set_boot_nodes(
+            T::MaxBootNodeUrlLen::get(),
+            T::MaxBootNodes::get(),
```
girazoki, line 404: We can probaby (at least in the second hint) inject boot_nodes.len()

pallets/registrar/src/tests.rs
```diff
@@ -281,10 +281,10 @@ fn mark_valid_for_collating_bad_origin() {
 #[test]
 fn genesis_loads_para_ids() {
     new_test_ext_with_genesis(vec![
-        (1.into(), empty_genesis_data()),
-        (2.into(), empty_genesis_data()),
-        (3.into(), empty_genesis_data()),
-        (4.into(), empty_genesis_data()),
+        (1.into(), empty_genesis_data(), vec![]),
```
girazoki, line 284: Let's add a test for the new extrinsic too

runtime/dancebox/src/lib.rs
```diff
@@ -702,12 +706,6 @@ impl_runtime_apis! {
             let mut list = Vec::<BenchmarkList>::new();
 
             list_benchmark!(list, extra, frame_system, SystemBench::<Runtime>);
-            list_benchmark!(
```
girazoki, line 705: Why was this deleted?

girazoki:
Perhaps `mark_valid_for_collating` should error if there are no bootnodes set for a container-chain. What do you think?

node/src/chain_spec.rs
tmpolaczyk: I added it for "symmetry", we can remove it.

runtime/dancebox/src/lib.rs
tmpolaczyk: It was wrong, notice it says "PalletConfigurationBench" when the name is "tp_author_noting_inherent", I noticed it when running the benchmark scripts because the config benchmarks were there twice.

node/src/chain_spec.rs
girazoki: Yeah I think it does not make much sense for now  in Tanssi.

tmpolaczyk:
TODO: remember to delete bootnodes on deregister

girazoki:
good point

girazoki:
we should update the weights after we delete the bootnodes. Other than that LGTM

PR #144, opened 2023-06-26
Fix test countUniqueBlockAuthors

tmpolaczyk:
This should fix some of the spurious CI failures.

### DIFF
--- a/test/suites/para/test_tanssi_containers.ts
+++ b/test/suites/para/test_tanssi_containers.ts
@@ -396,7 +396,7 @@ async function countUniqueBlockAuthors(paraApi, blockStart, blockEnd, numAuthors
 
   let uniq = [...new Set(actualAuthors)];
 
-  if (uniq.length <= numAuthors) {
+  if ((uniq.length > numAuthors) || (uniq.length == 1 && numAuthors > 1)) {
     console.error(
       "Mismatch between authorities and actual block authors: authorities: ",
       authorities,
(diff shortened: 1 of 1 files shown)


test/suites/para/test_tanssi_containers.ts
```diff
@@ -394,7 +396,7 @@ async function countUniqueBlockAuthors(paraApi, blockStart, blockEnd, numAuthors
 
   let uniq = [...new Set(actualAuthors)];
 
-  if (uniq.length != numAuthors) {
+  if (uniq.length <= numAuthors) {
```
girazoki, line 399: where are we checking that at least we have 2 authors?

girazoki: I mean we should check that numAuthors > 1 as well right?

tmpolaczyk: We could, but in one test we check that 5 consecutive blocks have 2 authors, so there will be a small probability of failure if one of them fails to produce 2 or 3 blocks in a row.

tmpolaczyk: Also I put the wrong `if` condition, it should be negated

girazoki: I'd say for now that is a very small probability, I would add the >1 author check. If we keep getting failures, we will evaluate, but at least we check that they take somewhat turns

girazoki (approved):

PR #146, opened 2023-06-27
start working on dockerizing

girazoki:
Sets a job in the build process to push images of` tanssi-node` and the container-chain-template nodes to the dockerhub repository


librelois (approved):
LGTM

PR #147, opened 2023-06-27
[MD-110] remove cmd arg --seeds

nbaztec:
remove cmd arg --seeds


container-chains/templates/frontier/node/src/chain_spec.rs
```diff
@@ -88,23 +88,14 @@ pub fn template_session_keys(
...
-    let collator_accounts: Vec<AccountId> = initial_collator_seeds
-        .iter()
-        .map(|seed| get_account_id_from_seed::<ecdsa::Public>(seed))
-        .collect();
-    let collator_keys: Vec<NimbusId> = initial_collator_seeds
-        .iter()
-        .map(|seed| get_collator_keys_from_seed(seed))
-        .collect();
     let mut default_funded_accounts = pre_funded_accounts();
     default_funded_accounts.extend(collator_accounts.clone());
```
girazoki, line 100: these will not work with the current implementation. I would say we keep `let initial_collator_seeds= ["Alice".to_string(), "Bob".to_string()]` for now, they will be removed later on, but for now this should work

container-chains/templates/simple/node/src/chain_spec.rs
```diff
@@ -84,23 +84,24 @@ pub fn template_session_keys(
...
     properties.insert("tokenDecimals".into(), 12.into());
     properties.insert("ss58Format".into(), 42.into());
     properties.insert("isEthereum".into(), false.into());
 
-    let initial_collator_seeds = seeds.unwrap_or(vec!["Alice".to_string(), "Bob".to_string()]);
+    let initial_collator_seeds = vec!["Alice".to_string(), "Bob".to_string()];
     let collator_accounts: Vec<AccountId> = initial_collator_seeds
         .iter()
-        .map(|seed| get_account_id_from_seed::<sr25519::Public>(seed))
+        .map(|seed| get_account_id_from_seed::<ecdsa::Public>(seed))
```
tmpolaczyk, line 98: Is this correct?

nbaztec: ah think I copied it wrong

girazoki (approved):

PR #148, opened 2023-06-27
[MD-111] remove authoring pallets from cc templates

nbaztec:
removes authoring pallets from cc templates


girazoki:
Thank you for the pr @nbaztec. For starters, all related to those pallets that you removed should be removed from the container-chain specs.

I think there is some rework to be done in the node too, as probably the runtime APIs have to change/ have to be reworked

nbaztec:
Seems `spawn_tasks` requires the client to impl `sp_session::SessionKeys` trait, so removing it causes an issue

girazoki:
let me see what I can do

girazoki:
@nbaztec I need you to resolve conflicts to trigger the CI

girazoki:
it looks like its calling generate_session_keys :)

tmpolaczyk:
The panic I see locally is:

```
Thread 'tokio-runtime-worker' panicked at 'Failed to start container chain node: Error at calling runtime api: Execution failed: Other: Exported method AuraApi_slot_duration is not found', node/src/container_chain_spawner.rs:222
```

This happens 30 seconds after starting the Collator2000-01 node.

So something is calling slot_duration? I don't see anything in the backtrace, but I can look into it if you want.

girazoki:
ah this is something I was expecting. Basically we need to call the aura api **in the orchestrator**, but not in the containers. This happens at

https://github.com/moondance-labs/tanssi/blob/9f639a5d08957576a22bb40d82bd45fed5b29404/node/src/service.rs#L836

we should call that api with the orchestrator client

tmpolaczyk:
I got dev tests working by simply setting

    parachain_config.dev_key_seed = None;

Before each call to `sc_service::spawn_tasks`.

zombienet tests still fail with the error I mentioned above.

girazoki:
I am trying what I said above

girazoki:
the change I suggested makes the zombie-tests work. So @nbaztec in node/src/service.rs, in the `build_consensus_container` function, where we have 

`let slot_duration = cumulus_client_consensus_aura::slot_duration(&*client)?;`

we should have

`let slot_duration = cumulus_client_consensus_aura::slot_duration(&*orchestrator_client)?;`

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -170,9 +170,7 @@ pub mod opaque {
 }
 
 impl_opaque_keys! {
-    pub struct SessionKeys {
-        pub aura: Aura,
-    }
+    pub struct SessionKeys { }
```
girazoki, line 173: do we need this anymore?

container-chains/templates/frontier/node/src/chain_spec.rs
```diff
@@ -202,7 +193,7 @@ pub fn local_testnet_config(para_id: ParaId) -> ChainSpec {
 }
 
 fn testnet_genesis(
-    invulnerables: Vec<(AccountId, NimbusId)>,
+    _invulnerables: Vec<(AccountId, NimbusId)>,
```
girazoki, line 196: I'd say lets remove this argument if not used, in both frontier and simple chain spec

girazoki (approved):

container-chains/templates/simple/runtime/src/lib.rs
nbaztec: This is being referenced in the impl block

```rust
impl sp_session::SessionKeys<Block> for Runtime {
        fn generate_session_keys(seed: Option<Vec<u8>>) -> Vec<u8> {
            SessionKeys::generate(seed)
        }

        fn decode_session_keys(
            encoded: Vec<u8>,
        ) -> Option<Vec<(Vec<u8>, sp_core::crypto::KeyTypeId)>> {
            SessionKeys::decode_into_raw_public_keys(&encoded)
        }
    }
```

container-chains/templates/frontier/node/src/chain_spec.rs
nbaztec: Yeah just wanted to see it compile. Will do :+1:

container-chains/templates/simple/runtime/src/lib.rs
girazoki: @tmpolaczyk suggested we could leave it !unimlemented() and pass parachain_config.dev_key_seed = None. But I dont really mind, we can merge the PR as is

container-chains/templates/frontier/node/src/chain_spec.rs
girazoki: @nbaztec I think you only needed to remove the `_invulnerables ` parameter

container-chains/templates/simple/runtime/src/lib.rs
tmpolaczyk: I'm fine either way, but we should add a comment saying that this is intentionally empty, because it is only required to satisfy a trait bound of `sc_service::spawn_tasks`, and session keys are not used in container chains.

PR #151, opened 2023-06-28
Move tests to different folder so that they are run in CI

tmpolaczyk:
And fix typo in "registrar"


girazoki:
probably a rebase-fault :(

girazoki (approved):

PR #157, opened 2023-07-04
Add proxy pallet to tanssi

tmpolaczyk:
Add proxy pallet to orchestrator runtime. These are the supported proxy types:

```rust
pub enum ProxyType {
    /// All calls can be proxied. This is the trivial/most permissive filter.
    Any = 0,
    /// Only extrinsics that do not transfer funds.
    NonTransfer = 1,
    /// Only extrinsics related to governance (democracy and collectives).
    Governance = 2,
    /// Only extrinsics related to staking.
    Staking = 3,
    /// Allow to veto an announced proxy call.
    CancelProxy = 4,
    /// Allow extrinsic related to Balances.
    Balances = 5,
}
```

Also add tests


runtime/dancebox/src/lib.rs
```diff
@@ -584,6 +604,89 @@ impl pallet_utility::Config for Runtime {
...
+)]
+pub enum ProxyType {
+    /// All calls can be proxied. This is the trivial/most permissive filter.
+    Any = 0,
+    /// Only extrinsics that do not transfer funds.
+    NonTransfer = 1,
+    /// Only extrinsics related to governance (democracy and collectives).
+    Governance = 2,
+    /// Only extrinsics related to staking.
+    Staking = 3,
```
girazoki, line 620: Not needed for now, nor governance

runtime/dancebox/src/lib.rs
```diff
@@ -543,6 +563,90 @@ impl pallet_utility::Config for Runtime {
...
+                    c,
+                    RuntimeCall::System(..)
+                        | RuntimeCall::ParachainSystem(..)
+                        | RuntimeCall::Timestamp(..)
+                        | RuntimeCall::Utility(..)
+                        | RuntimeCall::Proxy(..)
+                )
+            }
+            ProxyType::Governance => matches!(c, RuntimeCall::Utility(..)),
+            ProxyType::Staking => matches!(c, RuntimeCall::Utility(..)),
```
girazoki, line 608: I see polkadot has pallet session in Staking, I dont see why we should not

girazoki (approved):
If we can add rust tests for these, at least for a few more important examples, that would be great too. But overall LGTM

girazoki:
Can you please update the description of the PR? after that we can merge

PR #163, opened 2023-07-05
Remove Aura in orchestrator chain

nanocryk:


runtime/dancebox/src/lib.rs
```diff
@@ -613,14 +609,10 @@ construct_runtime!(
...
-        Authorship: pallet_authorship = 30,
-        CollatorSelection: pallet_collator_selection = 31,
-        Session: pallet_session = 32,
-        Aura: pallet_aura = 33,
-        AuraExt: cumulus_pallet_aura_ext = 34,
-        AuthorityMapping: pallet_authority_mapping = 35,
-
-        AuthorInherent: pallet_author_inherent = 50,
+        CollatorSelection: pallet_collator_selection = 30,
+        Session: pallet_session = 31,
```
girazoki, line 613: Does this change the well known key? (probably not because it is name based right?)

runtime/dancebox/src/lib.rs
```diff
@@ -629,11 +621,15 @@ construct_runtime!(
 impl_runtime_apis! {
     impl sp_consensus_aura::AuraApi<Block, NimbusId> for Runtime {
         fn slot_duration() -> sp_consensus_aura::SlotDuration {
-            sp_consensus_aura::SlotDuration::from_millis(Aura::slot_duration())
+            sp_consensus_aura::SlotDuration::from_millis(SLOT_DURATION)
         }
 
         fn authorities() -> Vec<NimbusId> {
-            Aura::authorities().into_inner()
+            pallet_authority_mapping::Pallet::<Runtime>::authority_id_mapping(Session::current_index())
```
girazoki, line 628:
```suggestion
           AuthorityMapping::authority_id_mapping(Session::current_index())
```

runtime/dancebox/tests/common/mod.rs
```diff
@@ -30,54 +31,55 @@ use {
...
-    AccountId, Aura, Authorship, Balance, Balances, Initializer, ParachainInfo, Registrar, Runtime,
-    RuntimeCall, RuntimeEvent, Session, System,
+    AccountId, Balance, Balances, Initializer, ParachainInfo, Registrar, Runtime, RuntimeCall,
+    RuntimeEvent, Session, System,
 };
 
-pub fn run_to_session(n: u32, add_author: bool) {
+pub fn run_to_session(n: u32) {
     let block_number = dancebox_runtime::Period::get() * n;
-    run_to_block(block_number + 1, add_author);
```
girazoki, line 42: we could keep the `add_author`:

- include the nimbus log only when `add_author` is true 
- run the kickoff authorship validation inherent only when  `add_author` is true

In the future we might add better testing for this

runtime/dancebox/src/lib.rs
girazoki, line 628: this should be `AuthorityAssignment(Session::current_index()).orchestrator` AFAICT

girazoki: it's not like we use this runtime API, and we could implement it the way you described too, but just to keep it consistent. What do you think?

runtime/dancebox/tests/common/mod.rs
nanocryk: I think it caused other `expect` to fail :/

girazoki: ok nvm, we can move it later

runtime/dancebox/src/lib.rs
```diff
@@ -309,8 +309,38 @@ impl pallet_timestamp::Config for Runtime {
...
+                let author_index = *slot % authorities.len() as u64;
+                return Some(author_index as u32);
+            }
+        }
+
+        None
+    }
+}
+
 impl pallet_authorship::Config for Runtime {
```
girazoki, line 342: pallet-authorship is removed from construct-runtime, so we can probably remove it too and forget about this

girazoki (approved):
small additional comment, but otherwise LGTM

runtime/dancebox/src/lib.rs
```diff
@@ -309,8 +309,38 @@ impl pallet_timestamp::Config for Runtime {
...
+        for (id, mut data) in digests.into_iter() {
+            if id == AURA_ENGINE_ID {
+                let slot = Slot::decode(&mut data).ok()?;
+
+                let authorities =
+                    AuthorityAssignment::collator_container_chain(Session::current_index())
+                        .expect("authorities should be set")
+                        .orchestrator_chain;
+
+                let author_index = *slot % authorities.len() as u64;
```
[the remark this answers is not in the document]
girazoki: should never happen, but we could add an early-exit if len is 0

runtime/dancebox/src/lib.rs
tmpolaczyk: It's unfortunate that we don't get a "unused code" warning

tmpolaczyk (approved):

runtime/dancebox/src/lib.rs
nanocryk: Keys are name based yeah, so indices are not important. We should probably cleanup the indices before a release.

girazoki: Yeah I am not too worried about the indices for now then. I think they are mostly clean the way they are, but if we want to change them, it will only affect in cases where other chains depend on this (which mostly I think its extrinsic-based, i.e., XCM only). So for now we should be on the safe side doing a release

PR #164, opened 2023-07-05
Dev service should run with arbitrary specs

girazoki:
makes sure the dev service runs with arbitrary specs and in particular that it runs with the dancebox-local specs, different to dev-service. We will later modify these to match whatever you launch in testnet.

For now only block creation is tested. Once we have official specs for our networks, the idea is that we test correctly certain parameteres of the genesis


girazoki:
For now we only check we can produce blocks, which I think it is sufficient

tmpolaczyk:
How did you generate the spec json file? We could have a script to make it more automatic.

tmpolaczyk (approved):

PR #167, opened 2023-07-06
same ci local zombienet setup

girazoki:
remove dual setup local-ci for zombienet tests. After this:

- specs are created in `test/tmp.`
- polkadot binary is downloaded in `test/tmp.`
- all other binaries stay in `target/release/`


test/configs/zombieTanssi.json
```diff
@@ -64,31 +64,31 @@
           {
             "name": "Collator2002-01",
             "command": "../target/release/tanssi-node",
-            "args": ["-- --tmp --prometheus-external --rpc-cors all --unsafe-rpc-external --rpc-methods unsafe --para-id 2002"]
+            "args": ["-- --tmp --para-id 2002"]
```
tmpolaczyk, line 67: This is not needed, the `--para-id` arg should only be used when running a full node, and this is a collator.

```suggestion
```

tmpolaczyk:
The test_tanssi_containers.ts test tries to read from both specs folders:

```
        // Read raw chain spec file
        // Different path in CI: ./specs vs ../specs
        let spec2002 = null;
        try {
          spec2002 = await fs.readFile("./specs/template-container-2002.json", "utf8");
        } catch {
          spec2002 = await fs.readFile("../specs/template-container-2002.json", "utf8");
        }
```

This should be

```
let spec2002 = await fs.readFile("./specs/template-container-2002.json", "utf8");
```

girazoki:
resolved

tmpolaczyk (approved):

PR #175, opened 2023-07-12
NIMBUS KEY ID instead of aura

girazoki:


client/consensus/src/consensus_orchestrator.rs
```diff
@@ -703,6 +704,8 @@ where
             .claim_slot(&slot_info.chain_head, slot, &aux_data)
             .await?;
 
+        log::info!("claim valid for slot {:?}", slot);
```
tmpolaczyk, line 707: Do we want to merge these logs as well?

girazoki: not really, thanks for the heads up

girazoki:
I left one new log indicating the validity of the claim

tmpolaczyk (approved):

PR #179, opened 2023-07-13
increment author noting tests

girazoki:
Increments author-noting test suite by adding some error-test

### DIFF
--- a/pallets/author-noting/src/test.rs
+++ b/pallets/author-noting/src/test.rs
@@ -15,8 +15,7 @@
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
 use {
-    crate::mock::*,
-    crate::Event,
+    crate::{mock::*, Event},
     cumulus_primitives_core::ParaId,
     frame_support::assert_ok,
     hex_literal::hex,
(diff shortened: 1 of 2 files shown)


pallets/author-noting/src/test.rs
```diff
@@ -445,3 +444,121 @@ fn test_set_author() {
...
+                        digest: sp_runtime::generic::Digest {
+                            logs: vec![DigestItem::PreRuntime(AURA_ENGINE_ID, slot.encode())],
+                        },
+                    });
+                sproof.items.push(s);
+            }
+            _ => unreachable!(),
+        })
+        .add(1, || {
+            assert_eq!(AuthorNoting::latest_author(ParaId::from(1001)), Some(13u64));
```
tmpolaczyk, line 474: This `add(` is not needed, right?

tmpolaczyk (approved):

pallets/author-noting/src/test.rs
girazoki: actually the add is needed but can be empty

tmpolaczyk (approved):

PR #180, opened 2023-07-13
try to ignore certain file patterns in coverage

girazoki:
The goal is to have a more accurate (even if lower) coverage. We remove the` tests.rs` and `mock.rs ` from the coverage, and also utilize the official weights instead of ()


tmpolaczyk:
cargo tarpaulin ignores tests by default, not sure how but may be worth taking a look.

tmpolaczyk (approved):

tmpolaczyk:
We probably want to ignore `weights.rs` and `tests` folders as well

girazoki:
Actually I dont think we want. I remember at some point we had a wrongly set weight to one of the extrinsics. So I am working on adding tests for every weight we associate in our pallets

PR #181, opened 2023-07-13
Turn MaxLengthTokenSymbol into a config param

tmpolaczyk:


primitives/container-chain-genesis-data/src/lib.rs
```diff
@@ -54,8 +56,19 @@ pub mod json;
...
+    DefaultNoBound,
+    PartialEqNoBound,
+    Encode,
+    Decode,
+    scale_info::TypeInfo,
+)]
+#[scale_info(skip_type_params(MaxLengthTokenSymbol))]
+pub struct ContainerChainGenesisData<MaxLengthTokenSymbol: Get<u32>> {
     pub storage: Vec<ContainerChainGenesisDataItem>,
     // TODO: make all these Vec<u8> bounded
```
tmpolaczyk, line 73: I want to add the other bounds in this PR as well, but I'm not sure how to do that. One option is to simply add them as additional generic params, but I don't like that because that will introduce a lot of noise every time this type is used:

```
ContainerChainGenesisData<MaxLengthTokenSymbol: Get<u32>, MaxLengthForkId: Get<u32>, ...>
```

The other option is to add one type parameter that can be used to get all the different bounds:

```
ContainerChainGenesisData<T: Config>
```

But I can't use the Config trait there, so I would probably need a new trait?

girazoki:
Let's keep that for another PR. This one does not look like it involves any migration to me, so let's keep it simple as it is

pallets/registrar/src/lib.rs
```diff
@@ -62,19 +62,23 @@ pub mod pallet {
...
 
     #[pallet::genesis_build]
-    impl<T: Config> GenesisBuild<T> for GenesisConfig {
+    impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
         fn build(&self) {
             // Sort para ids and detect duplicates, but do it using a vector of
             // references to avoid cloning the genesis data, which may be big.
             let mut para_ids: Vec<&_> = self.para_ids.iter().collect();
-            para_ids.sort();
+            para_ids.sort_by(|a, b| a.0.cmp(&b.0));
```
girazoki, line 81: was there any reason to change this?

girazoki (approved):
I think it looks good. We should remember that if we ever change this param, it might involve a migration

pallets/registrar/src/lib.rs
tmpolaczyk: Yes, because I had to remove the `Ord` and `PartialOrd` traits because substrate doesn't provide `OrdNoBounds` and `PartialOrdNoBounds`. That may be useful so I will open a feature request upstream.

PR #200, opened 2023-07-25
add more client-side tests

girazoki:
Adds tests for authorities-noting primitives client-side and consensus.

The consensus tests have a lot of boiler plate but it is the only way I found to propetly test this


client/consensus/src/consensus_orchestrator.rs
```diff
@@ -203,11 +201,7 @@ where
         Client:
             ProvideRuntimeApi<B> + BlockOf + AuxStore + HeaderBackend<B> + Send + Sync + 'static,
         AuthorityId<P>: From<<NimbusPair as sp_application_crypto::Pair>::Public>,
-        BI: BlockImport<B, Transaction = sp_api::TransactionFor<Client, B>>
```
girazoki, line 206: ParachainBlockImporter was an unneeded trait apparently

container-chains/primitives/authorities-noting-inherent/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+        .header(relay_genesis_hash)
+        .unwrap()
+        .unwrap();
+
+    // get latest header info, but for another paraId
+    let latest_header_info =
+        ContainerChainAuthoritiesInherentData::get_latest_orchestrator_head_info(
+            relay_header.hash(),
+            &relay_chain_interface,
+            (orch_para_id + 1).into(),
```
tmpolaczyk, line 341: This would be more clean as a new variable with a name:

```rust
let invalid_para_id = (orch_para_id + 1).into();
```

container-chains/primitives/authorities-noting-inherent/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+    fn new(session: u32) -> Self {
+        let builder = TestClientBuilder::new()
+            .add_extra_storage(
+                well_known_keys::SESSION_INDEX.to_vec(),
+                session.encode().to_vec(),
+            )
+            .add_extra_storage(
+                well_known_keys::authority_assignment_for_session(session).to_vec(),
+                VALUE.to_vec(),
+            );
```
tmpolaczyk, line 62: What's the purpose of this dummy storage? If I remove it all tests still pass.

client/consensus/Cargo.toml
```diff
@@ -45,3 +45,14 @@ futures = { workspace = true }
...
+[dev-dependencies]
+sc-network-test = { workspace = true }
+parking_lot = { workspace = true }
+sc-block-builder = { workspace = true }
+sp-keyring = { workspace = true }
+sc-keystore = { workspace = true }
+substrate-test-runtime-client = { workspace = true }
+tokio = { workspace = true }
+tempfile = { workspace = true }
+futures-timer = "3.0.1"
```
tmpolaczyk, line 58: Use `workspace = true`

container-chains/primitives/authorities-noting-inherent/src/tests.rs
```diff
@@ -0,0 +1,437 @@
...
+impl DummyRelayChainInterface {
+    fn new(orchestrator_para_id: ParaId, header: OrchestratorHeader) -> Self {
+        let builder = TestClientBuilder::new().add_extra_storage(
+            well_known_keys::para_id_head(orchestrator_para_id).to_vec(),
+            HeadData(header.encode()).encode().to_vec(),
+        );
+
+        Self {
+            relay_client: Arc::new(builder.build()),
+        }
```
tmpolaczyk, line 79:
```suggestion
    Self::new_with_head_data(orchestrator_para_id, HeadData(header.encode()).encode().to_vec())
```

client/consensus/src/tests.rs
```diff
@@ -0,0 +1,775 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>.
+
+// This tests have been greatly influenced by
+// https://github.com/paritytech/substrate/blob/master/client/consensus/aura/src/lib.rs#L832
+// Most of the items hereby added are intended to make it work with our current consensus mechanism
+
+#[cfg(test)]
+mod tests {
```
tmpolaczyk, line 22: This is redundant, this file is already called `tests.rs`

```suggestion
```

client/consensus/src/tests.rs
```diff
@@ -0,0 +1,775 @@
...
+
+    // Duration of slot time
+    const SLOT_DURATION_MS: u64 = 1000;
+
+    type Error = sp_blockchain::Error;
+
+    struct DummyFactory(Arc<TestClient>);
+    // We are going to create API because we need this to test runtime apis
+    // We use the client normally, bnut for testing certain runtime-api calls,
+    // we basically mock the runtime-api calñs
```
tmpolaczyk, line 75:
```suggestion
    // We are going to create API because we need this to test runtime apis
    // We use the client normally, but for testing certain runtime-api calls,
    // we basically mock the runtime-api calls
```

client/consensus/src/tests.rs
```diff
@@ -0,0 +1,775 @@
...
+                .sr25519_generate_new(NIMBUS_KEY_ID, Some(&key.to_seed()))
+                .expect("Creates authority key");
+            keystore_paths.push(keystore_path);
+
+            let environ = DummyFactory(client.clone());
+            import_notifications.push(
+                client
+                    .import_notification_stream()
+                    .take_while(|n| {
+                        future::ready(!(n.origin != BlockOrigin::Own && n.header.number() < &5))
```
tmpolaczyk, line 484: This line is hard to understand, and it doesn't match the comment above. The comment says that 

```
// The tests finishes when we see a block higher than block 5 that has
// not being authored by us
```

But the condition is "a block lower than block 5 that has not being authored by us".

Removing one layer of negation can make it easier to understand, assuming I did it correctly:

```suggestion
                        future::ready(n.origin == BlockOrigin::Own || n.header.number() >= &5)
```

But I still need to understand this test.

container-chains/primitives/authorities-noting-inherent/src/tests.rs
girazoki: yes you are right this is not needed (since we dont verify against anything this storage). This can be null. Thes session one cannot be null thought bcause we need it to construct the second key

client/consensus/src/tests.rs
girazoki: that is right, thank you

girazoki: actually the test is correct, is my comment which is wrong.

You have 3 authors, so you would expect to have a different author after 2 slots. I guess the magic number 5 has something to do with potential slot missing

girazoki:
@tmpolaczyk all comments fixed, can you re-take a look whenever you can?

client/consensus/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>.
+
+// This tests have been greatly influenced by
+// https://github.com/paritytech/substrate/blob/master/client/consensus/aura/src/lib.rs#L832
+// Most of the items hereby added are intended to make it work with our current consensus mechanism
+
+#[cfg(test)]
```
tmpolaczyk, line 21: This is also redundant

```suggestion
```

tmpolaczyk (approved):

PR #202, opened 2023-07-26
Add pallet-maintenance-mode and pallet-migrations to tanssi and both templates

tmpolaczyk:
Added tests to the common folder so that they run in both tanssi and frontier template.


girazoki:
I think we also want to add it to the templates. Templates will also likely need migrations.

girazoki:
We might also want to incorporate maintenance-mode, now that we are already bringing moonkit stuff. I know we had this in the list but and the sooner we incorporate it, the better I think

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -516,6 +520,101 @@ impl pallet_utility::Config for Runtime {
...
+    type MigrationsList = ();
+    type XcmExecutionManager = ();
+}
+
+/// Maintenance mode Call filter
+pub struct MaintenanceFilter;
+impl Contains<RuntimeCall> for MaintenanceFilter {
+    fn contains(c: &RuntimeCall) -> bool {
+        match c {
+            RuntimeCall::Balances(_) => false,
```
girazoki: We probably want to prevent the evm too here

runtime/dancebox/src/lib.rs
```diff
@@ -650,6 +653,108 @@ impl pallet_proxy::Config for Runtime {
...
+    type MigrationsList = ();
+    type XcmExecutionManager = ();
+}
+
+/// Maintenance mode Call filter
+pub struct MaintenanceFilter;
+impl Contains<RuntimeCall> for MaintenanceFilter {
+    fn contains(c: &RuntimeCall) -> bool {
+        match c {
+            RuntimeCall::Balances(_) => false,
```
girazoki, line 667: is proxy not in the list of moonbeam?

tmpolaczyk: Yes, I may have accidentally removed it

container-chains/templates/frontier/runtime/src/lib.rs
tmpolaczyk: Added evm but accidentally removed balances, will add it back

runtime/dancebox/src/lib.rs
tmpolaczyk: Actually no, proxy is not in the MaintenanceFilter list.

runtime/dancebox/src/lib.rs
```diff
@@ -650,6 +653,108 @@ impl pallet_proxy::Config for Runtime {
...
+                pallet_proxy::Call::kill_pure { .. } => false,
+                _ => true,
+            },
+            _ => true,
+        }
+    }
+}
+
+/// The hooks we want to run in Maintenance Mode
+pub struct MaintenanceHooks;
```
girazoki, line 695: You can probably avoid this until we have XCM.

JUst set `MaintenanceExecutiveHooks=AllPalletsWithSystem`

girazoki: same for template runtimes

girazoki (approved):

PR #205, opened 2023-07-31
Allow modularizing releases tanssi-templates

girazoki:
This pull request allows us to make releases for tanssi runtimes, templates, or both. The CI job now allows you to choose between releasing:

- `tanssi-only`
- `templates-only`
- `run-all`

The PR modifies the CI job such that it is quite easy to add a new template, since it only involves adding a new element to a json, formatted structure in the job prepare_matrix, e.g.:

`dancebox='{"runtime_name": "dancebox","runtime_path" : "./runtime/dancebox/src/lib.rs"}'
`
I also decoupled the draft release preparation into first pushing the release (non-matrixed) and then uploading the files (matrixed)

The scripts have been modified to support releasing both containers and tanssi runtimes too.

An example of a run only for templates can be seen in: https://github.com/moondance-labs/tanssi/releases/tag/untagged-07ed789acd3b36e2abed

Additionally we push the runtime-changes and tags not only for tanssi, but also for containers

**### How will releases look like from now on?**

Runtimes releases willl contain two tags:
- `runtime-xx.yy`
- `rutime-xx.yy-templates`

We have two different docker jobs for pushing template runtime releases or tanssi runtime releases. Ideally these will be the same (if no bug is encountered), but it might happen that either tanssi or one of the containers contains a bug and we have to do a minor release. In that case:

- If it's a tanssi minor release, we increment the tanssi spec, create the draft release only for tanssi (not the templates) by issuing a new tag `runtime-xx.yz` with the cherry-picked commit ,but without modifying the template tags. Then we just publish the tanssi-runtime release in docker.

- If it's a templates minor release, we increment the templates specs, create the draft release only for templates (not tanssi+) by issuing a new tag `runtime-xx.yz-templates` with the cherry-picked commit ,but without modifying the tanssi tags. Then we just publish the template-runtimes release in docker.

### DIFF
--- a/.github/workflows/publish-docker-runtime-containers.yml
+++ b/.github/workflows/publish-docker-runtime-containers.yml
@@ -1,4 +1,4 @@
-name: Publish Docker runtime
+name: Publish Docker runtime containers
 
 on:
   workflow_dispatch:
--- a/.github/workflows/publish-docker-runtime-containers.yml
+++ b/.github/workflows/publish-docker-runtime-containers.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: runtime tag (ex. runtime-2200) to publish on docker
+        description: runtime tag (ex. runtime-2200-templates) to publish on docker
         required: true
 
 jobs:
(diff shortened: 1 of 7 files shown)


girazoki:
Closing till I get something working, not willing to bug with commits

nbaztec (approved):
looks fine to me

PR #211, opened 2023-08-08
Separate spawning container chain and starting collation

tmpolaczyk:
This PR allows starting container chains that will start collating in the future, for example in the next session.

The main feature is the `partial_start_collator` function, which starts the collation subsystem for a container chain but doesn't enable it, leaving that to the caller. This works the same way as the `collate_on_tanssi` feature, which has already been implemented before.

Now we start container chains one session in advance, and start collating later when they we are actually assigned. 

Also changed the way `ContainerChainSpawner` works, instead of having `Start` and `Stop` messages, we now only have a `UpdateAssignment` message. This allows the `ContainerChainSpawner` to keep track of which container chains should be running or should be stopped, and which one should be collating.

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -317,7 +317,7 @@ impl ContainerChainSpawner {
             chains_to_stop,
             chains_to_start,
         } = handle_update_assignment_state_change(
-            &mut *self.state.lock().expect("poison error"),
+            self.state.lock().expect("poison error"),
             self.orchestrator_para_id,
             self.collate_on_tanssi.clone(),
             current,
@@ -442,11 +442,17 @@ mod tests {
             let currently_collating_on = Arc::new(Mutex::new(Some(orchestrator_para_id)));
             let currently_collating_on2 = currently_collating_on.clone();
             let collate_closure = move || async move {
-                let mut lco = currently_collating_on2.lock().unwrap();
-                // TODO: investigate why this assert fails
-                // TODO: found it, see comment in stop_collating_orchestrator
-                //assert_ne!(*lco, Some(orchestrator_para_id), "Received CollateOn message when we were already collating on this chain: {}", orchestrator_para_id);
-                *lco = Some(orchestrator_para_id);
+                let mut cco = currently_collating_on2.lock().unwrap();
+                // TODO: this sometimes fails, see comment in stop_collating_orchestrator
+                /*
+                assert_ne!(
+                    *cco,
+                    Some(orchestrator_para_id),
+                    "Received CollateOn message when we were already collating on this chain: {}",
+                    orchestrator_para_id
+                );
+                */
+                *cco = Some(orchestrator_para_id);
             };
             let collate_on_tanssi: Arc<
                 dyn Fn() -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync,
@@ -468,14 +474,17 @@ mod tests {
             let (signal, _on_exit) = exit_future::signal();
             let currently_collating_on2 = self.currently_collating_on.clone();
             let collate_closure = move || async move {
-                let mut lco = currently_collating_on2.lock().unwrap();
+                let mut cco = currently_collating_on2.lock().unwrap();
+                // TODO: this is also wrong, see comment in test fuzz2
+                /*
                 assert_ne!(
-                    *lco,
+                    *cco,
                     Some(container_chain_para_id),
                     "Received CollateOn message when we were already collating on this chain: {}",
                     container_chain_para_id
                 );
-                *lco = Some(container_chain_para_id);
+                */
+                *cco = Some(container_chain_para_id);
             };
             let collate_on: Arc<
                 dyn Fn() -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync,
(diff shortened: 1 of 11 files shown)


node/src/container_chain_spawner.rs
```diff
@@ -56,12 +56,20 @@ pub struct ContainerChainSpawner {
     pub spawn_handle: SpawnTaskHandle,
 
     // State
-    pub spawned_para_ids: Arc<Mutex<HashMap<ParaId, StopContainerChain>>>,
+    pub spawned_para_ids: Arc<Mutex<HashMap<ParaId, ContainerChainStuff>>>,
```
girazoki, line 59: Change this name please :)

node/src/service.rs
```diff
@@ -259,35 +261,29 @@ pub fn build_check_assigned_para_id(
...
-            // Spawn new container chain node
-            cc_spawn_tx.send(CcSpawnMsg::Spawn(new_para_id))?;
-        }
 
-        if let Some(prev_para_id) = *initial_assigned_para_id {
-            // Stop old container chain node
-            cc_spawn_tx.send(CcSpawnMsg::Stop(prev_para_id))?;
-        }
+    // Check assignment in the next session
+    let res2 = tc_consensus::first_eligible_key_next_session::<Block, ParachainClient, NimbusPair>(
```
tmpolaczyk, line 276: Here instead of calling two functions `first_eligible_key` and `first_eligible_key_next_session` we should call one that returns both

primitives/consensus/src/lib.rs
```diff
@@ -35,6 +35,9 @@ sp_api::decl_runtime_apis! {
 
         /// Returns the paraId for which an authority is assigned (if any)
         fn check_para_id_assignment(authority: AuthorityId) -> Option<ParaId>;
+
+        /// Returns the paraId for which an authority is assigned in the next session (if any)
+        fn check_future_para_id_assignment(authority: AuthorityId) -> Option<ParaId>;
```
tmpolaczyk, line 40: Not satisfied with this new method but it's probably the simplest solution, any better ideas?

girazoki: we can refactor it later but I think this is correct

node/src/service.rs
girazoki: yes, but we can do a refactor for this later

node/src/container_chain_spawner.rs
```diff
@@ -289,4 +301,67 @@ impl ContainerChainSpawner {
...
+        // State mutex cannot be used in the same scope as `.await`, so start a new scope here
+        {
+            let mut state = self.state.lock().expect("poison error");
+
+            if (state.assigned_para_id, state.next_assigned_para_id) == (current, next) {
+                // If nothing changed there is nothing to update
+                return;
+            }
+
+            running_chains_before.extend(state.assigned_para_id);
```
girazoki, line 320: Can you add comments here? tis a bit difficult to understand

node/src/service.rs
```diff
@@ -827,7 +875,103 @@ pub async fn start_node_impl_container(
...
     start_network.start_network();
 
-    Ok((task_manager, client))
+    Ok((task_manager, client, start_collation))
+}
+
+// Copy of `cumulus_client_service::start_collator`, that doesn't fully start the collator: it is
+// missing the final call to `cumulus_client_collator::start_collator`.
+// Returns the params of the call to `cumulus_client_collator::start_collator`.
+pub fn partial_start_collator<'a, Block, BS, Client, Backend, RCInterface, Spawner>(
```
girazoki, line 884: Hopefully we can in the future propose a pr to cumulus to make it more modular

girazoki: But maybe we can move it to a different file in a follow-up refactor

girazoki:
I dont see anything fundamentally wrong but I want to give a closer look

node/src/container_chain_spawner.rs
```diff
@@ -189,11 +207,23 @@ impl ContainerChainSpawner {
 
             // Signal that allows to gracefully stop a container chain
             let (signal, on_exit) = exit_future::signal();
+            let collate_on = collate_on.unwrap();
```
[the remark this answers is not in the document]
tmpolaczyk: Actually this was None when running a full node, fixed.

node/src/service.rs
tmpolaczyk: I think this is too specific to be accepted upstream, however we should try to add `Clone` to `StartCollatorParams`, it will save us some lines of code.

node/src/container_chain_spawner.rs
```diff
@@ -289,4 +301,67 @@ impl ContainerChainSpawner {
...
+                // If the assigned container chain was already running but not collating, we need to call collate_on
+                if let Some(para_id) = current {
+                    // Check if we get assigned to orchestrator chain
+                    if para_id == self.orchestrator_para_id {
+                        call_collate_on = Some(self.collate_on_tanssi.clone());
+                    } else {
+                        // When we get assigned to a different container chain, only need to call collate_on if it was already
+                        // running before
+                        if running_chains_before.contains(&para_id) {
+                            let c = state.spawned_container_chains.get(&para_id).unwrap();
```
girazoki, line 350: change unwrap for expect

node/src/container_chain_spawner.rs
```diff
@@ -259,26 +290,72 @@ impl ContainerChainSpawner {
...
-                        let f = (self.collate_on_tanssi)();
-                        f.await;
-                    } else {
-                        // Spawn new container chain node
-                        self.spawn_handle.spawn(
-                            "container-chain-spawner",
-                            None,
-                            self.spawn(para_id),
-                        );
+                CcSpawnMsg::UpdateAssignment { current, next } => {
```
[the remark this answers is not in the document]
tmpolaczyk: Done, now the tests are more complex than the code

girazoki (approved):

PR #212, opened 2023-08-08
Add xcm to runtimes

girazoki:
Add XCM to dancebox and templates, mainly config including:

- Pallets: PolkadotXCM and queues
- Ability to send a message to a remote chain and execute locally
- Acceptance of only the local token as a form of payment

RIght now  there are XCM tests in both rust (with emulator) and typescript (with manual-xcm). We will keep adding more tests as we evolve but I think these are a good set of tests that we can start with.

Missing: XCM pause on maintenance mode (might come on a follow-up)


tmpolaczyk:
There's a TODO for pallet_maintenance:

```
    // TODO: enable xcm-support feature when we enable xcm
```

container-chains/primitives/xcm/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+    frame_support::traits::OriginTrait,
+    sp_core::Get,
+    xcm::latest::{Junction::AccountKey20, MultiLocation, NetworkId},
+};
+
+// Convert a local Origin (i.e., a signed 20 byte account Origin)  to a Multilocation
+pub struct SignedToAccountKey20<Origin, AccountId, Network>(
+    sp_std::marker::PhantomData<(Origin, AccountId, Network)>,
+);
+impl<Origin: OriginTrait + Clone, AccountId: Into<[u8; 20]>, Network: Get<NetworkId>>
```
tmpolaczyk, line 29: Can you move this trait bounds to the where clause below? Maybe it will be easier to read

```suggestion
impl<Origin, AccountId, Network>
```

test/util/expect.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+  } else {
+    const foundEvents = block.result.events.filter(
+      ({ event }) => event.section.toString() == section && event.method.toString() == method
+    );
+    if (foundEvents.length > 0) {
+      events.push(...foundEvents);
+    }
+  }
+  expect(events.length > 0).to.not.be.null;
```
tmpolaczyk, line 143: I guess this file is copied from somewhere else, but this line seems useless, `events.length > 0` can only be true of false, and both of the are `!== null`. See https://www.chaijs.com/api/bdd/#method_null

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -0,0 +1,240 @@
...
+    // One XCM operation is 1_000_000_000 weight - almost certainly a conservative estimate.
+    pub UnitWeightCost: Weight = Weight::from_parts(1_000_000_000, 64 * 1024);
+
+    // TODO: revisit
+    pub const RelayNetwork: NetworkId = NetworkId::Polkadot;
+
+    // The relay chain Origin type
+    pub RelayChainOrigin: RuntimeOrigin = cumulus_pallet_xcm::Origin::Relay.into();
+
+    pub const MaxAssetsIntoHolding: u32 = 64;
```
Agusrodri, line 63: Shouldn't be better to use instead the `MAX_ASSETS` constant that comes from xcm_primitives? Just in case this changes in the future and so that we can have a better track of it.
```suggestion
    pub const MaxAssetsIntoHolding: u32 = xcm_primitives::MAX_ASSETS;
```

test/suites/common/xcm/test-mock-dmp-transact.ts
```diff
@@ -0,0 +1,120 @@
...
+        .push_any({
+            Transact: {
+            originType: "SovereignAccount",
+            requireWeightAtMost: new BN(1000000000),
+            call: {
+                encoded: transferCallEncoded,
+            },
+            },
+        })
+        .as_v2();
```
Agusrodri, line 103: I think it should be better to build these tests using xcm V3 from the beginning.

test/suites/common/xcm/test-mock-hrmp-transact.ts
```diff
@@ -0,0 +1,204 @@
...
+            let random: KeyringPair;
+            random = chain == 'frontier-template' ? generateKeyringPair() : generateKeyringPair("sr25519");
+
+            // Get Pallet balances index
+            const metadata = await polkadotJs.rpc.state.getMetadata();
+            const balancesPalletIndex = (metadata.asLatest.toHuman().pallets as Array<any>).find(
+            (pallet) => {
+                return pallet.name === "Balances";
+            }
+            ).index;
```
Agusrodri, line 152: This is a cleaner way to get pallet indices.
```suggestion
const balancesPalletIndex = metadata.asLatest.pallets
          .find(({ name }) => name.toString() == "Balances")!
          .index.toNumber();
```

Agusrodri (approved):
Some small suggestions but in general LGTM.

test/suites/common/xcm/test-mock-dmp-transact.ts
```diff
@@ -0,0 +1,120 @@
...
+        const { originAddress, descendOriginAddress } = descendFunction(context);
+
+        sendingAddress = originAddress;
+        descendAddress = descendOriginAddress;
+
+        transferredBalance = 10_000_000_000_000n;
+
+        const txSigned = polkadotJs.tx.balances.transfer(descendOriginAddress, transferredBalance);
+
+        await expectOk(
```
Agusrodri, line 43: `expectOk` is no needed here, if you want you can use the `{ allowFailures: false }` option inside `createBlock()` function instead.

test/suites/common/xcm/test-mock-dmp-transact.ts
```diff
@@ -0,0 +1,120 @@
...
+          context.createBlock(
+            await txSigned.signAsync(alice)
+          )
+        );
+        const balanceSigned = (
+          (await polkadotJs.query.system.account(descendOriginAddress)) as any
```
Agusrodri, line 49: This cast to `any` is not needed here (and also inside `test-mock-hrmp-transact.ts` file where similar queries are made).

test/suites/common/xcm/test-mock-dmp-transact.ts
girazoki: I know but I did not have your PR yet :), I might do it now

PR #213, opened 2023-08-10
Delete container chain db when stopping, and use warp sync to sync container chains

tmpolaczyk:
Container chain collators don't need all the blocks from the container chains, so they will use warp sync to download the latest state, and delete the database after rotating away.

In this PR:

* If running as collator, the database is deleted when stopping the container chain gracefully (when the collator is unassigned or the node process is stopped by the operator), and it is also deleted right before starting (warp sync only works with an empty database, otherwise it falls back to normal sync). If running as full node, nothing is deleted.

* Force container chains to always use warp sync, ignoring the `--sync` CLI param. Except when the container chain is still at genesis, then it uses full sync because warp sync doesn't work.

* This uses the warp sync from upstream, modified so that it only needs 1 peer to start syncing instead of 3, and it downloads the full block history in the background.

TODO:

* Allow users to override sync param
* Add CLI param to never delete the database
* Delete empty parent folders along with db (see TODO in delete_container_chain_db)


girazoki:
Cant we incorporate your commit?

girazoki:
its a good sign that they are failing btw :), that means we are really using it

tmpolaczyk:
Exactly, I wanted to see what happens with the default implementation. Will cherry-pick the branch with the min peers set to 1 now.

node/src/container_chain_spawner.rs
```diff
@@ -172,6 +172,11 @@ impl ContainerChainSpawner {
             db_path.set_file_name(format!("full-container-{}", container_chain_para_id));
             container_chain_cli_config.database.set_path(&db_path);
 
+            // Delete existing database
+            if db_path.exists() {
+                std::fs::remove_dir_all(&db_path).expect("failed to remove old container chain db");
```
tmpolaczyk, line 177: This path is

```
Collator2002-01/data/containers/chains/simple_container_2002/db/full-container-2002
```

But we probably want to delete the parent folder:

```
Collator2002-01/data/containers/chains/simple_container_2002
```

But be careful to not delete more than we should, check where does this `simple_container_2002/db` come from

tmpolaczyk:
CI doesn't pass because I edited Cargo.lock in a way that's not allowed (to avoid pushing a commit to our tanssi-polkadot-v0.9.43 branch), but you can try it locally and zombienet tests work now.

node/src/container_chain_spawner.rs
```diff
@@ -173,6 +177,29 @@ impl ContainerChainSpawner {
             // Update CLI params
             container_chain_cli.base.para_id = Some(container_chain_para_id.into());
 
+            // Force container chains to use warp sync
+            // If the container chain is still at genesis block, use full sync because warp sync is broken
+            let container_chain_is_at_genesis = if !orchestrator_runtime_api
```
girazoki, line 182: I would call this variable `full_sync_needed`, because `container_chain_is_at_genesis` does not accurately represent when the runtime-api is not there (the container-chain might not be at genesis here)

test/configs/zombieTanssiWarpSync.json
```diff
@@ -0,0 +1,104 @@
...
+            "name": "FullNode-2000",
+            "validator": false,
+            "command": "../target/release/container-chain-template-simple-node",
+            "ws_port": 9949,
+            "p2p_port": 33049
+          }
+        ]
+      },
+      {
+        "id": 2001,
```
girazoki, line 84: We dont need to containrs to check this right? what about just using one and potentially running these in ubuntu-latest?

Orchestrator: A,B ,E
Container2000: C,D

Delete D, E should go to container2000 with warp sync

tmpolaczyk: Correct, container 2001 is unused, I can remove it.

tmpolaczyk:
Removed container chain 2001 from warp-sync tests, that needed a custom build-spec-warp-sync.sh script because the registered chains are defined there. An alternative would have been to unregister chain 2001 as soon as possible, but I believe Collator1000-03 would still be assigned there, and we don't want that.

girazoki:
you mean we only need one registration? yeah that is fine if we just generate a new build-spec script

girazoki (approved):

PR #214, opened 2023-08-14
ApiAugment package refactor and workspaces

timbrinded:
This PR introduces a few new things:

- PNPM workspaces to the repo which will reduce complexity between tests and ApiAugment
   - The test directory automatically symlinks to apiAugment folder in the repo
   - To regenerate a local copy of the ApiAugment (e.g. you make a change in a pallet and want to write a test for it), regenerate the interface with: `pnpm create-local-interfaces`
- Api Augment build process is also a lot simpler now, and benefits from type safety
- Publishing to NPM uses a whitelist instead of the previous system of manually copying individual files and creating a package out of that.
- Moves the rust test out of the build step, since `cargo test` actually recompiles the binary


girazoki:
I am susprised about the drop in the test percentage rate. Is there any reason explaining it?

girazoki:
Actually it might be because this PR nor the base PR is not rebased with latest master. Will fix that once we merge this one into the other

typescript-api/scripts/postbuild.ts
```diff
@@ -0,0 +1,33 @@
...
+  // const pck = JSON.parse(readFileSync(path.join(process.cwd(), "package.json"), "utf-8"));
+  const buildPath = `${process.env.PWD}/build`;
+
+  // pck.scripts = {};
+  // pck.private = false;
+  // pck.type = "module";
+  // pck.files = ["**/*", "!**/tsconfig.tsbuildinfo", "!**/*.tgz"];
+
+  // console.log(`Writing ${buildPath}/package.json`);
+  // writeFileSync(`${buildPath}/package.json`, JSON.stringify(pck, null, 2));
```
girazoki, line 16: Are there commentefd for some reason?

girazoki (approved):
Looks good to me! I think its a massive improvement over what we had :)

typescript-api/scripts/postbuild.ts
timbrinded: Yes, because before this script was trying to handcraft a package.json via find and replace. Instead we should just use a package manager like god would intend.

girazoki:
@timbrinded can you merge master in?

PR #215, opened 2023-08-16
Change AuthorNoting to include block number, and add runtime api

tmpolaczyk:
Change type of `pallet_author_noting::LatestAuthor` storage item from `AccountId` to `(BlockNumber, AccountId)`. This is because we want to store the container chain block number along with the author.

Add runtime api that allows to query latest author and latest block number for a container chain.

### DIFF
--- a/pallets/author-noting/src/lib.rs
+++ b/pallets/author-noting/src/lib.rs
@@ -153,8 +153,8 @@ pub mod pallet {
             let total_weight =
                 T::WeightInfo::set_latest_author_data(registered_para_ids.len() as u32);
             for para_id in registered_para_ids {
-                match Self::fetch_author_slot_from_proof(&relay_storage_rooted_proof, para_id) {
-                    Ok(author) => LatestAuthor::<T>::insert(para_id, author),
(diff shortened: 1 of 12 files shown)


pallets/author-noting/src/lib.rs
```diff
@@ -195,14 +197,15 @@ pub mod pallet {
...
+            block_number: BlockNumber,
             new_author: T::AccountId,
         },
     }
 
     #[pallet::storage]
     #[pallet::getter(fn latest_author)]
     pub(super) type LatestAuthor<T: Config> =
-        StorageMap<_, Blake2_128Concat, ParaId, T::AccountId, OptionQuery>;
+        StorageMap<_, Blake2_128Concat, ParaId, (BlockNumber, T::AccountId), OptionQuery>;
```
girazoki, line 208: I would make a new struct instead of a tuple, I think it makes things more clear. Something like containerChainBlockInfo

girazoki:
Reasons why I dont think we need migration:

- This storage item is not read anywhere but in the runtime-api
- As soon as one block is created then the storage should be updated correctly
- Previous node versions should stay compatible with this change (and we can always deploy the runtime-api reading node later)

girazoki (approved):

PR #219, opened 2023-08-22
"Pooled Staking" pallet

nanocryk:
[README](https://github.com/moondance-labs/tanssi/blob/jeremy-pooled-staking/pallets/pooled-staking/README.md)

 ## Storage Items

- `SortedEligibleCandidates`: Stores a sorted list of eligible candidates. Is bounded by `T::EligibleCandidatesBufferSize`, which may lead eligible candidates to be out of the list and not be accounted for elections even if higher candidates leave. They will appear again in the list after any stake change, and an extrinsic could allow to update them without requiring stake changes.
- `Pools`: DoubleMap storing balances for all pools. First key is the candidate AccountId, which helps querying all the staking information for a given candidate (for UI). The second key is an enum type `PoolsKey<A>` representing the various balances that needs to be stored:
  - `CandidateTotalStake`: total amount of stake for that candidate across all pools
  - `JoiningShares { delegator: A }`: amount of joining shares this delegator have towards this candidate.
  - `JoiningSharesSupply`: supply of this pool shares for that candidate.
  - `JoiningSharesTotalStaked`: total amount of stake/currency in that pool. Collateral for all the shares of that pool.
  - `JoiningSharesHeldStake { delegator: A }`: amount of currency held in the delegator account for that pool and candidate. Can mismatch the value of the share the delegator owns in case of slashing or rewarding, which can be rebalanced with `rebalance_hold` extrinsic.
  - The above 4 keys also exists with prefix `AutoCompounding`, `ManualRewards` and `Leaving`.
  - `ManualRewardsCounter`: amount of stake rewarded **per share** to delegators of this candidate since genesis.
  - `ManualRewardsCheckpoint { delegator: A }`: value of the counter when the delegator last claimed rewards or changed their stake. The difference between counter and checkpoint allow to distribute the correct amount of rewards.
- `PendingOperations`: DoubleMap storing pending operations. First key is the delegator AccountId, which helps querying all the pending operations for a user (for UI). The second key is an enum type `PendingOperationKey<A, B>` representing the various operations:
  - `JoiningAutoCompounding { candidate: A, at: B }`: delegator requested to join the auto-compounding pool of given candidate. The request was made at height `at`, whose type is customizable to support both block height and session index. The value at that key is the amount of **joining shares** that will be transformed into auto compounding shares once the operation can be executed. The amount is in **joining shares** instead of currency to allow slashing.
  - `JoiningManualRewards { .. }`: same as above for manual rewards shares.
  - `Leaving { .. }`: same for leaving staking. Candidate gets **leaving shares** that can be redeemed for currency once the operation can be executed.
  
## Extrinsics
  - `rebalance_hold(origin, candidate, delegator, pool)`: With rewards and slashing the value of shares in a given pool can differ from the amount help in the delegator account for those shares. This extrinsic allow to transfer from/to the delegator account the difference. The main use case is to allow the delegator to participate in other pallets such as democracy with its stake. Auto compounding rewards are distributed indirectly using the share system, and the actual currency is held in the staking pallet account. This extrinsic can be used to transfer it to the user account (with a hold to prevent spending) so that this new stake can be used too in democracy. This extrinsic can be called by anyone on anyone else's shares to allow collecting slashed currency still held in the delegators accounts.
  - `request_delegate(origin, candidate, pool, stake)`: Request to delegate `stake` (currency amount) in the `pool` of given `candidate`. Delegator gets **joining shares** and a **joining operation** is registered. `stake` (minus shares rounding errors) are **held** in the delegator account.
  - `request_undelegate(origin, candidate, pool, amount)`: Request to undelegate `amount` (expressed either in currency or share amount) from the `pool` of given `candidate`. The `pool` shares of that delagator are removed, they get **leaving shares** and a **leaving operation** is registered. Share rounding error is released immediately.
  - `execute_pending_operations(origin, operations)`: Execute provided `operations` expressed as a `PendingOperationQuery`, which contains the delegator AccountId and a `PendingOperationKey`. Anyone can execute operations of anyone.


pallets/pooled-staking/src/traits.rs
```diff
@@ -0,0 +1,100 @@
...
+        Error::MathUnderflow
+    }
+}
+
+pub trait MulDiv: Sized {
+    /// Multiply self by `a` then divide the result by `b`.
+    /// Computation will be performed in a bigger type to avoid overflows.
+    /// After the division, will return `None` if the result is to big for
+    /// the real type or if `b` is zero.
+    fn mul_div(self, a: Self, b: Self) -> Result<Self, OverflowError>;
```
tmpolaczyk, line 45: There's a similar trait in substrate: [sp_arithmetic::rational::MultiplyRational](https://docs.rs/sp-arithmetic/latest/sp_arithmetic/rational/trait.MultiplyRational.html)

The implementation looks a bit different, so if you decide to use it make sure to test edge cases. For instance, `multiply_rational` returns `None` on `0 / 0` while `mul_div` returns `0`.

pallets/pooled-staking/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+
+[dependencies]
+log = { workspace = true }
+serde = { workspace = true, optional = true }
+
+# Substrate
+frame-benchmarking = { workspace = true, optional = true }
+frame-support = { workspace = true }
+frame-system = { workspace = true }
+# substrate-fixed = { git = "https://github.com/encointer/substrate-fixed", default-features = false }
```
girazoki, line 20: Do you not need this?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+            }
+        };
+
+        // Transfer is done using withdraw to ensure it works regardless of ED.
+        if let Some(diff) = stake.0.checked_sub(&held.0) {
+            T::Currency::transfer(
+                &T::StakingAccount::get(),
+                &delegator,
+                dbg!(diff),
+                Preservation::Preserve,
```
girazoki, line 84: This means we rely on the staking account to have ED right?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+            )?;
+            T::Currency::hold(&T::CurrencyHoldReason::get(), &delegator, diff)?;
+            return Ok(().into());
+        }
+
+        if let Some(diff) = held.0.checked_sub(&stake.0) {
```
girazoki, line 90: Can this be an else of the previous branch? it seems the way it is right now both can be true, but it is actually not possible right?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+                let held = pools::Leaving::<T>::hold(&candidate, &delegator);
+                let shares = pools::Leaving::<T>::shares(&candidate, &delegator);
+                let stake = pools::Leaving::<T>::shares_to_stake(&candidate, shares)?;
+                pools::Leaving::<T>::set_hold(&candidate, &delegator, stake.clone());
+                (held, stake)
+            }
+        };
+
+        // Transfer is done using withdraw to ensure it works regardless of ED.
+        if let Some(diff) = stake.0.checked_sub(&held.0) {
```
girazoki, line 79: If diff is 0 maybe we should do nothing? did we test this case?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        // We hold the funds of the delegator and register its stake into the candidate stake.
+        T::Currency::hold(&T::CurrencyHoldReason::get(), &delegator, stake.0)?;
+        pools::Joining::<T>::increase_hold(&candidate, &delegator, &stake)?;
+        Candidates::<T>::add_total_stake(&candidate, &stake)?;
+
+        // We create/mutate a request for joining.
+        let now = frame_system::Pallet::<T>::block_number();
+        let operation_key = match pool {
+            TargetPool::AutoCompounding => PendingOperationKey::JoiningAutoCompounding {
```
girazoki, line 137: This means I can either select autoCompounding when I delegate or manual rewards, but not a percentage for each of them right? Just wondering how easy would be to add this in the future

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+    },
+    frame_system::pallet_prelude::*,
+    sp_runtime::traits::{CheckedSub, Zero},
+};
+
+pub struct Calls<T>(PhantomData<T>);
+
+impl<T: Config> Calls<T> {
+    pub fn rebalance_hold(
+        _: OriginFor<T>,
```
tmpolaczyk, line 42: Is this safe? In another call we have

```
        // We don't care about the sender.
        let _ = ensure_signed(origin)?;
```

but here we don't even check if it's signed

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+        let removed_stake = match pool {
+            TargetPool::AutoCompounding => {
+                let stake = pools::AutoCompounding::<T>::sub_shares(
+                    &candidate,
+                    &delegator,
+                    Shares(shares),
+                )?;
+
+                if stake.0 > pools::AutoCompounding::<T>::hold(&candidate, &delegator).0 {
+                    Self::rebalance_hold(
```
girazoki, line 193: this basically means that the hold was not well balanced and as such we need to rebalance it before we decrease it correct?

pallets/pooled-staking/src/calls.rs
girazoki: I guess one can always do 2 delegations targeting different pools?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+                continue;
+            }
+
+            match &operation {
+                PendingOperationKey::JoiningAutoCompounding {
+                    candidate,
+                    at_block,
+                } => {
+                    ensure!(
+                        T::JoiningRequestFilter::can_be_executed(&candidate, &delegator, *at_block),
```
girazoki, line 299: Is the goal that we set JoiningRequestFilter to have the delay of at least 1 session?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+                    ensure!(
+                        T::LeavingRequestFilter::can_be_executed(&candidate, &delegator, *at_block),
+                        Error::<T>::RequestCannotBeExecuted(index as u16)
+                    );
+
+                    Self::execute_leaving(candidate.clone(), delegator.clone(), Shares(value))?;
+                }
+            }
+
+            PendingOperations::<T>::set(&delegator, &operation, Zero::zero());
```
girazoki, line 339: Could we simply kill the entry?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+            TargetPool::AutoCompounding => {
+                pools::AutoCompounding::<T>::stake_to_shares_or_init(&candidate, stake.clone())?
+            }
+            TargetPool::ManualRewards => {
+                pools::ManualRewards::<T>::stake_to_shares_or_init(&candidate, stake.clone())?
+            }
+        };
+
+        // If stake doesn't allow to get at least one share we release all the funds.
+        if Zero::is_zero(&shares.0) {
```
girazoki, line 366: If I understand correctly, this is because the stake was held before executing this operation (because this comes from a pending operation), and therefore if you cannot create at least one share you need to release it?

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,459 @@
...
+            T::Currency::release(
+                &T::CurrencyHoldReason::get(),
+                &delegator,
+                stake.0,
+                Precision::Exact,
+            )?;
+            return Ok(().into());
```
girazoki, line 373: shouldnt we do also `Candidates::<T>::sub_total_stake(&candidate, Stake(stake.0))` here?;

pallets/pooled-staking/src/traits.rs
nanocryk: Thanks will fix this edgecase. I didn't find this trait in Substrate, but we don't need the rounding complexity, only something that behaves like `a * b / c` while preventing overflow on `a * b`.

pallets/pooled-staking/src/calls.rs
nanocryk: I couldn't find a way to ignore ED, so yeah the staking account should have at least the ED when no one is staking.

pallets/pooled-staking/src/candidate.rs
```diff
@@ -0,0 +1,199 @@
...
+
+pub struct Candidates<T>(PhantomData<T>);
+
+impl<T: Config> Candidates<T> {
+    pub fn total_stake(candidate: &Candidate<T>) -> Stake<T> {
+        Stake(Pools::<T>::get(candidate, &PoolsKey::CandidateTotalStake))
+    }
+
+    pub fn add_total_stake(candidate: &Candidate<T>, stake: &Stake<T>) -> Result<(), Error<T>> {
+        if Zero::is_zero(&stake.0) {
```
girazoki, line 66: maybe we can implement the Zero trait to stake?

pallets/pooled-staking/src/calls.rs
nanocryk: Not sure if it is necessary to check the origin is signed if we don't care about who it is. We should check in substrate pallet calls that don't care about the sender what they do.

pallets/pooled-staking/src/calls.rs
nanocryk: We should check before if `stake.0 == held.0` indeed.

pallets/pooled-staking/src/calls.rs
nanocryk: Yes but we still need to get `diff`. An `else if` is also not required as the previous `if` ends with a `return`.

pallets/pooled-staking/src/calls.rs
nanocryk: One would make 2 delegation requests, with the amounts being split according to the percentage you want.
We can add another extrinsic `request_delegate_split` that takes such percentage and make the 2 requests. We could also add a new request kind for that but that may introduce more complexity.

pallets/pooled-staking/src/calls.rs
nanocryk: Exactly. Delegator "received" auto rewards but those are in the staking account without a rebalance. If the delegator wants to unstake all its stake + those rewards there is not enough held for that.

Note that this should only be necessary for auto compounding, since manual rewards are .. not compounding :p So this kind of "positive imbalance" should not occur (only negative on with slashing).

pallets/pooled-staking/src/calls.rs
nanocryk: Exactly. This trait allows to use delays in blocks or session boundaries. We may remove `candidate` and `delegator` arguments tho.

pallets/pooled-staking/src/calls.rs
nanocryk: We could, is it better than setting to zero?

pallets/pooled-staking/src/calls.rs
girazoki: I guess it depends if None origins pay fees, or whether we want them to pay fees. I have the impression  that something we might want to do is:

- In the joining/leaving extrinsics, account for fees related to this extrinsics plus the execute-pending-operations ones.
- in the execute-pending-operations extrinsic we might not charge for fees (as it was accounted before)

is rebalance-hold an extrinsic by itself? if so I think we need people that call it to pay fees, because there is no-where where we can account for this right?

pallets/pooled-staking/src/calls.rs
girazoki: we will do it if we find the need in the future, I was just wondering if it was possible. Right now the way it is its good, thanks!

pallets/pooled-staking/src/calls.rs
nanocryk: When creating the request we cannot anticipate the value of a share due to rewards and slashing. So it could be that a user make a request for an amount that is now worth 1 share (+ dust) and that at execution is worth less than 1.

pallets/pooled-staking/src/calls.rs
nanocryk: Indeed, nice catch. I have to add tests for each of those cases.

pallets/pooled-staking/src/calls.rs
nanocryk: Can a user call an extrinsic without an origin? The rationale here is that anyone can call this extrinsic, especially in case of slashing to get the funds out of the delegators accounts.

We can have an `ensure_signed` in the `pallet::call` section and don't have an origin parameter in `call::Calls`.

tmpolaczyk: > Can a user call an extrinsic without an origin?

Yes, according to @girazoki that's what happens when you press the "send unsigned" button in polkadot js. In that case we need custom logic to handle fees, so I don't think we want to support unsigned transactions here.

pallets/pooled-staking/src/pools.rs
```diff
@@ -0,0 +1,396 @@
...
+    AutoCompoundingSharesTotalStaked,
+    AutoCompoundingSharesHeldStake,
+    T::InitialAutoCompoundingShareValue::get(),
+);
+
+impl_pool!(
+    ManualRewards,
+    ManualRewardsShares,
+    ManualRewardsSharesSupply,
+    ManualRewardsTotalStaked,
```
tmpolaczyk, line 314: This name is inconsistent with `AutoCompoundingSharesTotalStaked`

```suggestion
    ManualRewardsSharesTotalStaked,
```

pallets/pooled-staking/src/calls.rs
nanocryk: Added unit tests.

pallets/pooled-staking/src/pools.rs
nanocryk: Fixed

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,473 @@
...
+            .map_err(|_| Error::<T>::MathOverflow)?;
+        PendingOperations::<T>::set(&delegator, &operation_key, operation);
+
+        // We release the dust if non-zero.
+        let dust = removed_stake
+            .0
+            .err_sub(&leaving_stake.0)
+            .map_err(Error::<T>::from)?;
+
+        if dust.is_zero() {
```
tmpolaczyk, line 252: The comment above says

```
// We release the dust if non-zero.
```

So I guess this should be negated?

```suggestion
        if !dust.is_zero() {
```

If so, add a test.

pallets/pooled-staking/src/candidate.rs
```diff
@@ -0,0 +1,182 @@
...
+    candidate: C,
+    stake: S,
+}
+
+impl<C: Ord, S: Ord> Ord for EligibleCandidate<C, S> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.stake
+            .cmp(&other.stake)
+            .reverse()
+            .then(self.candidate.cmp(&self.candidate))
```
tmpolaczyk, line 46: This is comparing self to self, should be other.

Also this can benefit from lazy evaluation, because the stake amount is unlikely to be equal (using `then(` will always compare it, `and_then` only if previous condition is equal):

```suggestion
            .and_then(|| self.candidate.cmp(&other.candidate))
```

pallets/pooled-staking/src/candidate.rs
nanocryk: I would avoid implementing all arithmetic traits on those. Their goal is just to be wrappers to avoid confusion between amounts in currency or shares.

pallets/pooled-staking/src/tests.rs
```diff
@@ -0,0 +1,664 @@
...
+                },
+            ]);
+        })
+    }
+);
+
+pool_test!(
+    fn delegation_request_more_than_available<P>() {
+        ExtBuilder::default().build().execute_with(|| {
+            let amount = DEFAULT_BALANCE + 1;
```
tmpolaczyk, line 361: If I change this to `DEFAULT_BALANCE` it also says "FundsUnavailable", I guess because of existential deposit being set to 1. Is that expected? `DEFAULT_BALANCE - 1` works as expected.

pallets/pooled-staking/src/calls.rs
girazoki: yes exactly, users will have a clear incentive to do this, hence I would keep it behind a signed origin

pallets/pooled-staking/src/calls.rs
girazoki: I guess for storage purposes it is.

girazoki:
@nanocryk I think we are ready to try to inject it into the runtime itself and make integrations tests on it. It should be quite self-contained for now (just check that joining, leaving etc works), and then we will wire it to the collator selection stuff.

We can even help doing a PR above this one for this. But it would be good if you could help us installing the palelt in the runtime already

pallets/pooled-staking/README.md
```diff
@@ -0,0 +1,39 @@
...
+  (decided in advance so that anyone can trigger the convertion).
+- **Auto compounding pool**: The delegator is eligible to rewards which are automatically compounded. This is
+  done by increasing the total amount of stake backing the pool without changing the amount of shares owned, which indirectly increase the value of each share.
+- **Manual rewards pool**: The delegator is eligible to rewards which are kept out of the pool. It is based
+  around a counter of how much reward has been distributed per share since genesis. For each delegator is stored the value of the counter when they joined the pool or last claimed, such that it is possible to compute the amount of withdrawable rewards based on the amount of owned shares. Any change of the amount
+  of shares of a delegator (joining/leaving) requires to force claiming the rewards to keep the calculations
+  correct.
+- **Leaving pool**: The delegator requested to stop delegating for that candidate. However they are still
+  accountable if the candidate is slashed until the end of the leaving delay. They no longer count towards
+  the candidate score nor are eligible to rewards.
```
notlesh, line 29: Where do the rewards for this stake go? Are they effectively not emitted?

nanocryk: Rewards will be distributed according to the distribution of auto compounding shares and manual rewards shares only.

pallets/pooled-staking/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+
+# Substrate
+frame-benchmarking = { workspace = true, optional = true }
+frame-support = { workspace = true }
+frame-system = { workspace = true }
+
+# Nimbus
+nimbus-primitives = { workspace = true }
+parity-scale-codec = { version = "3.0.0", default-features = false, features = [ "derive" ] }
+scale-info = { version = "2.0", default-features = false, features = [ "derive" ] }
```
tmpolaczyk, line 24: Use `{ workspace = true }`

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,479 @@
...
+            AllTargetPool::Leaving => {
+                let held = pools::Leaving::<T>::hold(&candidate, &delegator);
+                let shares = pools::Leaving::<T>::shares(&candidate, &delegator);
+                let stake = pools::Leaving::<T>::shares_to_stake(&candidate, shares)?;
+                pools::Leaving::<T>::set_hold(&candidate, &delegator, stake.clone());
+                (held, stake)
+            }
+        };
+
+        if stake == held {
```
tmpolaczyk, line 76: Now that we don't have rewards or slashing, this will always be true, right?

pallets/pooled-staking/src/pools.rs
```diff
@@ -0,0 +1,393 @@
...
+        if shares.0.is_zero() {
+            return Ok(Stake(Zero::zero()));
+        }
+
+        Self::shares_to_stake(candidate, shares)
+    }
+
+    /// Convert an amount of staked currency to an amount of shares for given candidate.
+    /// If this candidate have no shares then it uses `initial_share_value` to compute the value.
+    fn stake_to_shares_or_init(
```
tmpolaczyk, line 97: From the name of the function I expected that this would write the default value to storage if it doesn't exist (init = initialize), but it instead returns a default value (initial). How about this name instead?

```suggestion
    fn stake_to_shares_or_default(
```

tmpolaczyk: Actually, do we need this function? Any reason why can't `stake_to_shares` handle this case?

pallets/pooled-staking/src/pools.rs
```diff
@@ -0,0 +1,393 @@
...
+                delegator: delegator.clone(),
+            },
+        );
+
+        // TODO: Should be safe to wrap around.
+        let diff = counter.err_sub(&checkpoint)?;
+        Ok(Stake(diff.err_mul(&shares.0)?))
+    }
+
+    pub fn claim_rewards(
```
tmpolaczyk, line 349: This function and `pending_rewards` are unused, so maybe remove them until they are needed? That way we can review them better later.

pallets/pooled-staking/src/pools.rs
```diff
@@ -0,0 +1,393 @@
...
+            fn set_shares_supply(candidate: &Candidate<T>, value: Shares<T>) {
+                Pools::<T>::set(candidate, &PoolsKey::$supply, value.0)
+            }
+
+            fn set_total_staked(candidate: &Candidate<T>, value: Stake<T>) {
+                Pools::<T>::set(candidate, &PoolsKey::$total, value.0)
+            }
+
+            fn set_hold(candidate: &Candidate<T>, delegator: &Delegator<T>, value: Stake<T>) {
+                Pools::<T>::set(
```
tmpolaczyk, line 273: For each candidate that joins and then leaves, the storage grows a few bytes. All these functions should kill the storage item when the value is 0. But not sure if that solves the problem because an attacker could simply leave some dust and pollute the storage anyway. Do you think this is worth it?

And does anyone know if this being a `ValueQuery` means that a value of `0` is handled specially? Because it doesn't seem like that in tests, but in production it may be different.

girazoki:
One thing that just occured to me is that we probably need to check if a candidate has an authoring key registered before making it a candidate. This should be similar to how we do it for collatorSelection:

https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/pallets/collator-selection/src/lib.rs#L434

pallets/pooled-staking/src/pools.rs
nanocryk: We want to use this default value only when we create shares. When we want to know how much shares/stake a delegator have, it should not default to this default amount.

pallets/pooled-staking/src/pools.rs
nanocryk: This part should be reviewed now as they deal with the counter/checkpoint system. If there is an issue in there it will require massive migrations to fix the checkpoints for all delegators. I'll write tests to ensure they work properly.

nanocryk:
> One thing that just occured to me is that we probably need to check if a candidate has an authoring key registered before making it a candidate. This should be similar to how we do it for collatorSelection:
> 
> https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/pallets/collator-selection/src/lib.rs#L434

I can add a filter trait in the Config that allows to set additional constraints to be considered eligible, which can then be used for this purpose and others.

pallets/pooled-staking/src/calls.rs
nanocryk: Indeed

pallets/pooled-staking/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        /// Condition for when a joining request can be executed.
+        type JoiningRequestFilter: Contains<Self::BlockNumber>;
+        /// Condition for when a leaving request can be executed.
+        type LeavingRequestFilter: Contains<Self::BlockNumber>;
+        /// All eligible candidates are stored in a sorted list that is modified each time
+        /// delegations changes. It is safer to bound this list, in which case eligible candidate
+        /// could fall out of this list if they have less stake than the top `EligibleCandidatesBufferSize`
+        /// eligible candidates. One of this top candidates leaving will then not bring the dropped candidate
+        /// in the list. An extrinsic is available (TODO) to manually bring back such dropped candidate.
```
tmpolaczyk, line 258: This extrinsic is `update_candidate_position`, right? 
```suggestion
        /// in the list. An extrinsic is available to manually bring back such dropped candidate.
```

pallets/pooled-staking/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+        TryingToLeaveTooSoon,
+        InconsistentState,
+        UnsufficientSharesForTransfer,
+        CandidateTransferingOwnSharesForbidden,
+        RequestCannotBeExecuted(u16),
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        pub fn rebalance_hold(
```
tmpolaczyk, line 417: Shouldn't extrinsics have call_index and weight?

pallets/pooled-staking/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+use frame_support::pallet;
+
+pub use pallet::*;
+
+#[pallet(dev_mode)]
```
tmpolaczyk, line 51:
```suggestion
```

Remember to remove this when benchmarks are ready

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,526 @@
...
+        candidate: Candidate<T>,
+        delegator: Delegator<T>,
+        pool: TargetPool,
+        amount: SharesOrStake<T::Balance>,
+    ) -> DispatchResultWithPostInfo {
+        // Converts amount to shares of the correct pool
+        let shares = match (amount, pool) {
+            (SharesOrStake::Shares(s), _) => s,
+            (SharesOrStake::Stake(s), TargetPool::AutoCompounding) => {
+                pools::AutoCompounding::<T>::stake_to_shares(&candidate, Stake(s))?.0
```
tmpolaczyk, line 178: This rounds down to the number of shares, so it is possible for this method to return `Ok` when the `amount` is greater than the available amount:

67 shares at 1000000 per share = 67000000, but calling request_undelegate with 67999999 succeeds.

Obviously the returned amount is 67000000, but it may be confusing that the call succeeded.

runtime/dancebox/src/lib.rs
```diff
@@ -739,6 +741,53 @@ impl pallet_maintenance_mode::Config for Runtime {
 
 impl pallet_root_testing::Config for Runtime {}
 
+parameter_types! {
+    pub StakingAccount: AccountId32 = PalletId(*b"POOLSTAK").into_account_truncating();
+    pub const CurrencyHoldReason: [u8; 8] = *b"POOLSTAK";
+    pub const InitialManualClaimShareValue: u128 = currency::KILODANCE;
+    pub const InitialAutoCompoundingShareValue: u128 = currency::KILODANCE;
+    pub const InitialJoiningShareValue: u128 = currency::DANCE;
+    pub const InitialLeavingShareValue: u128 = currency::DANCE;
```
tmpolaczyk, line 750: The doc comment says

```
        /// Should be 1 to avoid rounding. (configurable for tests)
```

But currency::DANCE is not 1, it is 1_000_000_000_000.

```suggestion
    pub const InitialJoiningShareValue: u128 = 1;
    pub const InitialLeavingShareValue: u128 = 1;
```

pallets/pooled-staking/src/lib.rs
nanocryk: Currently the pallet is in `dev_mode`, which allows to skip adding the weights. This should be removed and benchmark should be run.

pallets/pooled-staking/src/calls.rs
nanocryk: We should document that it is **up to** the provided amount, but it may use less.

runtime/dancebox/src/lib.rs
nanocryk: Those parameters were mostly added to be able to properly test rounding. If you see another way to customize the value in tests we could remove those type parameters.

tmpolaczyk: > If you see another way to customize the value in tests we could remove those type parameters.

Idk, maybe adding a storage item only for testing like we sometimes do for benchmarks, but for now you can leave it like this.

pallets/pooled-staking/src/calls.rs
```diff
@@ -0,0 +1,522 @@
...
+        stake: T::Balance,
+    ) -> DispatchResultWithPostInfo {
+        ensure!(!stake.is_zero(), Error::<T>::StakeMustBeNonZero);
+
+        // Convert stake into joining shares quantity.
+        let shares = pools::Joining::<T>::stake_to_shares_or_init(&candidate, Stake(stake))?;
+
+        // If the amount was stake and is less than the value of 1 share it will round down to
+        // 0 share. We avoid doing any work for 0 shares.
+        ensure!(!shares.0.is_zero(), Error::<T>::StakeMustBeNonZero);
```
rimbi, line 125: is the error returned correct? Should it include `Share` instead of `Stake`?

tmpolaczyk: `stake = shares * share_price` so it's not wrong, but I agree that it is a bit unintuitive that as a user you can pass a non-zero stake value and see this error.

pallets/pooled-staking/src/lib.rs
girazoki: I can remove the dev_mode there

pallets/pooled-staking/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+    /// `MaxCollatorSetSize`.
+    #[pallet::storage]
+    pub type SortedEligibleCandidates<T: Config> = StorageValue<
+        _,
+        BoundedVec<
+            candidate::EligibleCandidate<Candidate<T>, T::Balance>,
+            T::EligibleCandidatesBufferSize,
+        >,
+        ValueQuery,
+    >;
```
tmpolaczyk, line 278: It's surprising that this compiles when `candidate::EligibleCandidate` is a private type, I made it public in my PR #242

girazoki (commented):
We also need to add PooledStaking to Staking Proxy in the runtime

girazoki (approved):

PR #221, opened 2023-08-23
Add pallet-invulnerables (#163)

rimbi:
This pallet mimics what we do in collator-selection:
- A storage item that holds the accounts that are invulnerables
- A extrinsic that updates this storage item via root origin

We would also need to include a migration to migrate those invulnerables from pallet-collator-selection to pallet-invulnerables.


girazoki:
@rimbi Can you provide a description of what the PR does?

pallets/invulnerables/src/migration.rs
```diff
@@ -0,0 +1,91 @@
...
+/// Version 1 Migration
+/// This migration ensures that any existing `Invulnerables` storage lists are sorted.
+pub mod v1 {
+    use super::*;
+    use frame_support::pallet_prelude::*;
+    #[cfg(feature = "try-runtime")]
+    use sp_std::prelude::*;
+
+    pub struct MigrateToV1<T>(sp_std::marker::PhantomData<T>);
+    impl<T: Config> OnRuntimeUpgrade for MigrateToV1<T> {
```
girazoki, line 31: I dont understand this migraiton. At the beginning there wont be any invulnerables. What we need to do is set them for the first time, but that has nothing to do with the storage version of the pallet (as it did not change). You can remove this fiole for now, as we need to do the migration in some other way

rimbi: I assumed we would work on it together. You can ignore it for now.

girazoki:
At some point I think it might be good to offer this pallet-abstraction to cumulus, leaving this comment here to remember :)

runtime/dancebox/src/migrations.rs
```diff
@@ -0,0 +1,96 @@
...
+        assert_eq!(
+            number_of_invulnerables, stored_invulnerables_len,
+            "after migration, there should be the same number of invulnerables"
+        );
+
+        Ok(())
+    }
+}
+
+pub struct CommonMigrations<Runtime>(PhantomData<Runtime>);
```
girazoki, line 84:
```suggestion
pub struct DanceboxMigrations<Runtime>(PhantomData<Runtime>);
```

runtime/dancebox/src/lib.rs
```diff
@@ -669,7 +679,7 @@ impl pallet_proxy::Config for Runtime {
 
 impl pallet_migrations::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
-    type MigrationsList = ();
+    type MigrationsList = (migrations::CommonMigrations<Runtime>,);
```
girazoki, line 682:
```suggestion
    type MigrationsList = (migrations::DanceboxMigrations<Runtime>,);
```

runtime/dancebox/src/migrations.rs
rimbi: done

runtime/dancebox/src/lib.rs
rimbi: done

runtime/dancebox/tests/integration_test.rs
```diff
@@ -1983,3 +1986,34 @@ fn check_well_known_keys() {
...
+        ])
+        .with_config(pallet_configuration::HostConfiguration {
+            max_collators: 100,
+            min_orchestrator_collators: 2,
+            max_orchestrator_collators: 2,
+            collators_per_container: 2,
+        })
+        .build()
+        .execute_with(|| {
+            let invulnerables_collator_selection = CollatorSelection::invulnerables();
```
girazoki, line 2010: One suggestion now that we got this test working: remove the

```
.with_collators(vec![
            (AccountId::from(ALICE), 210 * UNIT),
            (AccountId::from(BOB), 100 * UNIT),
        ])
```

and use `put_storage_value` like in https://github.com/moonbeam-foundation/moonbeam/blob/446faeb04153d127211a499244c900a1a5c71e54/pallets/asset-manager/src/tests.rs#L468. That way we are ready to remove the CollatorSelection in the following PR without having to change the test

girazoki (approved):
A small comment regarding the migration test, but otherwise looking good!

runtime/dancebox/tests/integration_test.rs
rimbi: done

PR #222, opened 2023-08-23
Define freeze and hold params (#126)

rimbi:
Define freeze and hold params with proper types.


girazoki:
its Ok if you dont modify those in the mock.rs files :)

girazoki:
We should add an integration test in dancebox making sure you cannot go over maxFreezes and maxHolds. I think it should be as easy as creating in a for loop 50 number of holds/freezes and check that the 51 fails

girazoki:
Actually I am thinking that we might want to set MaxFreezes to 0 and maxHolds to 1. I know there will be one hold identifier from pallet-staking, so lets do that. In the case of the containers for now, lets put 0 and 0

girazoki (approved):

PR #223, opened 2023-08-23
Fix ethereum gas pricing

girazoki:
Brings https://github.com/paritytech/frontier/pull/1115 and sets the correct value in the max block gas and miliseconds per block, so that gas calculation is correct with respect to weight

### DIFF
--- a/container-chains/templates/frontier/runtime/src/lib.rs
+++ b/container-chains/templates/frontier/runtime/src/lib.rs
@@ -341,7 +341,7 @@ const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
     cumulus_primitives_core::relay_chain::MAX_POV_SIZE as u64,
 );
 
-/// We allow for 2000ms of compute with a 6 second average block time.
+/// We allow for 500ms of compute with a 12 second average block time.
 pub const WEIGHT_MILLISECS_PER_BLOCK: u64 = 500;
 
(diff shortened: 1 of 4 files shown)


container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -652,7 +637,7 @@ impl pallet_dynamic_fee::Config for Runtime {
 }
 
 parameter_types! {
-    pub DefaultBaseFeePerGas: U256 = U256::from(1_000_000_000);
+    pub DefaultBaseFeePerGas: U256 = U256::from(2_000_000_000);
```
girazoki, line 640: This internal gets multiplied by the ideal target fullness (50%) to calculate the minimum gas price below which we cannot drop

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -344,7 +342,7 @@ const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
 );
 
 /// We allow for 2000ms of compute with a 6 second average block time.
```
tmpolaczyk, line 344:
```suggestion
/// We allow for 500ms of compute with a 6 second average block time.
```

And is 6 second block time correct?

girazoki: it should be 12

tmpolaczyk (approved):

PR #229, opened 2023-08-25
Girazoki allow to choose whether publish as latest

girazoki:
When having a new runtime for container-chains (not a node) we should allow to push as latest, as the dapp will need this to actually run correctly the proper runtime


girazoki:
Closing as we might not want this for now, pending conversation with ops

PR #233, opened 2023-08-25
Dancebox chopsticks test

girazoki:
Adds fork tests for stagenet-dancebox and dancebox with our own chopsticks version. We test we can upgrade the chain and additionally do balance transfer.

**There is likely** something that we still cannot correctly set, and this is the next authorities. Mainly because they are indexed by sessionId, and we cannot mock those in advance. One solution to this would be:

- mock the current session id
- mock authorities for this session id and the next one
- mock collators identically


pallets/author-noting/src/lib.rs
```diff
@@ -139,27 +139,33 @@ pub mod pallet {
...
-                    Err(e) => log::warn!(
-                        "Author-noting error {:?} found in para {:?}",
-                        e,
-                        u32::from(para_id)
-                    ),
+
+            // We do this first to make sure we dont do 2 reads (parachains and relay state)
+            // when we have no containers registered
+            // Essentially one can pass an empty proof if no container-chains are registered
+            if !registered_para_ids.is_empty() {
```
tmpolaczyk, line 149: So if this is empty we don't validate the `data`, but that's fine because we don't write it anywhere. Ok looks good.

girazoki: yes, I thought this was needed for the chopsticks test (as we inject an empty validation data) but apparently its not needed. So it is up to us to keep it...I dont see a reason why we should not tbh

runtime/dancebox/src/lib.rs
```diff
@@ -792,7 +792,19 @@ impl_runtime_apis! {
         }
 
         fn authorities() -> Vec<NimbusId> {
-            pallet_authority_assignment::CollatorContainerChain::<Runtime>::get(Session::current_index())
+
+            // Check whether we need to fetch the next authorities or current ones
```
girazoki, line 796: Modified this to actually match what the chain expects, which is, always the set of authorities that has to build the next orchestrator block

tmpolaczyk: We have a few integration tests that use this value and they didn't fail after this change, can you add a test for this edge case?

girazoki: they dont use this value, they use the other api (I believe it is `TanssiAuthorityAssignmentApi`). We maintain this api for two reasons though:
- because we are still using aura, and it might be expected by certain tools (i.e., chopsticks)
- because it implementes the slot_duration function (which we could also move to the other runtime api, but I think it does not hurt to maintain this on)

girazoki: I updated this one because chopsticks has to be a bit more generic, and therefore if it detects that aura consensus is being used, it queries this api

.github/workflows/release.yml
```diff
@@ -413,6 +413,47 @@ jobs:
...
+        with:
+          ref: ${{ needs.set-tags.outputs.git_ref }}
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20.x
+          cache: "pnpm"
+          cache-dependency-path: test/pnpm-lock.yaml
```
timbrinded, line 441: with workspaces (which is what tanssi uses) you dont need to add the direct path

PR #234, opened 2023-08-28
Migrate pallet-transaction-payment

tmpolaczyk:


girazoki:
We can probably close this one as I dont think the migration is needed

tmpolaczyk:
Closing, we could do a migration to v2 but it is currently not needed

PR #235, opened 2023-08-28
Remove CollatorSelection pallet (#170)

rimbi:
After adding `pallet-invulnerables`, as the next step, this change set removes the `pallet-collator-selection` and makes tanssi-node use the `pallet-invulnerables` to manage invulnerable collators.


pallets/invulnerables/src/lib.rs
```diff
@@ -142,9 +163,39 @@ pub mod pallet {
...
+                    None => {
+                        Self::deposit_event(Event::InvalidInvulnerableSkipped {
+                            account_id: account_id.clone(),
+                        });
+                        continue;
+                    }
+                }
+
+                new_with_keys.push(account_id.clone());
+            }
```
girazoki, line 193: You also need to add this in the add_invulnerable extrinsic

pallets/invulnerables/src/mock.rs
```diff
@@ -91,11 +91,19 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
     let invulnerables = vec![2, 1]; // unsorted
 
+    let balances = vec![(1, 100), (2, 100), (3, 100), (4, 100), (5, 100)];
+	let keys = balances
+		.iter()
+		.map(|&(i, _)| (i, i, MockSessionKeys { aura: UintAuthorityId(i) }))
+		.collect::<Vec<_>>();
+	let session = pallet_session::GenesisConfig::<Test> { keys };
+	session.assimilate_storage(&mut t).unwrap();
+fff
```
girazoki, line 101:
```suggestion
```

pallets/invulnerables/src/lib.rs
```diff
@@ -56,12 +60,26 @@ pub mod pallet {
         /// Overarching event type.
         type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
 
+        /// The currency mechanism.
+        type Currency: ReservableCurrency<Self::AccountId>;
```
girazoki, line 64: We should not need the Currency type I think

pallets/invulnerables/src/lib.rs
```diff
@@ -56,12 +60,26 @@ pub mod pallet {
...
+        /// A stable ID for a validator.
+        type ValidatorId: Member + Parameter;
+
+        /// A conversion from account ID to validator ID.
+        ///
+        /// Its cost must be at most one storage read.
+        type ValidatorIdOf: Convert<Self::AccountId, Option<Self::ValidatorId>>;
+
+        /// Validate a user is registered
+        type ValidatorRegistration: ValidatorRegistration<Self::ValidatorId>;
```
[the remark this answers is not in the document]
rimbi: done

pallets/invulnerables/src/lib.rs
rimbi: done

pallets/invulnerables/src/mock.rs
rimbi: done

pallets/invulnerables/src/lib.rs
rimbi: done

Cargo.lock
```diff
@@ -18,7 +18,16 @@ version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
 dependencies = [
- "gimli",
```
tmpolaczyk, line 21: Not sure why this `Cargo.lock` has so many changes, can you try copying the one from master and compiling again? I do it like this:

```
git checkout origin/master -- Cargo.lock
cargo test
git commit -am 'Update Cargo.lock'
```

pallets/invulnerables/Cargo.toml
```diff
@@ -12,15 +12,20 @@ targets = ["x86_64-unknown-linux-gnu"]
 [dependencies]
 log = { workspace = true }
 parity-scale-codec = { workspace = true }
+rand = { version = "0.8.5", features = ["std_rng"], default-features = false }
```
girazoki, line 15: rand should be marked as optional since AFAIK is only used in benchmarks. It should be only incorporated with runtime-benchmarks

girazoki: Same for pallet-session and pallet-balances

tmpolaczyk: And it should use workspace = true:

```
rand = { workspace = true, features = ["std_rng"], optional = true }
```

Cargo.lock
rimbi: It does not work, unfortunately.

pallets/invulnerables/Cargo.toml
rimbi: done except `pallet_session` which is used in the pallet to implement `SessionManager`.

Cargo.lock
tmpolaczyk: Worked for me, fixed

pallets/invulnerables/src/lib.rs
```diff
@@ -142,12 +164,39 @@ pub mod pallet {
...
 
-            // should never fail since `new` must be equal to or shorter than `TooManyInvulnerables`
-            let mut bounded_invulnerables = BoundedVec::<_, T::MaxInvulnerables>::try_from(new)
-                .map_err(|_| Error::<T>::TooManyInvulnerables)?;
+            let mut new_with_keys = Vec::new();
+
+            // check if the invulnerables have associated validator keys before they are set
+            for account_id in &new {
+                // don't let one unprepared collator ruin things for everyone.
+                let validator_key = T::CollatorIdOf::convert(account_id.clone());
```
tmpolaczyk, line 172: Is the validator_key the same as the "CollatorId"? If not this shouldn't be called `validator_key`.

pallets/invulnerables/src/lib.rs
```diff
@@ -62,6 +72,17 @@ pub mod pallet {
         /// Maximum number of invulnerables.
         type MaxInvulnerables: Get<u32>;
 
+        /// A stable ID for a collator.
+        type CollatorId: Member + Parameter;
+
+        /// A conversion from account ID to collator ID.
+        ///
+        /// Its cost must be at most one storage read.
+        type CollatorIdOf: Convert<Self::AccountId, Option<Self::CollatorId>>;
```
tmpolaczyk, line 81: What's the purpose of this? It is set to "IdentityCollator", which does nothing because `CollatorId` is defined to be `AccountId`.

pallets/invulnerables/src/lib.rs
```diff
@@ -142,12 +164,39 @@ pub mod pallet {
...
+                        Self::deposit_event(Event::InvalidInvulnerableSkipped {
+                            account_id: account_id.clone(),
+                        });
+                        continue;
+                    }
+                }
 
-            // Invulnerables must be sorted for removal.
-            bounded_invulnerables.sort();
+                new_with_keys.push(account_id.clone());
```
tmpolaczyk, line 193: This loop would be easier to follow without the `continue`s, there are only 3 paths:

```
match key {
    Some(key) if T::CollatorRegistration::is_registered(&key) => {
        // Good, push key
    }
    Some(key) => {
        // Bad, key not registered
    }
    None => {
        // Bad, key does not exist
    }
}
```

pallets/invulnerables/src/lib.rs
```diff
@@ -171,10 +220,10 @@ pub mod pallet {
             T::UpdateOrigin::ensure_origin(origin)?;
 
             <Invulnerables<T>>::try_mutate(|invulnerables| -> DispatchResult {
-                match invulnerables.binary_search(&who) {
-                    Ok(_) => return Err(Error::<T>::AlreadyInvulnerable)?,
-                    Err(pos) => invulnerables
-                        .try_insert(pos, who.clone())
+                match invulnerables.contains(&who) {
+                    true => return Err(Error::<T>::AlreadyInvulnerable)?,
+                    false => invulnerables
```
tmpolaczyk, line 225: Use if/else

pallets/invulnerables/src/lib.rs
```diff
@@ -213,4 +263,31 @@ pub mod pallet {
...
     }
+
+    use pallet_session::SessionManager;
+    use sp_staking::SessionIndex;
+
+    /// Play the role of the session manager.
+    impl<T: Config> SessionManager<T::AccountId> for Pallet<T> {
+        fn new_session(index: SessionIndex) -> Option<Vec<T::AccountId>> {
+            log::info!(
+                "assembling new invulnerable collators for new session {} at #{:?}",
```
tmpolaczyk, line 274: We will see this log on every session, right? It shouldn't say new if the invulnerables did not change, that may be confusing. And since pallet_collator_selection did not print any logs, maybe we should remove it or move it to debug level.

```suggestion
                "Assembling invulnerable collators for new session {} at #{:?}",
```

pallets/invulnerables/src/mock.rs
```diff
@@ -74,28 +78,128 @@ impl system::Config for Test {
...
+
+pub struct TestSessionHandler;
+impl pallet_session::SessionHandler<u64> for TestSessionHandler {
+    const KEY_TYPE_IDS: &'static [sp_runtime::KeyTypeId] = &[UintAuthorityId::ID];
+    fn on_genesis_session<Ks: OpaqueKeys>(keys: &[(u64, Ks)]) {
+        SessionHandlerCollators::set(keys.iter().map(|(a, _)| *a).collect::<Vec<_>>())
+    }
+    fn on_new_session<Ks: OpaqueKeys>(_: bool, keys: &[(u64, Ks)], _: &[(u64, Ks)]) {
+        SessionChangeBlock::set(System::block_number());
+        dbg!(keys.len());
```
tmpolaczyk, line 149:
```suggestion
```

pallets/invulnerables/src/lib.rs
```diff
@@ -213,4 +263,31 @@ pub mod pallet {
             Ok(())
         }
     }
+
+    use pallet_session::SessionManager;
+    use sp_staking::SessionIndex;
```
tmpolaczyk, line 268: Move these imports to the top of the file

pallets/invulnerables/src/lib.rs
girazoki: I think we should name it `collator_id`

pallets/invulnerables/src/lib.rs
girazoki: we did the same that cumulus does, because yes it does nothing but it could be that collatorId is different to accountId. We can remove it if we are sure we will not change it..+

pallets/invulnerables/src/lib.rs
rimbi: done

pallets/invulnerables/src/lib.rs
rimbi: This is the exact behavior in the `collator-selection` pallet. Even if there is no new collators, it signals the new session.

pallets/invulnerables/src/lib.rs
rimbi: does that make any difference?

pallets/invulnerables/src/lib.rs
rimbi: done

pallets/invulnerables/src/lib.rs
rimbi: This was the implementation from the original code. But you are right. I have simplified it in another way.

pallets/invulnerables/src/mock.rs
rimbi: done

pallets/invulnerables/src/lib.rs
tmpolaczyk: It makes the code easier to read for me, if you're asking about performance it is the same.

rimbi: can we keep it for the sake of diversity? :)

pallets/invulnerables/src/lib.rs
tmpolaczyk: You are right that this log is copied from `collator-selection`, but when I run tanssi-node from master I don't see that log. Any ideas why? Maybe because it's from another crate? Or we didn't update invulnerables on every new session on master?

rimbi: checking...

rimbi: It is there:
```
2023-08-31 08:18:49 assembling new collators for new session 0 at #0
2023-08-31 08:18:49 assembling new collators for new session 1 at #0
```

rimbi: May be you looked for the exact string `assembling new invulnerable collators for new session`?

tmpolaczyk: Thank you, I see it now, indeed I was looking for "new invulnerable". We can keep this log as is then.

pallets/invulnerables/src/lib.rs
rimbi: the majority won, done. :)

tmpolaczyk (approved):

girazoki (approved):

PR #236, opened 2023-08-28
Use FeeMultiplierUpdate = 1 instead of default of 0

tmpolaczyk:
Previously we charged `weight * 0` fee, now we charge `weight * 1`.

Also added a test to check that proofSize is not part of the fee.


test/suites/dev-tanssi/fees/test_fee_balance_transfer.ts
```diff
@@ -41,7 +36,7 @@ describeSuite({
 
         const events = await polkadotJs.query.system.events();
         const fee = extractFee(events).amount.toBigInt();
-        const expectedFee = 1000000n + BigInt(signedTx.encodedLength);
+        const expectedFee = 1000000n + BigInt(signedTx.encodedLength) + 1630678n;
```
tmpolaczyk, line 39: I hardcoded this value here assuming that it is supposed to be fixed, like ethereum transfer which always costs 21000 gas, if it is variable we can estimate it using some rpc method.

girazoki: we probably can yes. I think if we use `extractInfo` from moonwall we can get the `dispatchInfo`, which will contain the weight used. Then we can use the runtimeAPi `transactionPaymentApi->WeightToFee` to know how much fee would correspond to the weight part

PR #239, opened 2023-09-01
Allows the container chain registrar to set boot nodes for the chain (#178)

rimbi:
Allows the container chain registrar to set boot nodes for the chain as well as root.

### DIFF
--- a/pallets/registrar/src/lib.rs
+++ b/pallets/registrar/src/lib.rs
@@ -41,17 +41,19 @@ pub use pallet::*;
 
 #[frame_support::pallet]
 pub mod pallet {
-    use sp_runtime::traits::BadOrigin;
 
     use {
         crate::weights::WeightInfo,
         frame_support::{
             pallet_prelude::*,
-            traits::{Currency, ReservableCurrency},
+            traits::{Currency, EitherOfDiverse, ReservableCurrency},
             DefaultNoBound, LOG_TARGET,
         },
-        frame_system::pallet_prelude::*,
-        sp_runtime::{traits::AtLeast32BitUnsigned, Saturating},
+        frame_system::{pallet_prelude::*, EnsureSigned},
+        sp_runtime::{
+            traits::{AtLeast32BitUnsigned, BadOrigin},
+            Either, Saturating,
+        },
         sp_std::prelude::*,
         tp_container_chain_genesis_data::ContainerChainGenesisData,
         tp_traits::{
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
Add another test that checks the error case, Bob trying to set bootnodes for Alice's parachain

pallets/registrar/src/lib.rs
```diff
@@ -419,7 +421,14 @@ pub mod pallet {
             para_id: ParaId,
             boot_nodes: BoundedVec<BoundedVec<u8, T::MaxBootNodeUrlLen>, T::MaxBootNodes>,
         ) -> DispatchResult {
-            T::RegistrarOrigin::ensure_origin(origin)?;
+            let res = T::RegistrarOrigin::ensure_origin(origin.clone());
+            if res.is_err() {
+                let origin = ensure_signed(origin)?;
+                let deposit_info = RegistrarDeposit::<T>::get(para_id).ok_or(BadOrigin)?;
```
tmpolaczyk, line 427: This is a new db read so we will need to update the benchmarks. Not sure if it's possible but we want a different weight depending on the origin. If not, the benchmark should test the worst case, which is a signed origin (instead of root).

girazoki: lets account for worst case yes

rimbi: updated the benchmark.

tmpolaczyk (approved):

pallets/registrar/src/lib.rs
girazoki: I think there is a more substrate-friendly way of doing is. I propose two options:
- we use ensure_signed_or_root()  (and not registrarOrigin) for this one, and if the result is not none(), then we check that signer == depositor

- We add a new origin that is called SetBootnodesOrigin, which we can set to EitherOfDiverse<EnsureRoot, EnsureSIgned>. After this, same procedure as before would apply

rimbi: I don't think the first one is a better solution as it ignores `RegistrarOrigin`. The second solution is applicable but degrades the readability IMHO:
```rust
            let origin =
                EitherOfDiverse::<T::RegistrarOrigin, EnsureSigned<T::AccountId>>::ensure_origin(
                    origin,
                )?;

            if let sp_runtime::Either::Right(caller) = origin {
                let deposit_info = RegistrarDeposit::<T>::get(para_id).ok_or(BadOrigin)?;
                if deposit_info.creator != caller {
                    Err(BadOrigin)?;
                }
            }
```
What do you think?

rimbi: Btw, it's good to be aware of `EitherOfDiverse`, thanks!

girazoki: `RegistrarOrigin` can be changed to something like `MarkValidForCollation` origin, which basically would refer to the origin able to mark chains as valid for collation.

The problem that I see right now is that you are defining a registrarOrigin for the extrinsic, but wait, if its an error, we apply some other filter. This to me does not capture very well and its difficult to read.

PR #240, opened 2023-09-01
Bring benchmarks to pallet staking

girazoki:
Important:
- I made it such that executing pending requests does not incur into fee payment, as we are charging this extrinsics weight in the request_delegate and request_undelegate extrinsics.

We can change this if we want


pallets/pooled-staking/src/lib.rs
```diff
@@ -440,16 +450,20 @@ pub mod pallet {
...
         pub fn execute_pending_operations(
             origin: OriginFor<T>,
             operations: Vec<PendingOperationQuery<T::AccountId, T::BlockNumber>>,
         ) -> DispatchResultWithPostInfo {
             // We don't care about the sender.
             let _ = ensure_signed(origin)?;
 
-            Calls::<T>::execute_pending_operations(operations)
+            Calls::<T>::execute_pending_operations(operations)?;
+            // these fees have been pre-paid by request_delegate or request_undelegate
```
nanocryk, line 462: So one pay in advance for the operations execution? Can't this be abused by providing big  `operations` Vec without having to pay anything? One can provide a non existing operation, which will trigger a storage read to know that it has an amount of 0, which will skip the rest of the execution.

nanocryk:
Does the benchmark takes into account that steps changing manual rewards shares amounts will call `claim_manual_rewards` ?

pallets/pooled-staking/src/lib.rs
```diff
@@ -440,16 +449,20 @@ pub mod pallet {
             Calls::<T>::request_delegate(candidate, delegator, pool, stake)
         }
 
+        #[pallet::weight(T::WeightInfo::execute_pending_operations())]
```
nanocryk, line 452: Weight doesn't depend of the number of operations in the Vec?

pallets/pooled-staking/src/lib.rs
girazoki: I probably thought the providing a non-existent operation errored and you ended up paying fees associated with it. I see now we dont error when providing a non-existent operation.

If we make the one executing `execute_pending_operations` pay for the fees I see it unfair. Ideally executing pending operations is something healthy for the system (you are cleaning the pending storage) and it should be allowed to be done by anyone. It think it is a bit unfair to make this caller pay

pallets/pooled-staking/src/lib.rs
girazoki: fixed

tmpolaczyk:
`update_candidate_position` benchmark looks wrong, calling `request_delegate` already updates the candidate position. If you can mock pallet_session, a simple benchmark would be to first call `request_delegate` for all candidates with no keys, then insert the keys in pallet_session, and then calling `update_candidate_position` will do some work.

pallets/pooled-staking/src/lib.rs
nanocryk: This is a good idea. We can make non-existent operations error and make pay fees, however it might lead to multiple users calling it and pay fees while they expected to pay none.

PR #243, opened 2023-09-04
Manage XCM through maintenance mode filters

Agusrodri:
This PR adds the possibility of managing XCM functionalities through maintenance mode filters.

The following changes were applied to frontier and simple templates, and also to the dancebox runtime:

* `MaintenanceFilter` will forbid any call to `polkadotXcm` pallet.

* `xcm-support` feature was added to `pallet-maintenance-mode`, alongside with the custom implementations of `XcmExecutionManager`, `NormalDmpHandler` and `MaintenanceDmpHandler` types.


container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -432,8 +445,43 @@ impl Contains<RuntimeCall> for MaintenanceFilter {
 /// This can change in the future
 pub struct NormalFilter;
 impl Contains<RuntimeCall> for NormalFilter {
-    fn contains(_c: &RuntimeCall) -> bool {
-        true
+    fn contains(c: &RuntimeCall) -> bool {
+        match c {
+            // Only allow force_default_xcm_version
+            RuntimeCall::PolkadotXcm(method) => match method {
```
girazoki, line 451: This is actually not something we need, as we dont want to prohibit calls to polkadotXcm

test/suites/dev-frontier-template/test-maintenance/test-maintenance-xcm-normal-filter.ts
```diff
@@ -0,0 +1,76 @@
...
+
+                const dest = {
+                    V3: destMultilocation,
+                };
+
+                const message = {
+                    V3: [{ ClearOrigin: null }],
+                };
+
+                const sendTx = context.polkadotJs().tx.polkadotXcm.send(dest, message);
```
girazoki, line 67: These are pretty much the same test, we should test it in common

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -520,10 +522,20 @@ impl pallet_utility::Config for Runtime {
     type WeightInfo = pallet_utility::weights::SubstrateWeight<Runtime>;
 }
 
+pub struct XcmExecutionManager;
```
girazoki, line 525: I am starting to think we should move this into xcm-pritmitives

Agusrodri:
TODO: add tests for DMP and general XCM messages in maintenance mode.

girazoki:
@Agusrodri can you merge master in? Also we can merge as is but I will create an issue to move these into their proper primitives crates in moonkit

girazoki (approved):

PR #248, opened 2023-09-06
Benchmark generic XCM instructions

Agusrodri:
Adds benchmarking for XCM generic instructions of `pallet_xcm_benchmarks::generic`.

Also refactors benchmarks config inside the `dancebox` runtime. Individual calls to `list_benchmark!` and `add_benchmark!` macros were replaced by a single module called `benches` that contains all the desired pallets to be benchmarked in a cleaner way, making use of the `define_benchmarks!` macro.


runtime/dancebox/src/lib.rs
```diff
@@ -920,6 +920,17 @@ construct_runtime!(
...
 );
 
+#[cfg(feature = "runtime-benchmarks")]
+mod benches {
+    frame_benchmarking::define_benchmarks!(
+        [pallet_author_noting, AuthorNoting]
+        [pallet_configuration, Configuration]
+        [pallet_registrar, Registrar]
+        [pallet_invulnerables, Invulnerables]
+        [pallet_xcm_benchmarks::generic, pallet_xcm_benchmarks::generic::Pallet::<Runtime>]
```
tmpolaczyk, line 930: Unless it's somewhere else, this list is missing pallet_pooled_staking and frame_system:

```
            list_benchmark!(list, extra, frame_system, SystemBench::<Runtime>);

            list_benchmark!(
                list,
                extra,
                pallet_pooled_staking,
                PalledPooledStaking::<Runtime>
            );
```

Agusrodri: Yes, I might have missed the staking one after merging master. I'll add them.

runtime/dancebox/Cargo.toml
```diff
@@ -71,6 +71,7 @@ tp-traits = { workspace = true }
 
 # Polkadot
 pallet-xcm = { workspace = true }
+pallet-xcm-benchmarks = { workspace = true }
```
girazoki, line 74: pallet-xcm-benchmarks is probably only needed under the runtime-benchmarks feature? can we make it optional in that case?

girazoki (approved):
LGTM. Note this is only generic instructions, so I would add that to the PR title. Do we envision how can we use them in templates as well?

Agusrodri:
@girazoki if we need to implement these benchmarks for templates and Tanssi at the same time, we could build a `common` package and put the weights module inside it, but for it to be actually useful we should ensure that we want to use the same weights in templates and Tanssi all the time. Otherwise, if we start needing specific weights for each runtime, this common package will become useless.

So, is it planned that the templates and Tanssi use the same weights from now on? If that's the case, would you like me to apply the `common` package in this PR or in a separate one?

girazoki:
lets for now keep it separate. We will evaluate later

PR #253, opened 2023-09-07
migration for hold identifier

girazoki:
Migrates hold identifier from [u8; 8] to a proper enum holdReason

### DIFF
--- a/runtime/dancebox/src/migrations.rs
+++ b/runtime/dancebox/src/migrations.rs
@@ -184,9 +184,26 @@ where
         let should_be_migrated: Vec<(
             T::AccountId,
             BoundedVec<IdAmount<[u8; 8], <T as pallet_balances::Config>::Balance>, T::MaxHolds>,
-        )> = Decode::decode(migrated_holds.as_slice());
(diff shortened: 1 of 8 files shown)


runtime/dancebox/src/migrations.rs
```diff
@@ -108,15 +111,117 @@ where
...
+        .collect();
+
+        let migrated_count_read = stored_data.len() as u64;
+        let mut migrated_count_write = 0u64;
+
+        // Write to the new storage
+        for (account_id, holds) in stored_data {
+            let mut new_holds = vec![];
+
+            for hold in holds {
```
nanocryk, line 144: Would that turn all holds into PooledStake holds? No other pallets perform holds?

nanocryk (changes requested):

runtime/dancebox/src/migrations.rs
girazoki: we only have one hold type yes. The problem if I evaluate the hold type is that I still need to convert the rest of the hold identifiers to something and that something right now is only PooledStake

nanocryk (approved):

PR #259, opened 2023-09-11
Pooled Staking Runtime API

nanocryk:
Stake in currency amount is not available in storage, and instead must be computed from the amount of shares and the value of each share. The pending manual rewards also needs to be computed based on the difference between the pool counter and delegator "checkpoint" times the amount of shares.

To simplify the implementation of UIs, a new Runtime API is added to perform those computations.


tmpolaczyk:
You need to update api-augment:

```
cd typescript-api
pnpm tsx ./scripts/generate-types.ts
```

girazoki:
I would suggest we use the same procedure as in the CI. Compile the node and do:

```
cd typescript-api
pnpm install
pnpm create-local-interfaces
```

pallets/pooled-staking/src/lib.rs
```diff
@@ -517,4 +495,39 @@ pub mod pallet {
             Calls::<T>::update_candidate_position(&candidates)
         }
     }
+
+    impl<T: Config> Pallet<T> {
+        pub fn staked_in_pool(
+            candidate: Candidate<T>,
+            delegator: Delegator<T>,
+            pool: AllTargetPool,
+        ) -> Option<T::Balance> {
```
tmpolaczyk, line 504: What does `None` mean here? If the stake is 0 this should return `Some(0)`, no?

tmpolaczyk:
Also can you add some integration tests?

PR #260, opened 2023-09-11
Benchmark collator assignment

tmpolaczyk:
And register new session weight.


pallets/collator-assignment/src/lib.rs
```diff
@@ -55,6 +55,10 @@ use {
     },
 };
 
+#[cfg(feature = "runtime-benchmarks")]
```
girazoki, line 58: We can probably now record the weight when we have a session change. In a similar manner to how we do it in https://github.com/paritytech/cumulus/blob/d4bb2215bb28ee05159c4c7df1b3435177b5bf4e/pallets/collator-selection/src/lib.rs#L664

PR #262, opened 2023-09-11
Add smoke test: Invulnerables have priority over staking candidates

tmpolaczyk:


test/suites/smoke-test/test-invulnerables-priority.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        it({
+            id: "C01",
+            title: "Invulnerables have priority over staking candidates",
+            test: async function () {
+                if (runtimeVersion < 200) {
+                    return;
+                }
+
+                // TODO: we should read the invulnerables at the start of this session, because that's when collators are updated
```
tmpolaczyk, line 25: Any tips for this?

girazoki: I think its alright. This tests run every hour so you will at least catch once the changes per session.

Maybe we can have another test for checking that at session changes, we have the event of selecting new candidates

girazoki (approved):

PR #263, opened 2023-09-11
Add Proxy pallet to templates

Agusrodri:
Adds the `pallet-proxy` to both `simple` and `frontier` templates. Also migrates the proxy TS tests to the `common` folder.


container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -432,8 +526,17 @@ impl Contains<RuntimeCall> for MaintenanceFilter {
...
 impl Contains<RuntimeCall> for NormalFilter {
-    fn contains(_c: &RuntimeCall) -> bool {
-        true
+    fn contains(c: &RuntimeCall) -> bool {
+        match c {
+            // We filter anonymous proxy as they make "reserve" inconsistent
+            // See: https://github.com/paritytech/substrate/blob/37cca710eed3dadd4ed5364c7686608f5175cce1/frame/proxy/src/lib.rs#L270 // editorconfig-checker-disable-line
+            RuntimeCall::Proxy(method) => match method {
+                pallet_proxy::Call::create_pure { .. } => false,
+                pallet_proxy::Call::kill_pure { .. } => false,
```
Agusrodri, line 535: Do we actually want to filter these calls in frontier and simple templates?

girazoki: yes, better safe than sorry

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -408,6 +418,90 @@ impl pallet_utility::Config for Runtime {
...
+#[allow(clippy::unnecessary_cast)]
+pub enum ProxyType {
+    /// All calls can be proxied. This is the trivial/most permissive filter.
+    Any = 0,
+    /// Only extrinsics that do not transfer funds.
+    NonTransfer = 1,
+    /// Only extrinsics related to governance (democracy and collectives).
+    Governance = 2,
+    /// Only extrinsics related to staking.
+    Staking = 3,
```
girazoki, line 435: You dont need staking here, there  is no staking

Agusrodri: Yes I was hesitating whether to add it or not just in case we need the variant in the future, that's why I disabled it inside the filter. But I agree, it's better to remove it.

girazoki (approved):

PR #267, opened 2023-09-13
Add randomness to pallet-collator-assignment

tmpolaczyk:
Add config param to pallet_collator_assigment that triggers a full rotation of all the assigned collators.

Full rotation is only triggered every 24 sessions (configurable). For the other sessions we keep the old behavior, except that the collators and the container chain lists are shuffled, so container chains with a low para id no longer have priority.

Collators may still rotate on normal sessions in the same cases as before this PR: if another collator leaves, or if a container chain joins or leaves.

The random shuffle does not happen in session 0, and it doesn't happen in rust integration tests unless the random seed is explicitly set. Zombienet tests use real randomness from the relay and there is no way to mock that randomness, so collator assignment will change every time these tests are run. Most of the tests are unit tests because there it is easier to control the randomness.

The full_rotation_period is configurable in pallet_configuration, the default value is 24 sessions but it can be changed. A value of 0 disables collator rotation.


girazoki:
key to use: `CURRENT_BLOCK_RANDOMNESS`

tmpolaczyk:
Zombienet tests are failing because of the rocksdb lock bug, I will fix that in another PR, this is ready for review now.

pallets/collator-assignment/src/benchmarking.rs
```diff
@@ -89,6 +91,9 @@ mod benchmarks {
             0
         );
 
+        // TODO: make sure this did not call assign_collators_rotate_all, because that's cheaper
```
girazoki, line 94: how can we make sure of this?

pallets/collator-assignment/src/lib.rs
```diff
@@ -121,26 +125,66 @@ pub mod pallet {
...
-            let new_assigned = Self::assign_collators_always_keep_old(
-                collators,
-                &container_chain_ids,
-                T::HostConfiguration::min_collators_for_orchestrator(target_session_index) as usize,
-                T::HostConfiguration::max_collators_for_orchestrator(target_session_index) as usize,
-                T::HostConfiguration::collators_per_container(target_session_index) as usize,
-                old_assigned.clone(),
-            );
+            let new_assigned =
+                if T::ShouldRotateAllCollators::should_rotate_all_collators(target_session_index) {
```
girazoki, line 152: Questioning myself whether it makes sense that we additionally have a storage item as a switch-kill method for the rotation in this pallet. What do you think? does it break many things?

pallets/collator-assignment/src/benchmarking.rs
girazoki: I guess fetching previous collator assignment and the new one and make sure they are equal?

pallets/collator-assignment/src/mock.rs
```diff
@@ -162,10 +168,26 @@ impl tp_traits::GetSessionContainerChains<u32> for ContainerChainsGetter {
     }
 }
 
+pub struct RotateCollatorsEveryNSessions<Period>(PhantomData<Period>);
```
girazoki, line 171: Isnt this what we also use in the runtime? if so, can we make it a type that we use both for tests and runtime-?

pallets/collator-assignment/src/tests.rs
```diff
@@ -638,3 +638,219 @@ fn assign_collators_reorganize_container_chains_if_not_enough_collators() {
...
+        run_to_block(11);
+
+        let initial_assignment =
+            BTreeMap::from_iter(vec![(1, 999), (2, 999), (3, 1001), (4, 1001)]);
+
+        assert_eq!(assigned_collators(), initial_assignment,);
+
+        MockData::mutate(|m| {
+            // Seed chosen manually to see the case where container 1002 is given priority
+            m.random_seed = [2; 32];
```
girazoki, line 788: this seed is chosen on purpose to favour 1002 right?

runtime/dancebox/src/lib.rs
```diff
@@ -436,6 +439,67 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
...
+    // https://github.com/paritytech/polkadot/issues/2601
+    fn get_block_randomness_mixed(subject: &[u8]) -> Option<Hash> {
+        Self::get_block_randomness()
+            .map(|random_hash| mix_randomness::<Runtime>(random_hash, subject))
+    }
+}
+
+/// Combines the vrf output of the previous relay block with the provided subject.
+/// This ensures that the randomness will be different on different pallets, as long as the subject is different.
+fn mix_randomness<T: frame_system::Config>(vrf_output: Hash, subject: &[u8]) -> T::Hash {
```
girazoki, line 485: target for putting it also in primitives?

runtime/dancebox/src/lib.rs
```diff
@@ -444,6 +508,26 @@ impl pallet_initializer::ApplyNewSession<Runtime> for OwnApplySession {
...
+            {
+                // Return random_hash as a [u8; 32] instead of a Hash
+                let mut buf = [0u8; 32];
+                let len = sp_std::cmp::min(32, random_hash.as_ref().len());
+                buf[..len].copy_from_slice(&random_hash.as_ref()[..len]);
+
+                buf
+            } else {
+                // If there is no randomness (e.g when running in dev mode), return [0; 32]
+                // TODO: smoke test to ensure this never happens in a live network
```
girazoki, line 523: good call

runtime/dancebox/src/lib.rs
```diff
@@ -529,6 +616,7 @@ impl SessionManager<AccountId> for CollatorsFromInvulnerablesAndThenFromStaking
 parameter_types! {
     pub const Period: u32 = prod_or_fast!(1 * HOURS, 1 * MINUTES);
     pub const Offset: u32 = 0;
+    pub const CollatorRotationSessionPeriod: u32 = 5;
```
girazoki, line 619: I would put it to something higher for now, like 12 or 24

girazoki:
The PR looks good, I just have a few comments and doubts :)

runtime/dancebox/src/lib.rs
girazoki: or probably we can configure this with prod_or_fast macro

pallets/collator-assignment/src/tests.rs
```diff
@@ -638,3 +638,219 @@ fn assign_collators_reorganize_container_chains_if_not_enough_collators() {
...
+                (11, 999),
+                (12, 999)
+            ]),
+        );
+
+        MockData::mutate(|m| {
+            // We don't want to assign collators to container chains anymore
+            m.collators_per_container = 0;
+        });
+        run_to_block(21);
```
girazoki, line 680: we are still not triggering any randomness here, I guess if we go to block 26 the collators in the orchestrator change?

pallets/collator-assignment/src/lib.rs
tmpolaczyk: A kill-switch for the method should_rotate_all_collators can be implemented directly in the runtime, for example the storage item could live in pallet_configuration.

runtime/dancebox/src/lib.rs
tmpolaczyk: Yeah, unfortunately.

runtime/dancebox/src/lib.rs
tmpolaczyk: So what should be the final value?

pallets/collator-assignment/src/tests.rs
tmpolaczyk: This test is not related to this PR

pallets/collator-assignment/src/benchmarking.rs
tmpolaczyk: It's not that easy because if the new assignment is random there is a small chance that it is equal to the output of the other function. I was thinking of adding an event or something.

pallets/collator-assignment/src/mock.rs
tmpolaczyk: Ok I can move this to primitives

pallets/collator-assignment/src/benchmarking.rs
girazoki: maybe we can add an event when randomness is [0:u32] as well and cover both things (smoke-test plus this?)

tmpolaczyk: Good idea, what about an event like:

```
// TODO: rename
CollatorAssignmentEvent {
    seed: [u8; 32],
    full_rotation: bool,
}
```

This way we can see the seed for debugging purposes, and also check that full_rotations happens when it should. Any more info that you think would be valuable? The session_index is already known, and for the collator assignment I believe there is another event in some other pallet.

girazoki (approved):
LGTM now. Good job!

pallets/collator-assignment/src/lib.rs
```diff
@@ -77,16 +80,34 @@ pub mod pallet {
...
+        type ShouldRotateAllCollators: ShouldRotateAllCollators<Self::SessionIndex>;
         /// The weight information of this pallet.
         type WeightInfo: WeightInfo;
     }
 
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        // TODO: rename
+        NewPendingAssignment {
```
Agusrodri, line 104: Is this the final name for the event?

Agusrodri (approved):
Looks great! Amazing work 🥇

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Yes, I should remove the TODO

pallets/collator-assignment/src/mock.rs
```diff
@@ -14,9 +14,17 @@
 // You should have received a copy of the GNU General Public License
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
+use frame_support::traits::Hooks;
```
tmpolaczyk, line 17: fmt

pallets/collator-assignment/src/lib.rs
```diff
@@ -104,6 +126,13 @@ pub mod pallet {
     pub(crate) type PendingCollatorContainerChain<T: Config> =
         StorageValue<_, Option<AssignedCollators<T::AccountId>>, ValueQuery>;
 
+    /// Randomness from previous block. Used to shuffle collators on session change.
+    /// Should only be set on the last block of each session and should be killed on the on_initialize of the next block.
+    /// The default value of [0; 32] disables randomness in the pallet.
+    #[pallet::storage]
```
girazoki, line 132: While I like this, I am also missing a way to disable it. This probably cannot serve as a kill switch, as it is set  on_finalize of the last block before the session, so we would need to set it  0 on that block.

I think we can have an additional storage that serves as s kill switch, any opposition for this?

runtime/dancebox/src/migrations.rs
```diff
@@ -209,19 +210,63 @@ where
...
+    T: pallet_configuration::Config,
+{
+    fn friendly_name(&self) -> &str {
+        "TM_MigrateConfigurationFullRotationPeriod"
+    }
+
+    fn migrate(&self, _available_weight: Weight) -> Weight {
+        log::info!(target: LOG_TARGET, "migrate");
+
+        Configuration::set_full_rotation_period(RuntimeOrigin::root(), 24u32)
```
girazoki, line 225: I dont think this is going to work. set full rotation period schedules the configuration change, but wont be applied until after 2 sessions. During those two sessions the configuration wont be readible in the system, as the metadata is not the same

girazoki: we need to read the existing values in currentConfig and pendingConfigs, and modify so that it accepts the new structure

girazoki (changes requested):
I think the PR is mergable provided that we mofidy the migration, good job!

runtime/dancebox/src/migrations.rs
girazoki: There are some examples in https://github.com/paritytech/polkadot-sdk/commit/3bbb336ea707ee1149a6ff689d3a237138f18b2c#diff-39d3b12d624dbf61df8f5ea6fc18a9f12611347f4d730388d6a96deafbba5a71, although I am not sure they are super clear. I think its better to just read whatever value it exists there with the previous struct format, and you just add the new parameter

runtime/dancebox/src/migrations.rs
```diff
@@ -209,19 +210,63 @@ where
...
+            .expect("Failed to set full_rotation_period");
+        <T as pallet_configuration::Config>::WeightInfo::set_config_with_u32()
+    }
+
+    /// Run a standard pre-runtime test. This works the same way as in a normal runtime upgrade.
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade(&self) -> Result<Vec<u8>, sp_runtime::DispatchError> {
+        let zero_period = Configuration::config().full_rotation_period;
```
girazoki, line 233: this should indeed panic with try-runtime no? the config does not still have `full_rotation_period`

tmpolaczyk: I am assuming it defaults to 0

tmpolaczyk: In this case what happens is that `Configuration::config()` returns the default value, I am assuming because it is a ValueQuery. The default value has full_rotation_period = 24, so this test was wrong.

girazoki (approved):

PR #268, opened 2023-09-15
test all candidates have session keys

girazoki:
New smoke test testing all candidates have nmbs session keys, because they should not become candidates if they dont

### DIFF
--- a/test/suites/smoke-test/test-staking-session-keys.ts
+++ b/test/suites/smoke-test/test-staking-session-keys.ts
@@ -1,0 +1,44 @@
+import { beforeAll, describeSuite, expect } from "@moonwall/cli";
+import { ApiPromise } from "@polkadot/api";
+
+describeSuite({
+    id: "S07",
+    title: "Check staking eligible candidates have session keys",
+    foundationMethods: "read_only",
+    testCases: ({ it, context }) => {
+        let api: ApiPromise;
+        let runtimeVersion;
+
+        beforeAll(async () => {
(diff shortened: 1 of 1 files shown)


test/suites/smoke-test/test-staking-session-keys.ts
```diff
@@ -0,0 +1,44 @@
...
+                const allEntries = await api.query.session.keyOwner.entries();
+                const accounts = allEntries.map(([, account]) => account.toHuman());
+                const keys = allEntries.map(([key]) => key.toHuman());
+
+                const eligibleCandidates = await api.query.pooledStaking.sortedEligibleCandidates();
+
+                for (const c of eligibleCandidates) {
+                    const index = accounts.indexOf(c.candidate.toHuman());
+
+                    expect(index, `Candidate ${c.candidate.toHuman()} should have session keys`).not.toBe(-1);
```
tmpolaczyk, line 33: You could have used something like

```js
expect(accounts, `Candidate ${c.candidate.toHuman()} should have session keys`).to.contain(c.candidate.toHuman())
```

which gives a better error message in case of failure, but this is fine.

girazoki: yeah no worries, I can change it to this

girazoki: oh but then I need the index, to look for the candidate key types

tmpolaczyk (approved):

PR #269, opened 2023-09-18
Staking rewards distribution

nanocryk:
Add staking reward distribution to the Auto Compounding and Manual Rewards pools.
The function is not called automatically and must be called by another pallet to manage the inflation and other details.


librelois (approved):
Looks good to me

girazoki:
Can we add description to the PR? it helps devrel categorize important/non-breaking prs etc

pallets/pooled-staking/src/pools.rs
```diff
@@ -396,3 +429,130 @@ impl<T: Config> ManualRewards<T> {
...
+/// self-delegated (instantly). It is thus implemented by creating new
+/// AutoCompounding shares. This can lead to some rounding, which will be
+/// absorbed in the ManualRewards distribution, which simply consist of
+/// transfering the funds to the candidate account.
+#[allow(dead_code)]
+pub fn distribute_rewards<T: Config>(
+    candidate: &Candidate<T>,
+    rewards: T::Balance,
+) -> DispatchResultWithPostInfo {
+    let candidate_manual_rewards = distribute_rewards_inner::<T>(candidate, rewards)?;
```
girazoki, line 463:
```suggestion
    let candidate_manual_rewards = distribute_delegation_rewards::<T>(candidate, rewards)?;
```

pallets/pooled-staking/src/pools.rs
```diff
@@ -396,3 +429,130 @@ impl<T: Config> ManualRewards<T> {
...
+        auto_compounding_rewards: candidate_auto_rewards,
+        manual_claim_rewards: candidate_manual_rewards,
+    });
+    Pallet::<T>::deposit_event(Event::<T>::RewardedDelegators {
+        collator: candidate.clone(),
+        auto_compounding_rewards: delegators_auto_rewards,
+        manual_claim_rewards: delegators_manual_rewards,
+    });
+
+    Ok(candidate_manual_rewards)
```
girazoki, line 557: this function is very well documented and is easy to follow, congrats!

girazoki (approved):
The only thing missing is integration tests, but you need some sort of inflation for that. Let's not forget to add them later!

pallets/pooled-staking/src/pools.rs
```diff
@@ -396,3 +429,130 @@ impl<T: Config> ManualRewards<T> {
...
+}
+
+fn distribute_rewards_inner<T: Config>(
+    candidate: &Candidate<T>,
+    rewards: T::Balance,
+) -> Result<T::Balance, Error<T>> {
+    {
+        let candidate_auto_stake = AutoCompounding::<T>::computed_stake(candidate, candidate)?.0;
+        candidate_auto_stake
+    };
```
tmpolaczyk, line 484: Isn't this dead code?

```suggestion
```

primitives/core/src/lib.rs
```diff
@@ -96,3 +99,7 @@ pub mod well_known_keys {
     pub const SESSION_INDEX: &[u8] =
         &hex_literal::hex!["cec5070d609dd3497f72bde07fc96ba072763800a36a99fdfc7c10f6415f6ee6"];
 }
+
+pub trait DistributeRewards<AccountId, Balance> {
```
girazoki, line 103: Even if trivial, lets try to add some doc to the trait too

girazoki:
> > you need some sort of inflation for that.
> 
> I'm working on the inflation mechanism in the branch [elois-pallet-rewards](https://github.com/moondance-labs/tanssi/compare/elois-pallet-rewards), when it's advanced enough I'll open the PR against `jeremy-staking-rewards-distribution` (or against `master` if thgis PR is merged in the meantime).

Yep I think is fine, we can merge this as is and work on integration tests once your PR is ready

girazoki:
Also ideally, we should benchmark the reward distribution so that we can at least report the utilized weight

pallets/pooled-staking/src/pools.rs
```diff
@@ -361,6 +366,34 @@ impl<T: Config> ManualRewards<T> {
...
+        rewards: Stake<T::Balance>,
+    ) -> Result<Stake<T::Balance>, Error<T>> {
+        let Shares(supply) = Self::shares_supply(&candidate);
+        if supply.is_zero() {
+            return Ok(Stake(Zero::zero()));
+        }
+
+        let rewards_per_share = rewards
+            .0
+            .checked_div(&supply)
```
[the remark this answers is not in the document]
nanocryk: It is the opposite, it should be at least equal to the supply of shares to distribute 1 unit per share.

That's why `T::InitialManualClaimShareValue` should be big enough so that the supply is somewhat low and allow rewards to be properly distributed.

pallets/pooled-staking/src/benchmarking.rs
```diff
@@ -525,6 +525,60 @@ mod benchmarks {
...
+            }, PendingOperationQuery {
+                delegator: caller.clone(),
+                operation: JoiningManualRewards {
+                    candidate: caller.clone(),
+                    at: timer.clone(),
+                },
+            }],
+        )?;
+
+        T::Currency::mint_into(&caller, source_stake).unwrap();
```
tmpolaczyk, line 572:
```suggestion
        T::Currency::mint_into(&T::StakingAccount::get(), source_stake).unwrap();
```

tmpolaczyk:
Not sure if intended, but this test passes:

```rust
#[test]
fn candidate_only_no_stake() {
    // Rewarding a candidate that does not have any stake works
    ExtBuilder::default().build().execute_with(|| {
        test_distribution(
            &[],
            RewardRequest {
                collator: ACCOUNT_CANDIDATE_1,
                rewards: 1_000_000,
            },
            &[],
            Distribution {
                collator_auto: 0,
                collator_manual: 1_000_000, // 100% of rewards
                delegators_auto: 0,
                delegators_manual: 0, // 0% of rewards
            },
        )
    });
}
```

You can add that test if that's intended, or add a check that the candidate has enough shares if this should be an error.

pallets/pooled-staking/src/tests/rewards.rs
```diff
@@ -0,0 +1,579 @@
...
+                collator: ACCOUNT_CANDIDATE_1,
+                rewards: 1_000_000,
+            },
+            &[],
+            Distribution {
+                collator_auto: 0,
+                collator_manual: 200_000, // 20% of rewards
+                delegators_auto: 800_000, // 80% of rewards
+                delegators_manual: 0,
+            },
```
tmpolaczyk, line 576: This case is a bit dangerous if we take invulnerables into account. Since invulnerables will have 0 stake, if we use pallet_staking to reward them it will only work as expected only if they have no delegators. If some attacked decides to delegate the min amount to all the invulnerables, this attacker will be able to get 80% of the block rewards. I assume we don't want invulnerables to have delegators?

nanocryk: The reward distribution piloting pallet should not call this pallet distribute_rewards for an invulnerable.

tmpolaczyk: Ok, and what if a regular collator calls `request_undelegate`? They will still be able to produce blocks until the next session, and they will only have "leaving" shares, which are not taken into account in this algorithm, so it is the same case: a collator with 0 shares that may have delegators.

nanocryk: If the candidate wants to leave it seems fair they don't get rewards (outside of the dedicated 20%). Thanks to the joining period one cannot notice a candidate leaving and join to snap the rewards.

tmpolaczyk: Yes, but it feels a bit strange. Because there is a small incentive for a delegator to delegate the min amount for all the candidates, because when a candidate leaves they can get 80% of the rewards, if there are no other delegators.

girazoki: the rest of the delegators keep getting the delegation rewards until the cnadidate is kicked out (in 1+session) right?

tmpolaczyk (approved):

PR #270, opened 2023-09-19
Fix metrics server not stopping if there are open connections

tmpolaczyk:
This is the fix: https://github.com/moondance-labs/substrate/commit/b06f357724147a9bc4d2b5cf831f8ff714d74f37

This PR includes a new zombienet suite just to test that case, but it is not being run in CI.

### DIFF
--- a/test/configs/zombieTanssi.json
+++ b/test/configs/zombieTanssi.json
@@ -44,33 +44,27 @@
                 },
                 {
                     "name": "Collator2000-01",
-                    "command": "../target/release/tanssi-node",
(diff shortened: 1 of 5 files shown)


test/configs/zombieTanssi.json
```diff
@@ -44,27 +44,33 @@
                 },
                 {
                     "name": "Collator2000-01",
-                    "command": "../target/release/tanssi-node"
+                    "command": "../target/release/tanssi-node",
```
tmpolaczyk, line 47: Remove this

girazoki (approved):

PR #271, opened 2023-09-20
Improve test README.md

tmpolaczyk:


girazoki (approved):

girazoki:
I would merge this, we can modify the readme later if we find a reason why the logs are not shown without the server

PR #272, opened 2023-09-21
Upgrade to 1.1.0

girazoki:
Upgrades dependencies to 1.1.0 version of polkadot


client/consensus/.vscode/launch.json
```diff
@@ -0,0 +1,613 @@
+{
```
tmpolaczyk, line 1: Remove this file

girazoki:
all zombienet jobs are failing because moonwall is not prepared to fetch the release from the new polkadot-sdk github repo, CC @timbrinded, for now we can test locally though

timbrinded:
Have you tried adding the repo using the new config option in Moonwall config?

You can specify the repo author, name , binary to look for

container-chains/primitives/authorities-noting-inherent/src/mock.rs
```diff
@@ -153,3 +153,41 @@ impl InherentDataProvider for MockAuthoritiesNotingInherentDataProvider {
         None
     }
 }
+
+impl MockAuthoritiesNotingInherentDataProvider {
```
girazoki, line 157: This could be reworked a bit

container-chains/templates/frontier/node/src/service.rs
```diff
@@ -440,6 +457,7 @@ async fn start_node_impl(
         sync_service,
     };
 
+    #[allow(deprecated)]
```
girazoki, line 460: add TODO to make it even clearer that we need to adapt to async backing

container-chains/templates/simple/node/src/command.rs
```diff
@@ -333,8 +319,7 @@ pub fn run() -> Result<()> {
 				let parachain_account =
 					AccountIdConversion::<polkadot_primitives::AccountId>::into_account_truncating(&id);
 
-				let state_version = Cli::native_runtime_version(&config.chain_spec).state_version();
-				let block: Block = generate_genesis_block(&*config.chain_spec, state_version)
+				let block: Block = generate_genesis_block(&*config.chain_spec, sp_runtime::StateVersion::V1)
```
girazoki, line 322: Change this to reading it from runtime!

node/src/service.rs
```diff
@@ -441,15 +444,28 @@ async fn start_node_impl(
             para_id,
             relay_chain_interface: relay_chain_interface.clone(),
             net_config,
+            sybil_resistance_level: CollatorSybilResistance::Resistant,
```
girazoki, line 447: re-check

girazoki:
modfy exportGenesisStateCommand to cumulus one in node

container-chains/templates/simple/node/src/command.rs
girazoki: To be fair this is only used for logging purposes. I can try to change it to something where we read it from the runtime, but it would take a bit more to spin up the node. Thoughts? @tmpolaczyk @nanocryk

tmpolaczyk: I see, this is used to print the genesis block hash. It sounds useful, why do you say it will take more time to spin up a node?

girazoki: we experienced that when we changed the export-genesis-state from using a static state version value to reading it from the runtime. I dont think this should be the case if it is only done for logging purposes.

I think the best way is to print both state v0 and state v1 genesis states, and let the user read whichever is useful for it.

container-chains/templates/frontier/node/src/command.rs
```diff
@@ -360,10 +360,14 @@ pub fn run() -> Result<()> {
 				let parachain_account =
 					AccountIdConversion::<polkadot_primitives::AccountId>::into_account_truncating(&id);
 
-                // TODO: modify
-				let block: Block = generate_genesis_block(&*config.chain_spec, sp_runtime::StateVersion::V1)
+                // We log both genesis states for reference, as fetching it from runtime would take significant timw
```
tmpolaczyk, line 363: Not sure how useful is logging both, but it's amazing that generating 2 states is faster than fetching it from the runtime.

```suggestion
                // We log both genesis states for reference, as fetching it from runtime would take significant time
```

node/src/service.rs
girazoki: Apparently this is for consensus mechanisms where there is no sybil-resistance, unlike Aura (our case) or nimbus

test/moonwall.config.json
```diff
@@ -60,7 +60,7 @@
                     {
                         "name": "frontier-template",
                         "binPath": "../target/release/container-chain-template-frontier-node",
-                        "options": ["--dev", "--sealing=manual"],
+                        "options": ["--dev", "--sealing=manual", "--para-id=2000"],
```
tmpolaczyk, line 63: Why this change?

node/src/service.rs
```diff
@@ -605,6 +625,8 @@ async fn start_node_impl(
             sync_service,
         };
 
+        // TODO: change for async backing
+        #[allow(deprecated)]
         start_full_node(params)?;
```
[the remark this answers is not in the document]
girazoki: the way to start the node with async backing is different, not  with start_full_ndoe. But we will implement it in the future

node/src/command.rs
```diff
@@ -487,8 +469,8 @@ pub fn run() -> Result<()> {
 				let parachain_account =
 					AccountIdConversion::<polkadot_primitives::AccountId>::into_account_truncating(&id);
 
-				let state_version = Cli::native_runtime_version(&config.chain_spec).state_version();
-				let block: Block = generate_genesis_block(&*config.chain_spec, state_version)
+                // TODO: fetch it from
```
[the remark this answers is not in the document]
girazoki: the TODO here does not make sense, will remove it

container-chains/primitives/authorities-noting-inherent/src/mock.rs
girazoki: yes I can do it in this one

girazoki: Done

test/suites/dev-tanssi/fees/test_fee_balance_transfer.ts
```diff
@@ -54,10 +56,12 @@ describeSuite({
                 const basePlusWeightFee = (
                     await polkadotJs.call.transactionPaymentApi.queryWeightToFee(info2.weight)
                 ).toBigInt();
-                expect(basePlusWeightFee).to.equal(1000000n + 1630678n);
+                // TODO: where do these values come from?
+                expect(basePlusWeightFee).to.equal(1000000n + 1479873n);
```
tmpolaczyk, line 60: 1000000n is the base fee, which I believe is the fee paid for a transaction with 0 weight, and the other value comes from the weight of the extrinsic and it's only here to detect when it changes. In this case it's the weight of balances.transfer, which I guess has changed with the release.

tmpolaczyk: You can try moving the +1 here by changing the value to 1479874n.

test/suites/dev-tanssi/fees/test_fee_balance_transfer.ts
```diff
@@ -54,10 +56,12 @@ describeSuite({
...
                 expectedBasePlusWeightFee = basePlusWeightFee;
 
-                const expectedFee = basePlusWeightFee + BigInt(signedTx.encodedLength);
+                // TODO: revisit this 1 unit that is getting injected here
+                const expectedFee = basePlusWeightFee + BigInt(signedTx.encodedLength) + 1n;
```
girazoki, line 64: This +1 bothers me a lot

test/suites/dev-tanssi/fees/test_fee_balance_transfer.ts
girazoki: yeah but I want  to understand where it i coming from

tmpolaczyk: Oh, you can't move it here because then this expect would fail and you would need a -1 here...

girazoki: I think its coming from here: https://github.com/paritytech/polkadot-sdk/blob/2c7166eb9c82bedb02bba9882bc5aa5cd38b8fad/substrate/frame/transaction-payment/src/lib.rs#L761

tmpolaczyk: I don't think so, git blame shows 2 years ago. If it's in that file, then it must be this commit https://github.com/paritytech/polkadot-sdk/commit/2c7166eb9c82bedb02bba9882bc5aa5cd38b8fad

girazoki: it is a rounding error based on the fact that one api returns unadjustedFee (without multiplier) and the other returns adjusted (with multiplier). In our case we have a constant multiplier of 1, but due to rounding issues we get a +1

container-chains/primitives/authorities-noting-inherent/src/mock.rs
```diff
@@ -60,51 +60,23 @@ pub struct MockAuthoritiesNotingInherentDataProvider {
     pub authorities: Vec<NimbusId>,
 }
 
+// TODO: lots of duplicated code between these two funcs, we should refactor
```
Agusrodri, line 63:
```suggestion
```

Agusrodri (approved):
Looks good!

test-sproof-builder/src/lib.rs
```diff
@@ -123,24 +123,32 @@ impl ParaHeaderSproofBuilder {
...
+        // Fetch relevant keys
+        for item in self.items {
+            let para_key = item.para_id.twox_64_concat();
+            let key = [well_known_keys::PARAS_HEADS_INDEX, para_key.as_slice()].concat();
+            relevant_keys.push(key.clone());
+        }
+
         // Insert new keys and add them to relevant keys
-        {
+        /*       {
```
tmpolaczyk, line 146: Should this be removed?

PR #273, opened 2023-09-22
Allow swapping AutoCompounding shares for ManualReward shares (and the other way around)

nanocryk:
Currently a delegator needs to undelegate and delegate again to change in which pool their stake is, which takes time as they need to wait for the joining and leaving delays. However there is no security risk in changing pool, so it should be instant instead. This PR adds a new call to go from one pool to another, with the dust from joining the target pool considered leaving stake.

TODO: Benchmark


pallets/pooled-staking/src/lib.rs
```diff
@@ -516,5 +528,18 @@ pub mod pallet {
...
         }
+
+        #[pallet::weight(0)]
+        pub fn swap_pool(
+            origin: OriginFor<T>,
+            candidate: Candidate<T>,
+            source_pool: TargetPool,
+            amount: SharesOrStake<T::Balance>,
+        ) -> DispatchResultWithPostInfo {
+            // We don't care about the sender.
```
girazoki, line 539:
```suggestion
```

girazoki:
So a couple of things missing on this PR yet, so that we dont forget:

- integrations tests in rust
- integrations tests in TS
- benchmarking

pallets/pooled-staking/src/calls.rs
```diff
@@ -509,4 +436,187 @@ impl<T: Config> Calls<T> {
...
+
+    pub fn swap_pool(
+        candidate: Candidate<T>,
+        delegator: Delegator<T>,
+        source_pool: TargetPool,
+        amount: SharesOrStake<T::Balance>,
+    ) -> DispatchResultWithPostInfo {
+        // Converts amount to shares of the correct pool
+        let old_shares = match (amount, source_pool) {
+            (SharesOrStake::Shares(s), _) => s,
```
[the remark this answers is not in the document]
girazoki: Ah never mind, target pool only contains automcpounding and manual rewards. So its only to cover cases in which the amount is given as shares

girazoki (commented):
I'm happy with the way the PR looks, but I think we should add benchmark + integration tests before merging

girazoki (approved):
I think if we add a simple typescript test we this PR can get merged

pallets/pooled-staking/src/lib.rs
```diff
@@ -516,5 +528,43 @@ pub mod pallet {
 
             Calls::<T>::update_candidate_position(&candidates)
         }
+
+        #[pallet::weight(0)]
```
girazoki, line 532:
```suggestion
        #[pallet::weight(0)]
```
Put the benchmarked value here

runtime/dancebox/src/lib.rs
```diff
@@ -482,13 +482,7 @@ pub struct CollatorsFromInvulnerablesAndThenFromStaking;
 
 /// Play the role of the session manager.
 impl SessionManager<AccountId> for CollatorsFromInvulnerablesAndThenFromStaking {
-    fn new_session(index: SessionIndex) -> Option<Vec<AccountId>> {
-        log::info!(
```
girazoki, line 486: Is there a reason why we are removing this log?

nanocryk: Is this log useful? It get spammed in the terminal when running benchmarks.

girazoki: Hm is useful when running the regular mode I would say, maybe you can tell it not to run with runtime-benhcmarks features-?

tmpolaczyk: Yeah it is useful to me, never noticed it's spamming on benchmarks

girazoki: I actually did :)

girazoki (approved):

tmpolaczyk (approved):

PR #277, opened 2023-09-28
Add pallet reward inflation

librelois:
Add a new pallet reward-inflation that handle new supply issuance and distribution.

The inflation rate is a proportion of the total supply per tanssi block.
We could also create new supply per relay block or per time duration, which could help to manage multi-block correctly, but that's left for a future PR.

TODO:
- [x] Adapt rust staking tests to the new Credit API
- [x] Create rust tests in the new pallet reward-inflation
- [x] Add inflation integration tests at the dancebox runtime level


girazoki:
@librelois can you merge master in so that its easier to review?

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+//! # Inflation Rewards Pallet
+//!
+//! This pallet handle native token inflation and rewards dsitribution.
```
girazoki, line 19:
```suggestion
//! This pallet handle native token inflation and rewards distribution.
```

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+                    .unwrap_or(CreditOf::<T>::zero())
+                } else {
+                    CreditOf::<T>::zero()
+                };
+
+            // Get the number of chains at this block
+            weight += T::DbWeight::get().reads_writes(1, 1);
+            let registered_para_ids = T::ContainerChains::current_container_chains();
+
+            let number_of_chains: BalanceOf<T> =
```
girazoki, line 84: Add comment here that this is for tanssi

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+
+            // Issue new supply
+            let new_supply =
+                T::Currency::issue(T::InflationRate::get() * T::Currency::total_issuance());
+
+            // Split staking reward portion
+            let total_rewards = T::RewardsPortion::get() * new_supply.peek();
+            let (rewards_credit, reminder_credit) = new_supply.split(total_rewards);
+            let rewards_per_chain: BalanceOf<T> = rewards_credit.peek() / number_of_chains;
+            let (mut total_reminder, staking_rewards) = rewards_credit.split_merge(
```
girazoki, line 95: what does split_merge do?

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
+    fn on_container_author_noted(
+        author: &T::AccountId,
+        block_number: BlockNumber,
+        para_id: ParaId,
+    ) -> Weight {
+        let mut total_weight = T::DbWeight::get().reads_writes(1, 1);
+        let pending_authors_per_chain = if let Some(mut pending_authors_per_chain) =
```
girazoki, line 244: how would we have something to read here if we are writing this value later on this function? I am not sure I understand this function, so comments would be good

girazoki: We cannot have two authors for the same block number for the same para id right?

tmpolaczyk:
> what happens when we reward an invulnerable? right now we are not handling differently as fart as I can tell, and it goes through pallet-staking

According to @nanocryk it shouldn't work that way, pallet_staking should never be called with an invulnerable because of the issue you mention. But if a candidate does not have any delegations it will get 100% of the reward, reward is never "lost".

runtime/dancebox/src/lib.rs
```diff
@@ -967,6 +969,46 @@ impl pallet_pooled_staking::Config for Runtime {
...
+pub struct OnUnbalancedInflation;
+impl frame_support::traits::OnUnbalanced<Credit<AccountId, Balances>> for OnUnbalancedInflation {
+    fn on_nonzero_unbalanced(credit: Credit<AccountId, Balances>) {
+        let _ = <Balances as Balanced<_>>::resolve(&ParachainBondAccount::get(), credit);
+    }
+}
+
+impl pallet_inflation_rewards::Config for Runtime {
+    type Currency = Balances;
+    type ContainerChains = Registrar;
```
girazoki, line 1002: This is likely not right, Registrar might contain chains that are to be started next session, and those should not be rewarded. We should either:

- on initialize new session, store them in inflatoin-rewards/authority-noting pallet
- take them from the btree-map of collator-assignation

runtime/dancebox/src/lib.rs
```diff
@@ -563,6 +564,7 @@ impl pallet_author_noting::Config for Runtime {
     type SelfParaId = parachain_info::Pallet<Runtime>;
     type ContainerChainAuthor = CollatorAssignment;
     type RelayChainStateProvider = cumulus_pallet_parachain_system::RelaychainDataProvider<Self>;
+    type AuthorNotingHook = ();
```
girazoki, line 567: Important to set this

pallets/inflation-rewards/src/lib.rs
nanocryk: In substrate code:
```rust
fn split_merge(self, amount: Balance, others: (Self, Self)) -> (Self, Self) {
		let (a, b) = self.split(amount);
		(a.merge(others.0), b.merge(others.1))
}
```
So it splits (sub) `rewards_credit` by `total_rewards % number_of_chains` then merge (add) it with `reminder_credit`. It returns the result and the remaining funds.

The merge_split using `CreditOf::<T>::zero()` is a bit confusing. I would prefer using split and merge directly.

girazoki:
@librelois @nanocryk and I decided to simplify a bit the container-chain author rewarding for now. We removed the pending author rewards and the flow simply consists of:

- container-chain authors are rewarded immediatly and only if they produced a new block
- reward associated to container-chains from whom we did not retrieve a new author go to onUnbalanced
- we dont store any pending rewards or anything

This is not definite, but is simple enough to start testing without holding the PR. We are not breaking any APIs you were developing before.

You can see the commit here: [a97b3a23cc151081eaafc970f1cda2bf8ff010a7](https://github.com/moondance-labs/tanssi/pull/277/commits/a97b3a23cc151081eaafc970f1cda2bf8ff010a7)

pallets/inflation-rewards/src/tests.rs
```diff
@@ -0,0 +1,249 @@
...
+            on_unbalanced_account
+                + non_staking_rewards
+                + reminder
+                + (Permill::from_percent(35) * new_supply_2),
+        );
+
+        // and the author is not rewarded
+        assert_eq!(
+            get_balance(&container_author),
+            // 70% rewards for 2 chains, so 35% per chain
```
librelois, line 245:
```suggestion

```

runtime/dancebox/src/lib.rs
```diff
@@ -966,6 +972,52 @@ impl pallet_pooled_staking::Config for Runtime {
     type WeightInfo = pallet_pooled_staking::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+    pub ParachainBondAccount: AccountId32 = PalletId(*b"ParaBond").into_account_truncating();
+    pub PendingRewardsAccount: AccountId32 = PalletId(*b"PENDREWD").into_account_truncating();
+    // 5%/year with 2_629_800 blocks per year -> 0.05/2_629_800 = 19/1_000_000_000
+    pub const InflationRate: Perbill = prod_or_fast!(Perbill::from_parts(19), Perbill::from_percent(1));
```
girazoki, line 979: @librelois what was the reason for choosing a different inflation rate for dev mode?

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        fn on_initialize(_: BlockNumberFor<T>) -> Weight {
+            let mut weight = T::DbWeight::get().reads(1);
+
+            // Collect indistributed rewards, if any
+            // Any parachain we have not rewarded is handled by onUnbalanced
+            let not_distributed_rewards =
+                if let Some(chains_to_reward) = ChainsToReward::<T>::take() {
+                    // Collect and sum all undistributed rewards
+                    let rewards_not_distributed: BalanceOf<T> = chains_to_reward.rewards_per_chain
+                        * (chains_to_reward.para_ids.len() as u32).into();
```
tmpolaczyk, line 70: saturating_mul?

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+        type Currency: Inspect<Self::AccountId> + Balanced<Self::AccountId>;
+
+        type ContainerChains: GetCurrentContainerChains;
+
+        /// Get block author for self chain
+        type GetSelfChainBlockAuthor: Get<Self::AccountId>;
+
+        /// Inflation rate per relay block (proportion of the total issuance)
```
tmpolaczyk, line 138: Per relay block or per orchestrator block?

pallets/inflation-rewards/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        type InflationRate: Get<Perbill>;
+
+        /// The maximum number of block authors
+        type MaxAuthors: Get<u32>;
+
```
tmpolaczyk, line 143: Unused
```suggestion
```

pallets/pooled-staking/src/pools.rs
```diff
@@ -455,22 +455,24 @@ impl<T: Config> ManualRewards<T> {
 /// AutoCompounding shares. This can lead to some rounding, which will be
 /// absorbed in the ManualRewards distribution, which simply consist of
 /// transfering the funds to the candidate account.
+#[allow(dead_code)]
```
tmpolaczyk, line 458: Should be used now?
```suggestion
```

test/suites/dev-tanssi/issuance-rewards/test_issuance.ts
```diff
@@ -0,0 +1,35 @@
...
+                const supplyBefore = (await polkadotJs.query.balances.totalIssuance()).toBigInt();
+                await context.createBlock();
+
+                const events = await polkadotJs.query.system.events();
+
+                const issuance = await fetchIssuance(events).amount.toBigInt();
+
+                const supplyAfter = (await polkadotJs.query.balances.totalIssuance()).toBigInt();
+
+                // in dev mode is 1%
```
fgamundi, line 28: Is it enough to only test the issuance of dev mode?

pallets/inflation-rewards/src/tests.rs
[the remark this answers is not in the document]
girazoki: I changed these so that we also have container chains. So right now:

- we test that orchestrator receives rewards
- we test that container-chain author receives reward
- we test that parachain bond receives rewards

all through invulnerables. The only thing missing is to test they do get the same rewards through staking

test/suites/dev-tanssi/issuance-rewards/test_issuance.ts
girazoki: this is a good point. I am not entirely sure why we have different values for dev and prod, I will try to unify

tmpolaczyk:
runtime/Cargo.toml is missing "pallet-inflation-rewards/try-runtime" under "try-runtime"

pallets/author-noting/src/lib.rs
```diff
@@ -151,15 +156,39 @@ pub mod pallet {
...
+                                                    &block_info.author,
+                                                    block_info.block_number,
+                                                    para_id,
+                                                ),
+                                            );
+                                            let _ = core::mem::replace(old_block_info, block_info);
+                                        }
+                                    } else {
+                                        let _ = core::mem::replace(
+                                            maybe_old_block_info,
```
nanocryk, line 185: We don't reward author for the first block of this para_id?

librelois: Indeed, we should differentiate between the case where it's really the 1st block in the container chain and the case where this container chain has been removed from the LatestAuthor storage item because it stopped producing blocks a long time ago (We don't have a LatestAuthor pruning strategy at the moment, but we'll probably have one int he future).
So, we should check that `block_info.block_number` is zero

librelois: @nanocryk I finally decided to reward the author is there is no previous block, because we set the LatestAuthor at the same time so it can't be abuse for more than 1 block after pruning

girazoki: we should also put a TODO to inject a vector to `on_container_author_noted`, to better handle the weight estimations

test/suites/dev-tanssi/issuance-rewards/test_issuance.ts
girazoki: we will leave it as is for now, I dont want to break any tests. But food for thought for the future

girazoki (approved):

librelois (commented):
The changes looks good to me

nanocryk (approved):

PR #279, opened 2023-10-03
Add extrinsic to pause container-chains in pallet-registrar

Agusrodri:
This PR adds a new extrinsic called `pause_container_chain` to `pallet-registrar` which allows to pause a container-chain from collating without the need of deregistering it. Thus, the boot nodes and genesis data associated to the container-chain are not removed by calling the new extrinsic. 

TODO:

- [x] Benchmarking
- [x] Add command for the new extrinsic to `sudoRegisterPara.ts` file


pallets/registrar/src/tests.rs
```diff
@@ -599,7 +717,6 @@ fn weights_assigned_to_extrinsics_are_correct() {
                 .get_dispatch_info()
                 .weight,
             <() as crate::weights::WeightInfo>::mark_valid_for_collating(
-                <Test as crate::Config>::MaxGenesisDataSize::get(),
                 <Test as crate::Config>::MaxLengthParaIds::get()
             )
         );
```
tmpolaczyk, line 722: You can also add the case of `crate::Call::<Test>::pause_container_chain` to this test. Its purpose is to detect copy-paste errors, because when you add a new extrinsic you will probably copy the weight annotation from another extrinsic, so this is to ensure that you don't forget to change it.

pallets/registrar/src/lib.rs
```diff
@@ -441,6 +442,37 @@ pub mod pallet {
...
+        pub fn pause_container_chain(origin: OriginFor<T>, para_id: ParaId) -> DispatchResult {
+            T::RegistrarOrigin::ensure_origin(origin)?;
+
+            let mut pending_verification = PendingVerification::<T>::get();
+            match pending_verification.binary_search(&para_id) {
+                Ok(_) => return Err(Error::<T>::ParaIdAlreadyPaused.into()),
+                Err(index) => {
+                    pending_verification
+                        .try_insert(index, para_id)
+                        .map_err(|_e| Error::<T>::ParaIdListFull)?;
```
tmpolaczyk, line 458: It's curious that pausing a container chain will fail when the list of pending container chains is full. I guess it's not a problem because this must be called by root, and root can deregister the pending chains in a batch transaction.

pallets/registrar/src/benchmarks.rs
```diff
@@ -179,5 +179,53 @@ mod benchmarks {
...
+            let (caller, _deposit_amount) =
+                create_funded_user::<T>("caller", i, T::DepositAmount::get());
+            Pallet::<T>::register(
+                RawOrigin::Signed(caller.clone()).into(),
+                i.into(),
+                storage.clone(),
+            )
+            .unwrap();
+            Pallet::<T>::mark_valid_for_collating(RawOrigin::Root.into(), i.into()).unwrap();
+        }
```
tmpolaczyk, line 203: I believe the worst case for this benchmark is when both `pending_verification` and `registered_para_ids` are full. So what do you think about adding another for loop like this one but without `mark_valid_for_collating`?
So that paraIds `0..=50` are in `registered_para_ids` and paraIds `1000..=1050` are in `pending_verification` as an example. And then paraId 49 is paused.

And the same applies to the `mark_valid_for_collating` benchmark, it is missing this small extra cost.

pallets/registrar/src/lib.rs
Agusrodri: Yeah I agree, but what if root tries to pause the container chain without a batch tx? In that case it would fail if we don't handle the situation, and I think it's a good idea to provide a proper error message if that happens.

pallets/registrar/src/benchmarks.rs
Agusrodri: Yes! Agree on this. Will add them.

pallets/registrar/src/lib.rs
tmpolaczyk: I think the current error is already descriptive enough

pallets/registrar/src/weights.rs
```diff
@@ -110,58 +104,77 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-			// Standard Error: 10_160
-			.saturating_add(Weight::from_parts(345_240, 0).saturating_mul(y.into()))
-			.saturating_add(T::DbWeight::get().reads(4_u64))
+		//  Measured:  `418 + y * (8 ±0)`
+		//  Estimated: `1899 + y * (8 ±0)`
+		// Minimum execution time: 17_420_000 picoseconds.
+		Weight::from_parts(32_834_209, 1899)
+			// Standard Error: 16_866
+			.saturating_add(Weight::from_parts(391_115, 0).saturating_mul(y.into()))
+			.saturating_add(T::DbWeight::get().reads(3_u64))
```
tmpolaczyk, line 134:
```suggestion
			.saturating_add(T::DbWeight::get().reads(4_u64))
```

tmpolaczyk (approved):

PR #280, opened 2023-10-05
Add cli-flag to keep db after changing assignments

Agusrodri:
This PR adds a simple `keep_db` flag to `ContainerChainRunCmd` struct that allows to keep container-chain databases after changing collator assignments.


node/src/container_chain_spawner.rs
```diff
@@ -238,6 +240,26 @@ impl ContainerChainSpawner {
...
+
+            let container_client_genesis_hash = container_chain_client.chain_info().genesis_hash;
+
+            if container_client_genesis_hash != chain_spec_genesis_hash {
+                log::info!(
+                    "Container genesis {:?} different from chain spec genesis {:?} - Deleting container db", 
+                    container_client_genesis_hash, 
+                    chain_spec_genesis_hash
+                );
+                delete_container_chain_db(&db_path);
```
tmpolaczyk, line 259: This won't work because the node has already been started, we need to stop the container chain, delete the db, and start it again.

tmpolaczyk: I pushed a commit to your branch to fix this, you can now return an error "NeedsRestartAndDbRemoval" and the container chain will delete the database and restart. Feel free to modify it. Also I merged master and there are some compile errors because of the 1.1.0 upgrade, sorry!

node/src/container_chain_spawner.rs
```diff
@@ -363,6 +453,10 @@ impl ContainerChainSpawner {
                 );
             }
         }
+
+        if state.assigned_para_id.unwrap() != container_chain_para_id && state.next_assigned_para_id.unwrap() != container_chain_para_id{
+            delete_container_chain_db(&self.container_chain_cli.base_path);
```
Agusrodri, line 458: Is this path the correct one? I didn't find any other that could be more explicit to point to the db in the current scope.

node/src/container_chain_spawner.rs
```diff
@@ -410,7 +504,7 @@ impl ContainerChainSpawner {
             // Edge case: when starting the node it may be assigned to a container chain, so we need to
             // start a container chain already collating.
             let start_collation = Some(para_id) == current;
-            self.spawn(para_id, start_collation).await;
+            self.spawn(para_id, start_collation, false).await;
```
Agusrodri, line 507: Do we need warp sync to be enabled in this case?

Agusrodri:
One of the warp sync tests is failing because it seems the container chain db for `Collator2000-02` is not being deleted after the new configuration. I'm investigating the reason of the issue.

tmpolaczyk:
>One of the warp sync tests is failing because it seems the container chain db for Collator2000-02 is not being deleted after the new configuration. I'm investigating the reason of the issue.

That was because it was panicking on this line: `state.assigned_para_id.unwrap()`, you could see it in the node logs. `assigned_para_id` can be `None` if the collator is not assigned to any chain.

I pushed a commit fixing that by using a different kind of channel to signal shutdown, now we can distinguish between "stopping because someone called .stop()" and "stopping because the node process is stopping". We will only delete the db in the first case.

node/src/container_chain_spawner.rs
tmpolaczyk: No, because that could potentially delete the db. So the first call to spawn will be warp_sync = false, and the recursive call will be warp_sync = true.

node/src/container_chain_spawner.rs
```diff
@@ -312,10 +405,13 @@ impl ContainerChainSpawner {
...
-                    _ = on_exit_future => {
-                        // Graceful shutdown
+                    stop_unassigned = on_exit_future => {
+                        // Graceful shutdown.
+                        // `stop_unassigned` will be `Ok` if `.stop()` has been called, which means that the
+                        // container chain has been unassigned, and will be `Err` if the handle has been dropped,
+                        // which means that the node is stopping.
                         // Delete existing database if running as collator
-                        if validator {
+                        if validator && stop_unassigned.is_ok() && !container_chain_cli.base.keep_db && warp_sync {
```
tmpolaczyk, line 414:
```suggestion
                        if validator && stop_unassigned.is_ok() && !container_chain_cli.base.keep_db {
```

When stopping a container chain we want to always delete the db to save space.

tmpolaczyk (approved):

node/src/container_chain_spawner.rs
```diff
@@ -249,8 +277,62 @@ impl ContainerChainSpawner {
...
 
+            // Get latest block number from the container chain client
+            let last_container_block = container_chain_client.chain_info().best_number;
+
+            // Get the container chain's latest block from orchestrator chain and compare with client's one
+            let last_container_block_from_orchestrator = orchestrator_runtime_api
+                .latest_block_number(orchestrator_chain_info.best_hash, container_chain_para_id)
+                .unwrap_or_default();
+
+            let max_block_diff_allowed = 10u32;
```
girazoki, line 288:
```suggestion
            let max_block_diff_allowed = 1000u32;
```

girazoki: I think 10 is very few

girazoki, line 288:
```suggestion
            let max_block_diff_allowed = 100u32;
```

girazoki (approved):

node/src/container_chain_spawner.rs
```diff
@@ -226,7 +254,7 @@ impl ContainerChainSpawner {
             container_chain_cli_config.database.set_path(&db_path);
 
             // Delete existing database if running as collator
-            if validator {
+            if validator && !container_chain_cli.base.keep_db && warp_sync {
```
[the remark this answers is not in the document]
girazoki: we should immediatly delete the container-chain-db if the spec is differnt I believe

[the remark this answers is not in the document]
girazoki: there is no reason to maintain a DB that does not match with the specs right?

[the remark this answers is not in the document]
tmpolaczyk: You are right

node/src/container_chain_spawner.rs
```diff
@@ -330,14 +416,24 @@ impl ContainerChainSpawner {
...
             match try_closure().await {
                 Ok(()) => {}
+                Err(sc_service::error::Error::Application(e)) if e.is::<NeedsRestart>() => {
+                    let e = e.downcast::<NeedsRestart>().unwrap();
+
+                    log::info!("Restarting container chain {}", container_chain_para_id);
+                    // self.spawn must return a boxed future because of the recursion here
+                    e.self2
+                        .spawn(container_chain_para_id, start_collation, e.warp_sync)
+                        .await;
```
tmpolaczyk, line 429: Unfortunately this doesn't work because the database doesn't have time to close, resulting in a db lock error. We should add a sleep(10) here, or check if the db is still alive using the container_chain_monitor.

tmpolaczyk (changes requested):

girazoki:
Leaving here that what we want is:
- possibly remove keep_db flag
- remove db if the runtime for the chain we are assigned returns 0
- remove all other container dbs that are not our current or next assignments

girazoki:
I like the way it looks now, we can unitest the delete db cases quite easily!

tmpolaczyk (approved):
self-approving

PR #281, opened 2023-10-09
Implement kill_author_data extrinsic

tmpolaczyk:
This extrinsic can be used to clear the LatestAuthor storage when a paraid has been removed from the relay chain, so it will no longer produce blocks.


girazoki:
Let's add a small unitest + integration test (only rust its fine)

girazoki (approved):

PR #282, opened 2023-10-10
Fix rocksdb lock panic and add container chain monitor

tmpolaczyk:
Fixes rocksdb lock panic by using a patched version of jsonrpsee.

Adds a "container chain monitor" that will warn when a node is running more container chains than expected, and can be used to debug the cause of why the container chains haven't stopped properly.


node/src/service.rs
```diff
@@ -169,6 +169,14 @@ pub fn new_partial(
...
         )?;
     let client = Arc::new(client);
+    log::info!(
+        "backend in new_partial refcount: {}",
+        Arc::strong_count(&backend)
+    );
+    log::info!(
+        "client in new_partial refcount: {}",
+        Arc::strong_count(&client)
+    );
```
tmpolaczyk, line 179:
```suggestion
```

node/src/service.rs
```diff
@@ -626,6 +634,46 @@ async fn start_node_impl(
             }
         }
 
+        let debug_start_same_cc = async move {
```
tmpolaczyk, line 637: Also remove this

node/src/service.rs
```diff
@@ -709,6 +765,14 @@ pub async fn start_node_impl_container(
...
         params_import_queue = params.import_queue;
     }
+    log::info!(
+        "backend start_node_impl_container refcount: {}",
+        Arc::strong_count(&backend)
+    );
+    log::info!(
+        "client start_node_impl_container refcount: {}",
+        Arc::strong_count(&client)
+    );
```
tmpolaczyk, line 775:
```suggestion
```

girazoki:
@tmpolaczyk I took the chance of pointing at the right repo and fixing clippy, please re-visist that I did not mess up

tmpolaczyk:
>@tmpolaczyk I took the chance of pointing at the right repo and fixing clippy, please re-visist that I did not mess up

Looks good, thanks! The reason I decided to use a patch section in Cargo.toml instead of modifying the repo url for all of our dependencies, like we usually do when we fork substrate, is because everything depends on jsonrpsee, so we would have needed to change all the repos.

girazoki:
yeah I think its the right approach!

node/src/service.rs
```diff
@@ -164,6 +164,7 @@ pub fn new_partial(
         .build();
 
     let executor = ParachainExecutor::new_with_wasm_executor(wasm);
+    //let wasm_cache = executor.cache;
```
girazoki, line 167:
```suggestion
```

girazoki (approved):

PR #288, opened 2023-10-16
pallet-initializer new session is applied immediately 

fgamundi:
Delaying the application of a new session until `on_finalize` was producing unwanted behavior. This changes it to be applied immediately on `apply_new_session`, and removes `BufferedSessionChange` as it was only used to temporarily save the session change until `on_finalize`

### DIFF
--- a/pallets/initializer/src/lib.rs
+++ b/pallets/initializer/src/lib.rs
@@ -32,7 +32,6 @@ mod tests;
 pub use pallet::*;
 use {
     frame_support::{pallet_prelude::*, traits::OneSessionHandler},
-    parity_scale_codec::{Decode, Encode},
     scale_info::TypeInfo,
     sp_runtime::{
         traits::{AtLeast32BitUnsigned, Zero},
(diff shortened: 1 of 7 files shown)


pallets/initializer/src/lib.rs
```diff
@@ -139,14 +95,7 @@ impl<T: Config> Pallet<T> {
...
         } else {
-            BufferedSessionChanges::<T>::mutate(|v| {
-                *v = Some(BufferedSessionChange {
-                    changed,
-                    validators,
-                    queued,
-                    session_index,
-                })
-            });
+            T::SessionHandler::apply_new_session(changed, session_index, validators, queued);
```
tmpolaczyk, line 98: This if-else is now no longer needed. I assume changed will be false when session_index == 0, so you can just always do:

```rust
T::SessionHandler::apply_new_session(changed, session_index, validators, queued);
```

girazoki (approved):
LGTM

tmpolaczyk (approved):

PR #293, opened 2023-10-17
Give priority to invulnerables as orchestrator collators

tmpolaczyk:
When rotating collators, the orchestrator chain will be filled with `min_orchestrator_collators` invulnerables. If there are not enough invulnerables, the rest is filled with regular collators. The rest (up to `max_orchestrator_collators`) is also filled with regular collators (regular means anything, can be invulnerable or not invulnerable).

### DIFF
--- a/pallets/collator-assignment/src/lib.rs
+++ b/pallets/collator-assignment/src/lib.rs
@@ -316,7 +316,7 @@ pub mod pallet {
             // Fill orchestrator chain collators up to min_num_orchestrator_chain
             // Give priority to invulnerables
             let num_missing_orchestrator_collators =
-                min_num_orchestrator_chain - new_assigned.orchestrator_chain.len();
+                min_num_orchestrator_chain.saturating_sub(new_assigned.orchestrator_chain.len());
             let invulnerables_for_orchestrator = T::RemoveInvulnerables::remove_invulnerables(
                 &mut new_collators,
                 num_missing_orchestrator_collators,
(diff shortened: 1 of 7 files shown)


pallets/collator-assignment/src/lib.rs
```diff
@@ -313,6 +313,13 @@ pub mod pallet {
             }
 
             // Fill orchestrator chain collators up to min_num_orchestrator_chain
+            // Give priority to invulnerables
+            let num_missing_orchestrator_collators = min_num_orchestrator_chain - new_assigned.orchestrator_chain.len();
```
tmpolaczyk, line 317: saturating_sub

runtime/dancebox/src/lib.rs
```diff
@@ -650,6 +651,30 @@ impl GetRandomnessForNextBlock<u32> for BabeGetRandomnessForNextBlock {
...
 
+pub struct RemoveInvulnerablesImpl;
+
+impl RemoveInvulnerables<AccountId> for RemoveInvulnerablesImpl {
+    fn remove_invulnerables(
+        collators: &mut Vec<AccountId>,
+        num_invulnerables: usize,
+    ) -> Vec<AccountId> {
+        // TODO: check if this works on session changes
+        let all_invulnerables = pallet_invulnerables::Invulnerables::<Runtime>::get();
```
girazoki, line 662: we should be able to test this with zombienet right?

girazoki: ah but in zombienet all of them are invulnerables, I see

PR #294, opened 2023-10-17
Add container chain 2002 to tanssi node chainspec

### LINKED ISSUE #286: Zombienet Warp Sync: template-simple-node does not produce blocks
Hey tanssi team,

I successfully generated the necessary binary files (tanssi, polkadot, container-chain-template-simple-node) for the execution of the following commands from the README.md:

```
# Generates the latest specs for orchestrator and container-chains
npm run build-spec

# Spawns Tanssi and container-chains with zombienet
/path/to/zombienet spawn -p native ./configs/zombieTanssiWarpSync.json
```

Now I go to the polkadot js interface and see all collators and polkadot validators producing blocks, except the `container-chain-template-simple-node` does not. Do you experience the same issue?

Chralt98:
Fixes #286

I couldn't produce blocks on the second container chain instance with parachain id 2002. 

So, I decided to fix that by adding the required information in the build-spec.sh script.

There is a current waiting time of 156 seconds on the local rococo relaychain to let the container chains produce their first blocks!


girazoki:
The reason why 2002 is not registered from the beginning is because we test live registrations in our of tests. If you really need 2002 to work, you can make a PR with a new script that builds the specs, and add it as a new test suite to the moonwall config

Chralt98:
No not required thanks. It was just confusing.

PR #295, opened 2023-10-19
Inflation rewards integration tests

nanocryk:
Adds integration tests for #277. Modifies the integration test logic to be able to produce a single block, and also collect data for each block (currently only the author and the inflation).


runtime/dancebox/tests/common/mod.rs
```diff
@@ -339,3 +369,21 @@ pub const ALICE: [u8; 32] = [4u8; 32];
 pub const BOB: [u8; 32] = [5u8; 32];
 pub const CHARLIE: [u8; 32] = [6u8; 32];
 pub const DAVE: [u8; 32] = [7u8; 32];
+
+/// Calls `f_loop` with all values in `iter`, and returns the first `Some(value)`.
+/// If only `None` is returned, the value returned by `f_else`  is returned. It
+/// is like a for loop which could have an else block in case no iteration
+/// breaks with a value.
+pub fn for_else<I, F, E, R>(iter: I, mut f_loop: F, f_else: E) -> R
```
tmpolaczyk, line 377: I believe you can use

```rust
iter.find_map(f_loop).unwrap_or_else(f_else)
```

So you shouldn't need this helper function.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map

nanocryk: Oh very neat, didn't knew `find_map` :p

runtime/dancebox/tests/integration_test.rs
```diff
@@ -3598,6 +3590,277 @@ fn test_migration_holds() {
...
+                .unwrap_or_else(|| panic!("ALICE doesn't seem to author any blocks"));
+
+            let balance_after = System::account(account).data.free;
+
+            let all_rewards = RewardsPortion::get() * summary.inflation;
+            // rewards are shared between orchestrator and registered paras
+            let orchestrator_rewards = all_rewards / 3;
+            assert_eq!(
+                orchestrator_rewards,
+                dbg!(balance_after - balance_before),
```
tmpolaczyk, line 3799:
```suggestion
                balance_after - balance_before,
```

runtime/dancebox/tests/integration_test.rs
```diff
@@ -3598,6 +3590,277 @@ fn test_migration_holds() {
...
+            assert_eq!(summary.author_id, ALICE.into());
+
+            let balance_after = System::account(account).data.free;
+
+            let all_rewards = RewardsPortion::get() * summary.inflation;
+            // rewards are shared between orchestrator and registered paras
+            let orchestrator_rewards = all_rewards / 3;
+            assert_eq!(
+                orchestrator_rewards,
+                dbg!(balance_after - balance_before),
```
tmpolaczyk, line 3858:
```suggestion
                balance_after - balance_before,
```

runtime/dancebox/tests/integration_test.rs
```diff
@@ -3598,6 +3590,277 @@ fn test_migration_holds() {
...
+                .unwrap_or_else(|| panic!("DAVE doesn't seem to author any blocks"));
+            let balance_after = System::account(account).data.free;
+
+            let all_rewards = RewardsPortion::get() * summary.inflation;
+            // rewards are shared between orchestrator and registered paras
+            let orchestrator_rewards = all_rewards / 3;
+            let candidate_rewards = orchestrator_rewards * 2 / 10;
+
+            assert_eq!(
+                candidate_rewards + 1, // TODO: account for rounding properly :p
```
tmpolaczyk, line 3696: Are we going to fix this? Maybe add another test that checks that the rewarded container chains plus the orchestrator is exactly equal to all_rewards?

tmpolaczyk (approved):

runtime/dancebox/tests/integration_test.rs
girazoki: I have a bunch of those as well, probably worth creating an issue

girazoki (approved):

PR #296, opened 2023-10-19
Fix try-runtime build and add missing migrations

fgamundi:
`try-runtime` now fails if there's a mismatch between a pallet's latest vs onchain versions. This flagged `PolkadotXcm` and `XcmpQueue` from live dancebox.
Missing migrations has been added


runtime/dancebox/src/migrations.rs
```diff
@@ -209,19 +210,56 @@ where
...
+impl<T> Migration for XcmpQueueMigration<T>
+where
+    T: cumulus_pallet_xcmp_queue::Config,
+{
+    fn friendly_name(&self) -> &str {
+        "MM_XcmpQueueMigration"
+    }
+
+    fn migrate(&self, _available_weight: Weight) -> Weight {
+        cumulus_pallet_xcmp_queue::migration::Migration::<T>::on_runtime_upgrade()
```
girazoki, line 237: do we know for certain this puts the v3 on the pallet?

girazoki: because we probably need to set v2 first before doing this on_runtime_upgrade

fgamundi: Yes, even if onchain is v1, it applies first the migration to v2 and immediately after the migration to v3 https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/pallets/xcmp-queue/src/migration.rs#L36C1-L46C4

tmpolaczyk:
Can you merge master? I just merged a big PR that includes migrations, so there will be conflicts.

girazoki (approved):

fgamundi:
@tmpolaczyk done

PR #302, opened 2023-10-28
Format files and add try-runtime feature for maintenance pallet

Chralt98:
I always find it better readable to split the std, benchmark and try-runtime features into multiple lines.

So I run `taplo fmt` on all `Cargo.toml` files. (In specific terms I used the `Even Better TOML` extension of VSCode.)

I also did run `cargo +nightly clippy --all-features --all-targets -p container-chain-template-simple-runtime` and found out that the maintenance pallet also has to enable the `try-runtime` feature.

Additionally I run `cargo +nightly fmt` to properly format mainly the imports.


tmpolaczyk:
Hi, sorry for no reply. We have improved our CI with new toml linters, zepter to ensure that the features are correctly propagated, and toml maid as an alternative to taplo:

https://github.com/moondance-labs/tanssi/pull/368
https://github.com/moondance-labs/tanssi/pull/369

This is because manually keeping track of the features was very error prone, as you have noticed, so we have automated it. You shouldn't run into similar issues in the future, but if you do, feel free to open a pull request!

PR #306, opened 2023-10-31
Increment ports of "odd" parachains to avoid port conflicts

tmpolaczyk:
The user can specify p2p, rpc, and prometheus ports using the CLI, and since there can be 2 container chains running at the same time, the port used may be the one specified by the user, or that +1


tmpolaczyk:
Superseded by #307, we will be using a different solution

PR #307, opened 2023-10-31
Always restart container chain before starting collation

tmpolaczyk:
Container chains can start in "sync mode", which uses random ports, or in "collation mode" which uses the ports provided by the user through CLI. To ensure that the container chain that is currently collating uses the correct ports, we restart it right when it is time to start collation.

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -189,6 +189,48 @@ impl ContainerChainSpawner {
                 container_chain_para_id
             );
 
+            let odd_parachain = {
+                let state = state.lock().expect("poison error");
+                let monitor_id = state.spawned_containers_monitor.count;
+
+                monitor_id % 2 == 1
+            };
+
+            if odd_parachain {
+                log::info!("This is an odd parachain, incrementing all the ports by 1");
+                // Increment all the ports by 1 to avoid conflicts with the other running container chain
+                container_chain_cli
+                    .base
+                    .base
+                    .prometheus_params
+                    .prometheus_port = Some(
+                    container_chain_cli
+                        .base
+                        .base
+                        .prometheus_params
+                        .prometheus_port
+                        .unwrap_or(9617)
+                        .saturating_add(1),
+                );
+                container_chain_cli.base.base.network_params.port = Some(
+                    container_chain_cli
+                        .base
+                        .base
+                        .network_params
+                        .port
+                        .unwrap_or(30335)
+                        .saturating_add(1),
+                );
+                container_chain_cli.base.base.rpc_port = Some(
+                    container_chain_cli
+                        .base
+                        .base
+                        .rpc_port
+                        .unwrap_or(9946)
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
Check edge case of going from (2000, 2001) to (2001, 2000)

node/src/container_chain_spawner.rs
```diff
@@ -89,14 +89,16 @@ pub struct ContainerChainSpawnerState {
 
 pub struct ContainerChainState {
     /// Async callback that enables collation on this container chain
+    // We don't use it since we are always restarting container chains
+    #[allow(unused)]
```
girazoki, line 93: Interesting should we remove it already then?

node/src/container_chain_spawner.rs
```diff
@@ -189,6 +191,19 @@ impl ContainerChainSpawner {
                 container_chain_para_id
             );
 
+            if !start_collation {
+                log::info!("This is a syncing container chain, using random ports");
+                // Use random ports to avoid conflicts with the other running container chain
+                let random_ports = [23456, 23457, 23458];
```
girazoki, line 197: wont it automatically use random ports even if we dont assign any?, in other words, can we use None here?

node/src/container_chain_spawner.rs
```diff
@@ -483,32 +510,60 @@ fn handle_update_assignment_state_change(
...
+        if running_chains_before.contains(&id) && !chains_to_stop.contains(&id) {
+            chains_to_stop.push(id);
+        }
+        if !chains_to_start.contains(&id) {
+            chains_to_start.push(id);
+        }
+    }
+
+    if need_to_restart_next {
+        // Handle edge case of going from (2000, 2001) to (2001, 2000). In that case we must restart both chains,
```
girazoki, line 551: in the second case, it restarts with random ports?

tmpolaczyk: Yes, we stop both of them, wait 10 seconds, and then start the new "syncing" with random ports, and the new "collating" with the good ports (in either order).

node/src/container_chain_spawner.rs
tmpolaczyk: We could, but it will make it a bit harder to revert this change in the future

node/src/container_chain_spawner.rs
tmpolaczyk: We maybe can use `[0, 0, 0]`, but the reason I selected the random ports myself is that if we actually use random ports there is a small probability of occupying a port from the collating container. The way it is implemented now that will only happen if this fixed random port is already taken, which can happen in zombienet because all of the collators try to use the same ports. But in production where every collator runs in isolation, I think it is better to use these fixed random ports instead of actually random.

node/src/container_chain_spawner.rs
girazoki: no worries then

girazoki (approved):

PR #313, opened 2023-11-03
add inflation rewards smoke test

girazoki:
Smoke tests for inflation rewards, only runs against runtimes >300


tmpolaczyk (approved):
Randomness test looks good

PR #314, opened 2023-11-03
Add pallet services payment to runtime

tmpolaczyk:
> Pallet-services payment allow developers to get block production credit. The mechanism we are thinking of releasing as a v0 is to burn a certain amount of tokens per block credit, which this parachain would obtain.

Changes in this PR:

- We only assign collators to parachains that have at least credit for 2 sessions.
- We substract 1 credit every time we see a block with a higher number than the one we have stored. This means that producing more than 1 block in between two tanssi blocks only costs 1 credit. And a parachain that has produced 0 blocks will not be charged any credits, even if it was assigned collators.
- Parachains registered in genesis start with 1000 credits (so we don't worry about credits when testing).
- For dancebox and live networks, there is a migration to give existing parachains enough credits for 60 days.
- Parachains registered afterwards need to purchase credits. The cost is 1 microDANCE for 1 block credit, and a parachain cannot have more than 60 days worth of credits. Anyone can buy credits for a parachain. Credits can be bought even before registering the parachain, as long as you know the para id.


test/suites/rotation-para/test_rotation.ts
```diff
@@ -95,10 +95,10 @@ describeSuite({
                 const keyring = new Keyring({ type: "sr25519" });
                 const alice = keyring.addFromUri("//Alice", { name: "Alice default" });
 
-                const tx1 = await paraApi.tx.configuration.setCollatorsPerContainer(1);
+                const tx1 = await paraApi.tx.configuration.setCollatorsPerContainer(3);
                 const tx2 = await paraApi.tx.configuration.setMinOrchestratorCollators(1);
                 const tx3 = await paraApi.tx.configuration.setMaxOrchestratorCollators(1);
-                const tx4 = await paraApi.tx.configuration.setFullRotationPeriod(5);
+                const tx4 = await paraApi.tx.configuration.setFullRotationPeriod(60);
```
tmpolaczyk, line 101: Undo this

runtime/dancebox/tests/integration_test.rs
```diff
@@ -633,6 +651,240 @@ fn test_authors_paras_inserted_a_posteriori_with_collators_already_assigned() {
...
+                extrinsics_root: Default::default(),
+                digest: sp_runtime::generic::Digest {
+                    logs: vec![DigestItem::PreRuntime(AURA_ENGINE_ID, slot.encode())],
+                },
+            });
+            sproof.items.push(s);
+            // TODO: this seems to break because of the function `relay_chain_state_proof`, which is used to
+            // read randomness from the relay. Panic msg:
+            // 'Invalid relay chain state proof, already constructed in `set_validation_data`: RootMismatch'
+            set_author_noting_inherent_data(sproof);
```
tmpolaczyk, line 864: @girazoki need some help getting this to work, I am successfully mocking the container chain block inclusion, but later when running this line:

```
run_to_session(4u32);
```

it panics with

```
'Invalid relay chain state proof, already constructed in `set_validation_data`: RootMismatch'
```

girazoki: will take a look at this

node/src/chain_spec.rs
```diff
@@ -293,6 +293,27 @@ fn testnet_genesis(
...
+                )
+            })
+        })
+        .chain(
+            mock_container_chains
+                .iter()
+                .map(|x| (*x, mock_container_chain_genesis_data(*x), vec![])),
+        )
+        .collect();
+    // Assign max block credits to all container chains registered in genesis
```
girazoki, line 312: If we assign max, is there any way we test running out of those? I think we should use non-max

pallets/services-payment/src/lib.rs
```diff
@@ -219,3 +224,28 @@ pub trait OnChargeForBlockCredit<T: Config> {
 pub trait ProvideBlockProductionCost<T: Config> {
     fn block_cost(para_id: &ParaId) -> (BalanceOf<T>, Weight);
 }
+
```
girazoki, line 227: the comment of this funciton is wrong

primitives/traits/src/lib.rs
```diff
@@ -110,3 +110,11 @@ pub trait RemoveInvulnerables<AccountId> {
         num_invulnerables: usize,
     ) -> Vec<AccountId>;
 }
+
+/// Helper trait for pallet_collator_assignment to be able to not assign collators to container chains with no credits
+/// in pallet_services_payment
+pub trait RemoveParaIdsWithNoCredits {
+    /// Remove para ids with not enough credits. The resulting order will affect priority: the first para id in the list
+    /// will be the first one to get collators.
```
girazoki, line 118: First meaning the paraId that has higher credits?

runtime/dancebox/src/lib.rs
```diff
@@ -686,6 +691,22 @@ impl RemoveInvulnerables<AccountId> for RemoveInvulnerablesImpl {
     }
 }
 
+pub struct RemoveParaIdsWithNoCreditsImpl;
+
+impl RemoveParaIdsWithNoCredits for RemoveParaIdsWithNoCreditsImpl {
+    fn remove_para_ids_with_no_credits(para_ids: &mut Vec<ParaId>) {
+        let blocks_per_session = Period::get();
+        let credits_for_2_sessions = 2 * blocks_per_session;
```
girazoki, line 699: nice!

runtime/dancebox/src/lib.rs
```diff
@@ -704,13 +726,90 @@ impl pallet_authority_assignment::Config for Runtime {
     type AuthorityId = NimbusId;
 }
 
+pub struct ChargeForBlockCredit<Runtime>(PhantomData<Runtime>);
```
girazoki, line 729: We probably can move this inside pallet-service-payment right?

runtime/dancebox/src/lib.rs
```diff
@@ -704,13 +726,90 @@ impl pallet_authority_assignment::Config for Runtime {
...
+
+        if imbalance.peek() != fee {
+            panic!("withdrawn balance incorrect");
+        }
+
+        Ok(())
+    }
+}
+
+pub const FIXED_BLOCK_PRODUCTION_COST: u128 = 100;
```
girazoki, line 756: this means we will charge 100 tokens

node/src/chain_spec.rs
tmpolaczyk: We test it by registering new container chains. I can set this to a high enough value so that all the tests pass, but I don't see the point.

primitives/traits/src/lib.rs
tmpolaczyk: Could be, now the input order is random and the output order is also random (same as input). I made this trait thinking about the case you mention, so in the future we can reorder the para ids and sort them by number of credits or by priority fee for instance.

runtime/dancebox/src/lib.rs
tmpolaczyk: 0.000...100 actually

node/src/chain_spec.rs
girazoki: in dev mode its easy for us to create blocks. I think we should use a non-max, in the dev mode at least, to test this.

girazoki: lets use something in the order of 1000, we will create 1000 blocks and make sure if you run out of this then you dont get selected

primitives/traits/src/lib.rs
girazoki: good!

runtime/dancebox/src/lib.rs
girazoki: haha I meant 100 units sorry, I will come back with a value for this

runtime/dancebox/src/lib.rs
tmpolaczyk: Moving this to pallet_services_payment adds a dependency to pallet_balances, is that ok?

tmpolaczyk: Nevermind, it does not

runtime/dancebox/tests/integration_test.rs
tmpolaczyk: Done, thank you!

tmpolaczyk:
Missing: remove dev_mode, add benchmarks, add migration that gives credit to existing container chains

pallets/services-payment/src/weights.rs
```diff
@@ -0,0 +1,91 @@
...
+	fn purchase_credits() -> Weight;
+}
+
+/// Weights for pallet_services_payment using the Substrate node and recommended hardware.
+pub struct SubstrateWeight<T>(PhantomData<T>);
+impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
+	/// Storage: `ServicesPayment::BlockProductionCredits` (r:1 w:1)
+	/// Proof: `ServicesPayment::BlockProductionCredits` (`max_values`: None, `max_size`: Some(24), added: 2499, mode: `MaxEncodedLen`)
+	/// Storage: `System::Account` (r:1 w:1)
+	/// Proof: `System::Account` (`max_values`: None, `max_size`: Some(128), added: 2603, mode: `MaxEncodedLen`)
```
tmpolaczyk, line 64: This storage item should probably be whitelisted, right?

girazoki: whcih one SYstem::Account? if so, I dont think so

girazoki:
Test that I would like to see in ts dev tests:
- We can buy additional block credit
- If we run out of credit our chain does not produce blocks because collators are not assigned to it

pallets/services-payment/src/lib.rs
```diff
@@ -214,8 +225,57 @@ pub trait OnChargeForBlockCredit<T: Config> {
     ) -> Result<(), Error<T>>;
 }
 
+pub struct ChargeForBlockCredit<Runtime>(PhantomData<Runtime>);
```
girazoki, line 228: Maybe we should add an extrinsic in this pallet that cleans the existing credits as well for a paraId, only callable by root, what do you think?

pallets/collator-assignment/src/lib.rs
```diff
@@ -161,6 +162,10 @@ pub mod pallet {
             // We get the containerChains that we will have at the target session
             let mut container_chain_ids =
                 T::ContainerChains::session_container_chains(target_session_index);
+            // Remove the containerChains that do not have enough credits for block production
+            T::RemoveParaIdsWithNoCredits::remove_para_ids_with_no_credits(
```
girazoki, line 166: this requires new reads, hence the weight should be corrected

runtime/dancebox/src/lib.rs
```diff
@@ -704,13 +724,65 @@ impl pallet_authority_assignment::Config for Runtime {
...
+pub struct BlockProductionCost<Runtime>(PhantomData<Runtime>);
+impl ProvideBlockProductionCost<Runtime> for BlockProductionCost<Runtime> {
+    fn block_cost(_para_id: &ParaId) -> (u128, Weight) {
+        (FIXED_BLOCK_PRODUCTION_COST, Weight::zero())
+    }
+}
+
+parameter_types! {
+    // 60 days worth of blocks
+    pub const MaxCreditsStored: BlockNumber = 60 * 7200;
```
girazoki, line 738: can we make this a constant? (the 7200 I mean), something like what we have in constants BLOCKS_PER_DAY

runtime/dancebox/src/lib.rs
```diff
@@ -704,13 +724,65 @@ impl pallet_authority_assignment::Config for Runtime {
...
+
+impl AuthorNotingHook<AccountId> for InflationRewardsAndServicesPayment {
+    fn on_container_author_noted(
+        author: &AccountId,
+        block_number: BlockNumber,
+        para_id: ParaId,
+    ) -> Weight {
+        let mut weight = Weight::zero();
+
+        weight = weight.saturating_add(InflationRewards::on_container_author_noted(
```
girazoki, line 764: Nice, now that we are clearly adding up on_containter_author_noted, maybe this could be implemented for tuples and pass (`InflationRewards`, `ServicesPayment`)

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Actually now the benchmarks are broken because this function removes all the registered chains...

girazoki: so we should add a function to `RemoveParaIdsWithNoCredits` with runtime-benchmarks, which sets the credit

runtime/dancebox/src/lib.rs
tmpolaczyk: It already existed, its called DAYS

runtime/dancebox/src/lib.rs
tmpolaczyk: I thought about that, but this way will be easier to maintain, in case we want some additional logic such as only executing a hook if the other one succeeded.

Unless there's some macro that implements a trait for tuples and will make this more readable.

pallets/collator-assignment/src/lib.rs
tmpolaczyk: But I believe the value used for the weight hint is wrong, it uses the number of parachains after filtering out the ones that do not have enough credit:

```rust
            let assigned_collators = Self::assign_collators(session_index, random_seed, collators);
            let num_parachains = assigned_collators.next_assignment.container_chains.len();
            frame_system::Pallet::<T>::register_extra_weight_unchecked(
                T::WeightInfo::new_session(num_collators as u32, num_parachains as u32),
                DispatchClass::Mandatory,
            );
```

Instead it should use the initial number of parachains, I guess I will add a new return param to the `assign_collators`  function.

runtime/dancebox/src/lib.rs
girazoki: there is this one yes: https://github.com/paritytech/polkadot-sdk/blob/29654a4d7154296e8e6cb2a067896490d67d96a1/polkadot/xcm/xcm-builder/src/fee_handling.rs#L39

girazoki: We will need to see wether it goes one by one or wether it returns in the first successful one

girazoki: I think this one is close to what you want: https://github.com/paritytech/polkadot-sdk/blob/29654a4d7154296e8e6cb2a067896490d67d96a1/polkadot/runtime/parachains/src/paras/mod.rs#L500

pallets/collator-assignment/src/lib.rs
tmpolaczyk: (done)

girazoki (approved):
LGTM, good job!

PR #316, opened 2023-11-07
Refactor common code among clients

nanocryk:
Adds a new crate which factors out common code between all clients, and allow to easily add node specific services.


client/node-common/src/service.rs
```diff
@@ -0,0 +1,652 @@
...
+    [ConstructedRuntimeApi] => {
+        <RuntimeApi as ConstructRuntimeApi<Block, T![Client]>>::RuntimeApi
+    };
+    [Where] => {
+        Block: cumulus_primitives_core::BlockT,
+        ParachainNativeExecutor: NativeExecutionDispatch + 'static,
+        RuntimeApi: ConstructRuntimeApi<Block, T![Client]> + Sync + Send + 'static,
+        T![ConstructedRuntimeApi]: TaggedTransactionQueue<Block> + BlockBuilder<Block>,
+    }
+}
```
tmpolaczyk, line 76: This macro looks like a bad idea, I'm mostly worried about the error messages a user will see if they pass some wrong input type (I don't know if the macro will be expanded in the error message or not), as well as the errors we will encounter on the next polkadot upgrade.

Maybe instead we can take a similar approach to the one used in `BlockNumberFor` https://paritytech.github.io/polkadot-sdk/master/frame_system/pallet_prelude/type.BlockNumberFor.html

So instead of making the functions generic over `Block`, `RuntimeApi`, and `ParachainNativeExecutor`, make them generic over `Runtime`, or something else that can be used to derive all the other types. So we would have `ClientFor<R>` instead of `T![Client]`. It may make this module a bit more verbose, but then I believe this crate will be easier to integrate with other projects.

nanocryk: In error messages it gives verbose names. The "config trait" approach can be used instead indeed. I'll make the change.

node/src/service.rs
```diff
@@ -115,6 +122,27 @@ thread_local!(static TIMESTAMP: std::cell::RefCell<u64> = std::cell::RefCell::ne
 /// Provide a mock duration starting at 0 in millisecond for timestamp inherent.
 /// Each call will increment timestamp by slot_duration making Aura think time has passed.
 struct MockTimestampInherentDataProvider;
```
tmpolaczyk, line 124: This can be moved to node_common as well

client/node-common/src/service.rs
```diff
@@ -0,0 +1,750 @@
...
+        )
+        .await
+        .map_err(|e| sc_service::Error::Application(Box::new(e) as Box<_>))
+    }
+
+    /// Given an import queue, calls `cumulus_client_service::build_network` and
+    /// stores the returned objects in `self.network` and `self.tx_handler_controller`.
+    ///
+    /// Can only be called once on a `NodeBuilder` that doesn't have yet network
+    /// data.
```
tmpolaczyk, line 285: I really like this, I knew it can be done using const generics and a state machine enum, but I didn't know about the `TypeIdentity` trick, will need to look into it.

nanocryk: This is the "Typestate Pattern" described here: http://cliffle.com/blog/rust-typestate/

Usually you need an `impl` block per typestate, but `core_extensions::TypeIdentity` allows to state that a type parameter is equal to a concrete type and convert between both, which allow to have a single `impl` block with `where` clauses on the functions. (except for `new`, as it doesn't take `self` Rust is unable to infer which type to use, even if there is only 1 type `T` implementing `TypeIdentity<Type = T>`)

client/node-common/src/service.rs
```diff
@@ -0,0 +1,869 @@
...
+                telemetry.as_ref().map(|(_, telemetry)| telemetry.handle()),
+                executor,
+            )?;
+        let client = Arc::new(client);
+
+        let telemetry_worker_handle = telemetry.as_ref().map(|(worker, _)| worker.handle());
+
+        let telemetry = telemetry.map(|(worker, telemetry)| {
+            task_manager
+                .spawn_handle()
```
girazoki, line 224: Does not this spawn telemetry?

client/node-common/src/service.rs
```diff
@@ -0,0 +1,869 @@
...
+        })
+    }
+
+    /// Given an import queue, calls `sc_service::build_network` and
+    /// stores the returned objects in `self.network` and `self.tx_handler_controller`.
+    ///
+    /// Can only be called once on a `NodeBuilder` that doesn't have yet network
+    /// data.
+    #[must_use]
+    pub fn build_substrate_network(
```
girazoki, line 371: So I cannot call together both `build_substrate_network` and `build_cumulus_network` right? I wonder whether this can be an enum as an argument, where in the case of type:

```
pub struct CumulusNetwork {
para_id: ParaId,
relay_chain_interface: RelayChainInterface
}
pub enum NetworkType {
Substrate,
Cumulus(CumulusNetwork)
}
```

girazoki:
Closing as this will be done in https://github.com/moondance-labs/tanssi/pull/316

client/node-common/src/service.rs
nanocryk: This enum would need to be generic over the RelayChainInterface type, which means that in Substrate mode it will not be able to infer the type and will require to explicitly name one.

client/node-common/src/service.rs
nanocryk: It does. Why?

node/src/service.rs
nanocryk: It depends on the `SLOT_DURATION` in the runtimes, but we can probably be generic over it.

nanocryk:
Left-over for future PRs:
- Add dev service to simple template
- Maybe factor out common command stuff?

tmpolaczyk:
Something in `container_chain_spawner.rs` is not right, the test `pnpm moonwall test zombie_tanssi_keep_db` is failing. Looking into it.

tmpolaczyk (approved):
tanssi-node and `container_chain_spawner` look good, I didn't check the templates.

container-chains/templates/simple/node/src/service.rs
```diff
@@ -64,118 +49,49 @@ impl sc_executor::NativeExecutionDispatch for ParachainNativeExecutor {
...
-    });
-
-    let transaction_pool = sc_transaction_pool::BasicPool::new_full(
-        config.transaction_pool.clone(),
-        config.role.is_authority().into(),
-        config.prometheus_registry(),
-        task_manager.spawn_essential_handle(),
-        client.clone(),
-    );
+pub struct NodeBuilderConfig;
```
tmpolaczyk, line 56:
```suggestion
pub struct NodeConfig;
```

Not a good idea to call the struct the same as the trait, same in frontier template

girazoki: +1, agree, I think the rest looks good

tmpolaczyk (approved):

girazoki (approved):

tmpolaczyk:
Can you resolve conflicts and merge? Thanks

tmpolaczyk (approved):

PR #319, opened 2023-11-08
Prefund pallet accounts in genesis to ensure they are always alive

tmpolaczyk:
This is needed for the StakingAccount, not sure if the other ones need it as well

### DIFF
--- a/test/suites/dev-tanssi/staking/test_staking_rewards_balanced.ts
+++ b/test/suites/dev-tanssi/staking/test_staking_rewards_balanced.ts
@@ -134,8 +134,8 @@ describeSuite({
                 const events = await polkadotJs.query.system.events();
                 const reward = await fetchRewardAuthorOrchestrator(events);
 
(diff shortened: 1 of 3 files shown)


test/suites/dev-tanssi/staking/test_staking_rewards_balanced.ts
```diff
@@ -134,8 +134,8 @@ describeSuite({
                 const events = await polkadotJs.query.system.events();
                 const reward = await fetchRewardAuthorOrchestrator(events);
 
-                // 20% plus 1 from rounding
-                const collatorPercentage = (20n * reward.balance.toBigInt()) / 100n + 1n;
```
tmpolaczyk, line 138: This +1 is not needed after initializing the staking account with the existential deposit, not sure why because it is only +1, not +1_000_000_000 which is the ED

girazoki (approved):

PR #322, opened 2023-11-10
Add precompiles and tests to evm template

fgamundi:
Adds these EVM precompiled smart contracts from Moonkit:

- pallet-balances
- batch
- call-permit
- xcm-utils

Depends on Moonkit's https://github.com/Moonsong-Labs/moonkit/pull/11


container-chains/templates/frontier/runtime/src/migrations.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        let db_weights = T::DbWeight::get();
+
+        let mut count = 0u64;
+
+        for address in Precompiles::used_addresses() {
+            pallet_evm::Pallet::<T>::create_account(address.into(), revert_bytecode.clone());
+            count += 1;
+        }
+        db_weights.reads_writes(0, count)
```
fgamundi, line 52: I wasn't sure what weight to return

tmpolaczyk: Migration weights are not important until we have multi-block migrations, so this is fine.

tmpolaczyk: But looking at the source code of create account I count 2 writes and 1 read, so maybe you can use that?

runtime/dancebox/Cargo.toml
```diff
@@ -234,7 +234,6 @@ try-runtime = [
 	"pallet-invulnerables/try-runtime",
 	"pallet-maintenance-mode/try-runtime",
 	"pallet-migrations/try-runtime",
-	"pallet-migrations/try-runtime",
```
girazoki, line 237: reason for this?

fgamundi: It's duplicated, right? The line above is the same

girazoki: ah right

Cargo.toml
```diff
@@ -194,31 +198,32 @@ cumulus-test-relay-sproof-builder = { git = "https://github.com/moondance-labs/p
...
+fc-cli = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", default-features = false }
+fc-consensus = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", default-features = false }
+fc-db = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", default-features = false }
+fc-mapping-sync = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", default-features = false }
+fc-rpc = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", features = [
 	"rpc-binary-search-estimate",
 ] }
-fc-rpc-core = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0", default-features = false }
-fc-storage = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0", default-features = false }
+fc-rpc-core = { git = "https://github.com/moondance-labs/frontier", branch = "tanssi-polkadot-v1.1.0-xcm-util", default-features = false }
```
girazoki, line 225: the last piece missing, just cherry-pick this in our tanssi-polkadot-v1.1.0 branch

girazoki (approved):
approved after fixing comments

container-chains/templates/frontier/runtime/src/precompiles.rs
```diff
@@ -14,71 +14,89 @@
...
+        Erc20BalancesPrecompile<R, NativeErc20Metadata>,
+        (CallableByContract, CallableByPrecompile),
+    >,
+    PrecompileAt<
+        AddressU64<2049>,
+        BatchPrecompile<R>,
+        (
+            SubcallWithMaxNesting<2>,
+            // Batch is the only precompile allowed to call Batch.
+            CallableByPrecompile<OnlyFrom<AddressU64<2050>>>,
```
tmpolaczyk, line 84:
```suggestion
            CallableByPrecompile<OnlyFrom<AddressU64<2049>>>,
```

tmpolaczyk: Doesn't this have a test?

girazoki: good catch, probably not. @fgamundi can you add a test for this?

fgamundi: `CallableByPrecompile` rule removed as it isn't needed for batch to call itself. Batch runs the inner calls without changing the original `caller`, so it doesn't count as coming from a precompile. It's the same case for CallPermit.
Added tests for Batch calling itself and CallPermit calling Batch

PR #325, opened 2023-11-16
Implement pallet_registrar hooks

tmpolaczyk:
Allows to execute code on container chain register and deregister. This PR only includes the deregister hooks:

* Clear Registrar storage: BootNodes and ParaGenesisData, and return parachain deposit (this used to be done immediately, now it will be done after 2 sessions)
* Call AuthorNoting::kill_author_data
* Call ServicesPayment::set_credits(0)

These hooks are executed 2 sessions after calling deregister, which is when the container chain won't have any collators anymore. Unless the para id was never marked as valid_for_collating, in that case the hooks are executed immediately.

Also, change the pausing mechanism. Add a new extrinsic to unpause: `unpause_container_chain`. Previously `mark_valid_as_collating` was used to unpause, now that won't work anymore. This change was needed because otherwise pausing and unpausing could trigger the implemented hooks when it should not, so now paused container chains are stored in a new storage item `Paused` instead of being in `PendingVerification`. Container chains that were paused before this PR will need to be unpaused using `mark_valid_for_collating`.


pallets/registrar/src/lib.rs
```diff
@@ -562,11 +745,108 @@ pub mod pallet {
...
+        fn cleanup_deregistered_para_id(para_id: ParaId) {
+            ParaGenesisData::<T>::remove(para_id);
+            BootNodes::<T>::remove(para_id);
+            T::RegistrarHooks::para_deregistered(para_id);
+        }
+
+        fn schedule_parachain_cleanup(para_id: ParaId) -> DispatchResult {
+            let mut pending_paras = PendingToRemove::<T>::get();
+            // First, we need to decide what we should use as the base paras.
+            let mut base_paras = pending_paras
```
tmpolaczyk, line 818: This is wrong, add a test where we deregister 2 container chains, 1 in session 2 and 1 in session 3:

```
State after deregister(2000):
(2, [2000])

Wait 1 session
State after deregister(2001):
(2, [2000])
(3, [2000, 2001])
```

Container chain 2000 will be cleaned up twice. Also add a test where they are deregistered in the same session.

pallets/registrar/src/weights.rs
```diff
@@ -55,247 +63,301 @@ pub trait WeightInfo {
...
-			// Standard Error: 16_866
-			.saturating_add(Weight::from_parts(391_115, 0).saturating_mul(y.into()))
-			.saturating_add(T::DbWeight::get().reads(4_u64))
+		//  Measured:  `477 + y * (8 ±0)`
+		//  Estimated: `1958 + y * (8 ±0)`
+		// Minimum execution time: 27_650_000 picoseconds.
+		Weight::from_parts(42_754_882, 1958)
+			// Standard Error: 9_207
+			.saturating_add(Weight::from_parts(276_708, 0).saturating_mul(y.into()))
+			.saturating_add(T::DbWeight::get().reads(6_u64))
```
tmpolaczyk, line 171: There are 2 reads more here because we read Paused and PendingPaused, which is not needed. I will add back the function `schedule_parachain_change` and use it in `mark_valid_for_collating`.

pallets/registrar/src/lib.rs
tmpolaczyk: Done, in `deregister_2_container_chains_in_consecutive_sessions` and `deregister_2_container_chains_in_same_block`

pallets/registrar/src/weights.rs
```diff
@@ -55,247 +63,301 @@ pub trait WeightInfo {
...
+		// Proof Size summary in bytes:
+		//  Measured:  `454 + y * (4 ±0)`
+		//  Estimated: `1937 + y * (4 ±0)`
+		// Minimum execution time: 25_398_000 picoseconds.
+		Weight::from_parts(32_217_636, 1937)
+			// Standard Error: 0
+			.saturating_add(Weight::from_parts(2, 0).saturating_mul(x.into()))
+			// Standard Error: 8_573
+			.saturating_add(Weight::from_parts(242_112, 0).saturating_mul(y.into()))
+			.saturating_add(T::DbWeight::get().reads(6_u64))
```
tmpolaczyk, line 146: This is missing 1 read for `RegisteredParaIds`, need to fix benchmark.

tools/benchmarking.sh
```diff
@@ -7,7 +7,7 @@
 
 set -e
 
-BINARY="./target/production/tanssi-node"
+BINARY="./target/release/tanssi-node"
```
tmpolaczyk, line 10: undo

pallets/registrar/src/tests.rs
```diff
@@ -559,28 +666,344 @@ fn deregister_returns_bond() {
...
+            Some(&empty_genesis_data())
+        );
+        assert_ok!(ParaRegistrar::deregister(RuntimeOrigin::root(), 42.into(),));
+        assert_ok!(ParaRegistrar::deregister(RuntimeOrigin::root(), 43.into(),));
+        assert_eq!(
+            Mock::get().called_hooks,
+            vec![
+                HookCall::Registered(42.into()),
+                HookCall::Registered(43.into()),
+            ]
```
tmpolaczyk, line 877: Maybe add a check that this hooks are called as part of the `mark_valid_for_collating` extrinsic.

tools/benchmarking.sh
girazoki: yes please

pallets/registrar/src/lib.rs
```diff
@@ -443,34 +562,59 @@ pub mod pallet {
...
-        /// Pause container-chain from collating without removing its boot nodes nor its genesis config
+        /// Pause or unpause container-chain from collating. Does not remove its boot nodes nor its genesis config.
+        /// Only container-chains that have been marked as valid_for_collating can be paused.
         #[pallet::call_index(4)]
         #[pallet::weight(T::WeightInfo::pause_container_chain(T::MaxLengthParaIds::get()))]
-        pub fn pause_container_chain(origin: OriginFor<T>, para_id: ParaId) -> DispatchResult {
+        pub fn pause_container_chain(
+            origin: OriginFor<T>,
+            para_id: ParaId,
+            pause: bool,
```
girazoki, line 572: isnt this boolean implicit in the extrinsic?

girazoki: hmm I see if pause if false then it means unpause. I would add two extrinsics for this, I think its more desdcriptive and usable

tmpolaczyk: Good idea, I agree

runtime/dancebox/src/lib.rs
```diff
@@ -821,6 +822,38 @@ impl pallet_configuration::Config for Runtime {
     type WeightInfo = pallet_configuration::weights::SubstrateWeight<Runtime>;
 }
 
+pub struct RegistrarHooksImpl;
+
+impl RegistrarHooks for RegistrarHooksImpl {
```
girazoki, line 827:
```suggestion
impl RegistrarHooks forDanceboxRegistrarHooks {
```

test/suites/dev-tanssi/registrar/test_registrar_deregister.ts
```diff
@@ -107,8 +107,8 @@ describeSuite({
 
                 // Check the number of keys in storage
                 const palletKeys = await polkadotJs.rpc.state.getKeys("0x3fba98689ebed1138735e0e7a5a790ab");
-                // 3 keys: version, registeredParas, pendingParas
-                expect(palletKeys.length).to.be.eq(3);
+                // 4 keys: version, registeredParas, pendingParas, pendingToRemove
+                expect(palletKeys.length).to.be.eq(4);
```
girazoki, line 111: I would add here a check that authorNoting gets cleaned, as well as credits. This should be easily doable by setting both storage items with a extrinsic

tools/benchmarking.sh
girazoki: undo :)

girazoki:
Ideally I would like to see tests that capture the following cases:

- a parachain is pending to be a registered para and we deregister it. In this case we should not call the hook immediately
- a prachain is pending to be paused and we deregister it. In this case, we should not call the hook immediately

girazoki (approved):

pallets/registrar/src/lib.rs
```diff
@@ -244,6 +273,100 @@ pub mod pallet {
         NotSufficientDeposit,
     }
 
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        #[cfg(feature = "std")]
+        fn integrity_test() {
```
tmpolaczyk, line 279:
```suggestion
        fn try_state(_block_number: u32) {
```

The `integrity_test` hook doesn't have access to the storage, despite docs saying otherwise.

fgamundi (approved):
LGTM

PR #329, opened 2023-11-20
rework coverage

girazoki:


tmpolaczyk:
Maybe add a limit to the files changed? So that by default it shows the first 5 or 6, and then you have a "details" tag to expand the rest

girazoki:
I have to ignore certain files first

tmpolaczyk (approved):
Good enough

PR #335, opened 2023-11-21
Ignore panic when node is assigned to para with no genesis data

tmpolaczyk:
```
// This is triggered when:
// 1. Collator is assigned to parachain 2000
// 2. Parachain 2000 is deregistered. This deletes the genesis data immediately.
// 3. Collator restarts parachain 2000. Panics because that container chain does not have any genesis data.
// Fixed by not deleting genesis data immediately in https://github.com/moondance-labs/tanssi/pull/325
```

Can be tested using `pnpm moonwall test zombie_tanssi_keep_db`, the test fails because it expects collators to keep producing blocks and they won't produce blocks, but at least they don't panic.

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -365,6 +365,14 @@ impl ContainerChainSpawner {
         async move {
             match try_closure().await {
                 Ok(()) => {}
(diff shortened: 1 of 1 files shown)


girazoki (approved):

girazoki:
needs to be merged against a branch perm-v0.3.2 where the node versions are incremented

PR #338, opened 2023-11-23
Give free credits once on container chain register

tmpolaczyk:
Adds a hook on container chain register that gives free credits to the container chain. Container chains that existed before this was implemented will also be given free credits using a migration.


pallets/services-payment/src/lib.rs
```diff
@@ -114,6 +114,11 @@ pub mod pallet {
     pub type BlockProductionCredits<T: Config> =
         StorageMap<_, Blake2_128Concat, ParaId, BlockNumberFor<T>, OptionQuery>;
 
+    /// List of para ids that have already been given free credits
+    #[pallet::storage]
+    #[pallet::getter(fn given_free_credits)]
+    pub type GivenFreeCredits<T: Config> = StorageMap<_, Blake2_128Concat, ParaId, (), OptionQuery>;
```
tmpolaczyk, line 120: Add a few tests for this:

* After migration, calling mark_valid_for_collating on a paused para id does not give it more free credits
* Register, mark_valid_for_collating, deregister, register again, mark_valid_for_collating again, should not receive free credits the 2nd time

pallets/services-payment/src/lib.rs
```diff
@@ -211,6 +216,29 @@ pub mod pallet {
 
             Ok(().into())
         }
+
+        pub fn give_free_credits(para_id: &ParaId) {
+            if GivenFreeCredits::<T>::contains_key(para_id) {
+                // This para id has already received free credits
+                return;
```
girazoki, line 223: We likely need to account weight for this, so we should at least return the weight used

pallets/services-payment/src/lib.rs
girazoki: pretty much yeah, also if I inject my own free-credits I at most get the max for free (it does not add up)

girazoki: As always, I would add a sudo extrinsic to clean this storage

tmpolaczyk: All done

pallets/services-payment/src/lib.rs
tmpolaczyk: Done but I set the weight to Weight::default because it is not accounted for anywhere yet

pallets/services-payment/src/weights.rs
```diff
@@ -82,6 +83,11 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 		Weight::from_parts(7_268_000, 0)
 			.saturating_add(T::DbWeight::get().writes(1_u64))
 	}
+	fn set_given_free_credits() -> Weight {
+		// TODO: run this benchmark, I just copied the values from above
```
girazoki, line 87: dont forget to do this

pallets/services-payment/src/lib.rs
```diff
@@ -188,6 +193,26 @@ pub mod pallet {
 
             Ok(().into())
         }
+
+        /// Helper to set and cleanup the `GivenFreeCredits` storage.
+        /// Can only be called by root.
+        #[pallet::call_index(2)]
+        #[pallet::weight(T::WeightInfo::set_given_free_credits())]
+        pub fn set_given_free_credits(
```
girazoki, line 201:
```suggestion
        pub fn clear_given_free_credits(
```

girazoki (approved):
Typescript tests please!

pallets/services-payment/src/lib.rs
tmpolaczyk: No, it is part of the mark_valid_for_collating extrinsic

girazoki: then we probably should find the way to benchmark it and account in such extrinsic. Probably through a trait?

tmpolaczyk: When we run the benchmarking script it is already taken into account.

fgamundi (approved):

girazoki:
Let's add one more test testing the sudo call that cleans the assigned credit flag

girazoki (approved):

PR #339, opened 2023-11-24
Ugprade moonwall to 4.4.2 and chopsticks templates

girazoki:
UPgrade moonwall to 4.4.2 beta, which does not contain the zombienet bug

Additionally it adds chopsticks support for templates


package.json
```diff
@@ -0,0 +1,7 @@
+ {
```
girazoki, line 1: Remove this once we have a moonwall version that has the updated package

girazoki:
A mistery on why the percentage drops like this..

test/suites/warp-sync/test_warp_sync.ts
```diff
@@ -249,11 +249,10 @@ async function directoryExists(directoryPath) {
 
 /// Returns the /tmp/zombie-52234... path
 function getTmpZombiePath() {
-    const logFilePath = process.env.MOON_MONITORED_NODE;
+    const zombiePath = process.env.MOON_ZOMBIE_DIR;
```
tmpolaczyk, line 252: This function is now useless (and that's good)

```suggestion
    return process.env.MOON_ZOMBIE_DIR;
```

PR #341, opened 2023-11-27
Implement pallet-data-preservers

tmpolaczyk:
Implement V0 of pallet-data-preservers, will not support data providers or payments yet. The container chain manager (or root) is expected to set the list of bootnodes manually.

Also, change mark_valid_for_collating extrinisic, now it will fail if the container chain does not have any bootnodes.


pallets/data-preservers/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+//! # Inflation Rewards Pallet
+//!
+//! This pallet handle native token inflation and rewards distribution.
```
tmpolaczyk, line 19: No

pallets/data-preservers/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    #[pallet::pallet]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(core::marker::PhantomData<T>);
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// Overarching event type.
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+        type Currency: Inspect<Self::AccountId> + Balanced<Self::AccountId>;
+        type ContainerChainManagerOrRootOrigin: EnsureOriginWithArg<Self::RuntimeOrigin, ParaId>;
```
tmpolaczyk, line 67:
```suggestion
        type SetBootNodesOrigin: EnsureOriginWithArg<Self::RuntimeOrigin, ParaId>;
```

node/src/chain_spec.rs
```diff
@@ -314,6 +314,15 @@ fn testnet_genesis(
         .iter()
         .map(|(para_id, _genesis_data, _boot_nodes)| (*para_id, 1000))
         .collect();
+    let para_id_boot_nodes: Vec<_> = para_ids
+        .iter()
+        .map(|(para_id, _genesis_data, boot_nodes)| (*para_id, boot_nodes.clone()))
+        .collect();
+    let para_ids: Vec<_> = para_ids
```
girazoki, line 321: is this to filter those not having bootnodes? if so please right a comment

pallets/data-preservers/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+                    .iter()
+                    .map(|x| BoundedVec::try_from(x.clone()).expect("boot node url too long"))
+                    .collect();
+                let boot_nodes = BoundedVec::try_from(boot_nodes).expect("too many boot nodes");
+                <BootNodes<T>>::insert(para_id, boot_nodes);
+            }
+        }
+    }
+
+    /// Inflation rewards pallet.
```
girazoki, line 89:
```suggestion
    /// Data preservers pallet.
```

pallets/data-preservers/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+        ) -> DispatchResult {
+            T::SetBootNodesOrigin::ensure_origin(origin, &para_id)?;
+
+            BootNodes::<T>::insert(para_id, boot_nodes);
+
+            Self::deposit_event(Event::BootNodesChanged { para_id });
+
+            Ok(())
+        }
+    }
```
girazoki, line 146: Ideally we should have a sudo-only extrinsic too to removeº

pallets/data-preservers/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    impl<T: Config> Pallet<T> {
```
girazoki, line 148: We probably need a trait here that is able to access bootnodes for a paraId

runtime/dancebox/src/lib.rs
```diff
@@ -767,6 +772,105 @@ impl pallet_services_payment::Config for Runtime {
     type WeightInfo = pallet_services_payment::weights::SubstrateWeight<Runtime>;
 }
 
+pub struct DanceboxContainerChainManagerOrRootOrigin<T, RootOrigin> {
```
girazoki, line 775: It might be easier to implement something like this in the pallet, and wrap it in a struct: https://github.com/paritytech/polkadot-sdk/blob/1f2ccaea05c8508f5cc916f8ae0ba60c102c5cf3/polkadot/runtime/common/src/paras_registrar/mod.rs#L540

runtime/dancebox/src/lib.rs
```diff
@@ -849,9 +953,36 @@ impl RegistrarHooks for DanceboxRegistrarHooks {
...
         }
+        // Remove bootnodes from pallet_data_preservers
+        DataPreservers::para_deregistered(para_id);
 
         Weight::default()
     }
+
+    fn check_valid_for_collating(para_id: ParaId) -> DispatchResult {
+        // To be able to call mark_valid_for_collating, a container chain must have bootnodes
+        if DataPreservers::boot_nodes(para_id).len() > 0 {
```
girazoki, line 964: You can probably add this as a trait that is implemented for the data-preservers palet, and you pass this as config in the registrar pallet. That way you can return an error of the pallet itself

node/src/chain_spec.rs
tmpolaczyk: No, it's because I removed the bootnodes from the registrar genesis config, now bootnodes are part of the data preservers genesis.

runtime/dancebox/src/lib.rs
girazoki: @tmpolaczyk check 160e00c2f201680ac579c73b0cbe66982c077212, revert it if you find it less clean, I do find it cleaner than what we had

girazoki (approved):
LGTM! if benchmarks are still to run please do

PR #342, opened 2023-11-27
Upgrade 1.3.0

girazoki:
Upgrades dependencies to 1.3.0


runtime/dancebox/src/lib.rs
```diff
@@ -386,15 +386,6 @@ parameter_types! {
     pub const ExistentialDeposit: Balance = EXISTENTIAL_DEPOSIT;
 }
 
-/// A reason for placing a hold on funds.
```
girazoki, line 389: TODO: check in Fork tests this does not create issues with existing holds!

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -788,6 +789,7 @@ parameter_types! {
     pub BlockGasLimit: U256 = U256::from(BLOCK_GAS_LIMIT);
     pub PrecompilesValue: TemplatePrecompiles<Runtime> = TemplatePrecompiles::<_>::new();
     pub WeightPerGas: Weight = Weight::from_parts(weight_per_gas(BLOCK_GAS_LIMIT, NORMAL_DISPATCH_RATIO, WEIGHT_MILLISECS_PER_BLOCK), 0);
+    pub SuicideQuickClearLimit: u32 = 0;
```
girazoki, line 792: check what this implies!

girazoki: it is to delete the storage associated with the contracts up to the limit: https://github.com/paritytech/frontier/pull/1221.

Setting it to 0 maintains the same behavior as before, but for future releases, maybe we should think about setting this value

runtime/dancebox/src/lib.rs
girazoki: We need a migration for this, running the fork tests has made me realize that the enums are not mapped equally!

girazoki: Migration added already

tmpolaczyk (approved):

runtime/dancebox/tests/common/xcm/foreign_signed_based_sovereign.rs
```diff
@@ -50,6 +50,10 @@ fn using_signed_based_sovereign_works_in_tanssi() {
     };
 
     let xcm = VersionedXcm::from(Xcm(vec![
+        DescendOrigin(X1(AccountId32 {
```
girazoki, line 53: This sucks but westend no longer lets you send a message form a signed origin

runtime/dancebox/src/migrations.rs
```diff
@@ -375,6 +375,125 @@ where
...
+                    <T as pallet_balances::Config>::RuntimeHoldReason,
+                    <T as pallet_balances::Config>::Balance,
+                > = pallet_balances::IdAmount {
+                    id: pallet_pooled_staking::HoldReason::PooledStake.into(),
+                    amount: hold.amount,
+                };
+                new_holds.push(new_item);
+            }
+            let bounded = BoundedVec::<_, T::MaxHolds>::truncate_from(new_holds.clone());
+            pallet_balances::Holds::<T>::insert(&account_id, bounded);
```
fgamundi, line 436: Should the old data be cleared from storage?

girazoki: we are writing into the same storage, but we are just overriding it with the new type

fgamundi (approved):

PR #343, opened 2023-11-28
Add dev service to simple template

nanocryk:
Add missing dev service to simple template, and fix runtime and common tests to make them pass for both templates.


test/moonwall.config.json
```diff
@@ -65,6 +65,25 @@
                 ]
             }
         },
+        {
+            "name": "dev_simple_template",
```
tmpolaczyk, line 69: Do we want to run this in CI? My guess is no, because we almost never modify this template, and we already test the frontier template which should be very similar anyway.

girazoki: I would run it to be honest, in increases the coverage

container-chains/templates/simple/node/src/service.rs
```diff
@@ -145,3 +183,139 @@ pub async fn start_parachain_node(
...
+    };
+
+    let node_builder = node_builder.spawn_common_tasks(parachain_config, rpc_builder)?;
+
+    // let relay_chain_slot_duration = Duration::from_secs(6);
+    // let node_builder = node_builder.start_full_node(
+    //     para_id,
+    //     relay_chain_interface.clone(),
+    //     relay_chain_slot_duration,
+    // )?;
```
fgamundi, line 314:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -355,6 +355,40 @@ jobs:
                   chmod uog+x target/release/container-chain-template-frontier-node
                   cd test
                   pnpm moonwall test dev_frontier_template
+      
+      typescript-tests-simple-template:
```
fgamundi, line 359:
```suggestion
    typescript-tests-simple-template:
```

fgamundi: I think this is what's preventing the CI to run

girazoki (approved):

PR #345, opened 2023-11-28
[Experimental] [Do not merge] Runtime config for async backing 

Agusrodri:
The goal of this draft PR is to start testing and trying different configs to see how the current node (without async-backing changes) reacts to a runtime that already has some basic async-backing configuration within it.

(And of course most of the tests will break at the beginning)


container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -902,7 +916,7 @@ impl_runtime_apis! {
 }
 
 struct CheckInherents;
-
+#[allow(deprecated)]
```
Agusrodri, line 919: TODO: re-check. The warning suggests removing `CheckInherents` struct. We should be careful about the implications of this.

PR #349, opened 2023-11-30
Add .gitattributes file to hide typescript-api in diffs

tmpolaczyk:
Ideally I would like those files to not affect the PR size (so that the +/- reflects actual changes, not autogenerated ones), but I don't know if that's possible.


tmpolaczyk:
I'll be using this extension, looks good, I hope it's not malware https://github.com/aklinker1/github-better-line-counts

girazoki (approved):

PR #358, opened 2023-12-04
Add Registrar proxy type

nanocryk:
Adds new `Registrar` and `SudoRegistrar` proxy types.


tmpolaczyk:
Let's add a typescript test, I guess it will be a combination of these 2:

test/suites/common/proxy/test-proxy-balances.ts
test/suites/dev-tanssi/registrar/test_registrar_register.ts

test/suites/dev-tanssi/registrar/test_registrar_proxy.ts
```diff
@@ -0,0 +1,123 @@
...
+                    return g;
+                };
+                const containerChainGenesisData = emptyGenesisData();
+
+                const tx2 = polkadotJs.tx.proxy.proxy(
+                    alice.address,
+                    null,
+                    polkadotJs.tx.registrar.register(2002, containerChainGenesisData)
+                );
+                const tx3 = polkadotJs.tx.registrar.markValidForCollating(2002);
```
girazoki, line 85: Isnt it sufficient for the test that we test only that the genesis data gets set? the rest is made with sudo

test/suites/dev-tanssi/registrar/test_registrar_proxy.ts
```diff
@@ -0,0 +1,107 @@
...
+                    null,
+                    polkadotJs.tx.registrar.setBootNodes(2002, ["dummy"])
+                );
+                await context.createBlock([await tx3.signAsync(charlie)]);
+
+                // Check that the on chain genesis data is set correctly
+                const onChainBootnodes = await polkadotJs.query.registrar.bootNodes(2002);
+                // TODO: fix once we have types
+                expect(onChainBootnodes.toHuman()).to.deep.equal(["dummy"]);
+            },
```
tmpolaczyk, line 104: This test could also check that the registrar deposit is set to alice when charlie registers the chain using proxy. And also add a test for sudoRegistrar, it should be easy, try calling sudo(markValidForCollating) through the proxy.

runtime/dancebox/src/lib.rs
```diff
@@ -903,6 +903,10 @@ pub enum ProxyType {
     CancelProxy = 4,
     /// Allow extrinsic related to Balances.
     Balances = 5,
+    /// Allow extrinsics related to Registrar
+    Registrar = 6,
+    /// Allow extrinsics related to Registrar that needs to be called through Sudo
+    SudoRegistrar = 8,
```
tmpolaczyk, line 909: Why 8?

```suggestion
    SudoRegistrar = 7,
```

runtime/dancebox/src/lib.rs
```diff
@@ -940,6 +944,16 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
                 matches!(c, RuntimeCall::Balances(..) | RuntimeCall::Utility(..))
             }
+            ProxyType::Registrar => {
+                matches!(c, RuntimeCall::Registrar(..) | RuntimeCall::Utility(..))
+            }
+            ProxyType::SudoRegistrar => {
+                matches!(
+                    c,
+                    RuntimeCall::Sudo(pallet_sudo::Call::sudo {call: ref x})
+                    if matches!(x.as_ref(), RuntimeCall::Registrar(..))
```
tmpolaczyk, line 954: Compared to SudoBalances this is missing RuntimeCall::Utility, but I guess that's good, we don't want the proxy to be able to call sudo(utility.dispatchAs)... But what about allowing utility.batch*, that should be ok?

girazoki: I am actually not entirely sure what would happen if we enable utility here, but I see that parity has it enabled in a similar proxy type: https://github.com/paritytech/polkadot-sdk/blob/4c4407a8930267e139db93019c01444acf49e1ce/polkadot/runtime/westend/src/lib.rs#L1066

@nanocryk can you try it out?

tmpolaczyk: I was wrong, it depends on how you define the proxy:

`Sudo(Registrar) || Utility` this is fine
`Sudo(Registrar || Utility)` this is bad

The current way of always allowing utility should be ok as well

Cargo.lock
```diff
@@ -853,7 +853,7 @@ dependencies = [
 [[package]]
 name = "binary-merkle-tree"
 version = "4.0.0-dev"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-v1.3.0#c3d18c8c2f4c5462b752b98eb2908dc32b5ec985"
```
tmpolaczyk, line 856: Can you merge master? It should remove the changes in Cargo.lock

runtime/dancebox/tests/integration_test.rs
```diff
@@ -2307,6 +2307,80 @@ fn test_proxy_non_transfer() {
...
+    let proxy_types = &[
+        ProxyType::Any,
+        ProxyType::NonTransfer,
+        ProxyType::Governance,
+        ProxyType::Staking,
+        ProxyType::CancelProxy,
+        ProxyType::Balances,
+        ProxyType::Registrar,
+        ProxyType::SudoRegistrar,
+    ];
```
tmpolaczyk, line 2336: There are some external crates that allow iterating over enums, not sure if we already have one, but as a nice hack we could use the Decode trait (not tested but should work):

```
for x in 0u8.. {
    if let Ok(x) = x.decode::<ProxyType>() {
        proxy_types.push(x);
    } else {
        break;
    }
}
```

test/suites/dev-tanssi/registrar/test_registrar_proxy.ts
```diff
@@ -0,0 +1,210 @@
...
+                // Check that the on chain genesis data is set correctly
+                const onChainGenesisData = await polkadotJs.query.registrar.paraGenesisData(2002);
+                // TODO: fix once we have types
+                expect(emptyGenesisData().toJSON()).to.deep.equal(onChainGenesisData.toJSON());
+
+                // assert we can inject bootnodes with proxy
+                const tx3 = polkadotJs.tx.proxy.proxy(
+                    bob.address,
+                    null,
+                    polkadotJs.tx.registrar.setBootNodes(2002, ["dummy"])
```
tmpolaczyk, line 99: Sorry, in #341 we moved the setBootNodes to a different pallet, pallet_data_preservers. I guess we want that to be included in the SudoRegistrar proxy as well. So this should work:

```suggestion
                    polkadotJs.tx.dataPreservers.setBootNodes(2002, ["dummy"])
```

runtime/dancebox/src/lib.rs
```diff
@@ -947,6 +951,12 @@ impl Default for ProxyType {
 
 impl InstanceFilter<RuntimeCall> for ProxyType {
     fn filter(&self, c: &RuntimeCall) -> bool {
+        // Since proxy filters are respected in all dispatches of the Utility
+        // pallet, it should never need to be filtered by any proxy.
+        if let RuntimeCall::Utility(..) = c {
+            return true;
+        }
```
tmpolaczyk, line 958: Good idea, can you do the same change to the templates as well?

test/suites/dev-tanssi/registrar/test_registrar_proxy.ts
girazoki: yes exactly

girazoki:
Please put a description in the PR and let's merge master back in

runtime/dancebox/src/lib.rs
```diff
@@ -955,25 +964,31 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
             ProxyType::Balances => {
-                matches!(c, RuntimeCall::Balances(..) | RuntimeCall::Utility(..))
+                matches!(c, RuntimeCall::Balances(..))
+            }
+            ProxyType::Registrar => {
+                matches!(c, RuntimeCall::Registrar(..))
             }
+            ProxyType::SudoRegistrar => match c {
+                RuntimeCall::Sudo(pallet_sudo::Call::sudo { call: ref x }) => {
+                    matches!(x.as_ref(), &RuntimeCall::Registrar(..))
```
tmpolaczyk, line 988:
```suggestion
                    matches!(x.as_ref(), &RuntimeCall::Registrar(..) | &RuntimeCall::DataPreservers(..))
```

runtime/dancebox/src/lib.rs
```diff
@@ -955,25 +964,34 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
             ProxyType::CancelProxy => matches!(
                 c,
                 RuntimeCall::Proxy(pallet_proxy::Call::reject_announcement { .. })
             ),
             ProxyType::Balances => {
-                matches!(c, RuntimeCall::Balances(..) | RuntimeCall::Utility(..))
+                matches!(c, RuntimeCall::Balances(..))
+            }
+            ProxyType::Registrar => {
+                matches!(c, RuntimeCall::Registrar(..))
```
tmpolaczyk, line 984: Here also

```suggestion
                matches!(c, RuntimeCall::Registrar(..) | RuntimeCall::DataPreservers(..))
```

runtime/dancebox/tests/integration_test.rs
```diff
@@ -2307,6 +2307,80 @@ fn test_proxy_non_transfer() {
...
+
+                let free_balance = Balances::free_balance(AccountId::from(BOB));
+
+                assert_ok!(Proxy::proxy(
+                    origin_of(BOB.into()),
+                    AccountId::from(ALICE).into(),
+                    None,
+                    Box::new(
+                        pallet_utility::Call::batch {
+                            calls: vec![pallet_balances::Call::force_set_balance {
```
tmpolaczyk, line 2369: force_set_balance will always fail because it needs root origin, maybe this test is missing a call to sudo first?

girazoki (approved):

runtime/dancebox/tests/integration_test.rs
girazoki: yes the test needs to be batch->sudo.>force_set_balacne

tmpolaczyk: But `ProxyType::Any` should allow this, right? Then why doesn't the test fail?

girazoki: hmm that's a good question

girazoki: its probably becaues we are still not going through the sudo pallet right?

PR #360, opened 2023-12-08
Assets xcm

girazoki:
Adds pallet-assets, foreign-asset-creator and asset-rate as necessary pallets to mint derivative tokens coming from other chains, with these respective goals:

- pallet-assets: handle representation and assets of derivative tokens
- foreign-asset-creator: maintain a mapping between MultiLocation and assetId, which should make it easier to handle future ML versions.
- asset-rate: maintains the rate at which we can charge weight execution fees for certain tokens

TODO:
- update benchmakrs

For future PRs:

- Incorporate these into templates


Agusrodri (approved):
I didn't see anything that could block the PR. Although I think we should keep an eye on the runtime code that could be moved to polkadot-sdk, like the one related to `NativeAssetReserve` for instance. Also a follow up task would be to add some TS tests (the rust ones look great). Other than those comments LGTM!

girazoki:
thank Agustin! Yeah I will add a few more tests in TS for sure before I merge this PR

test/suites/dev-tanssi/xcm-token-reception/test_dmp_token_reception _appendix.ts
```diff
@@ -0,0 +1,107 @@
+import { beforeAll, describeSuite, expect } from "@moonwall/cli";
```
tmpolaczyk, line 1: Some of these files have a space in its filename, can you rename them?

girazoki: yes!

fgamundi (approved):

girazoki:
We need to update the benchmark for buy_execution but it is dependent on `https://github.com/paritytech/polkadot-sdk/pull/2962`. So for now I think we can postpone this for a future PR

PR #361, opened 2023-12-11
Fix container chain sync mode being ignored

tmpolaczyk:
We were not setting the sync mode properly for container chains, which resulted in them always using full sync.

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -276,12 +276,13 @@ impl ContainerChainSpawner {
                 "Container chain sync mode: {:?}",
                 container_chain_cli.base.base.network_params.sync
             );
-            let container_chain_cli_config = sc_cli::SubstrateCli::create_configuration(
+            let mut container_chain_cli_config = sc_cli::SubstrateCli::create_configuration(
                 &container_chain_cli,
                 &container_chain_cli,
                 tokio_handle.clone(),
             )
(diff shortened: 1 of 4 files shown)


Agusrodri (approved):

fgamundi (approved):

tmpolaczyk:
The test failure is legit, I see this log being repeated multiple times, causing collators to not create blocks:

```
2023-12-12 09:08:30 [Container-2000] Skipping candidate production, because block `0x41b8b64e1b106936d784169996e19be8e9b3d9311655679ae9689f4089881561` is already pruned!
2023-12-12 09:08:30 [Container-2000] 💤 Idle (1 peers), best: #17 (0x41b8…1561), finalized #17 (0x41b8…1561), ⬇ 0.6kiB/s ⬆ 0.5kiB/s
```

This is what happened:
* Container chain syncs using warp sync. Also downloads all the blocks.
* Container chain restarts to start collation process. It restarts using full sync.
* The full sync is immediate because the restart only takes 10 seconds, so no new blocks are produced.
* However block 17 was not finalized before the node restarted, so I guess that this breaks something:

```
2023-12-12 09:06:48 [Container-2000] 💤 Idle (2 peers), best: #17 (0x41b8…1561), finalized #16 (0x3fa2…123d), ⬇ 0.8kiB/s ⬆ 0.9kiB/s
```

tmpolaczyk:
The issue I mentioned was resolved by using paritydb instead of rocksdb

PR #365, opened 2023-12-13
Flashbox runtime

fgamundi:
New Tanssi runtime Flashbox with the following configuration:
- Session length of 5 minutes
- No pooled staking (only invulnerables)
- No XCM
- No rotation collators


girazoki:
Good job @fgamundi! A few things tests that I think make sense to be moved to a folder called common-tanssi:
- author-slot-prediction
- fees
- in issuance-rewards, I would separate the tests from those related to the staking and those related to invlunerables, and move this last one into common
- pallet-configuration
- pallet-registrar
- services-payment

runtime/flashbox/src/lib.rs
```diff
@@ -0,0 +1,1533 @@
...
+
+impl Get<u32> for ConfigurationCollatorRotationSessionPeriod {
+    fn get() -> u32 {
+        Configuration::config().full_rotation_period
+    }
+}
+
+pub struct RemoveInvulnerablesImpl;
+
+// impl RemoveInvulnerables<AccountId> for RemoveInvulnerablesImpl {
```
girazoki, line 525: Remove

runtime/flashbox/src/lib.rs
```diff
@@ -0,0 +1,1533 @@
...
+    // Essentially just Aura, but let's be pedantic.
+    type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
+    type Keys = SessionKeys;
+    type WeightInfo = pallet_session::weights::SubstrateWeight<Runtime>;
+}
+
+/// Read full_rotation_period from pallet_configuration
+pub struct ConfigurationCollatorRotationSessionPeriod;
+
+impl Get<u32> for ConfigurationCollatorRotationSessionPeriod {
```
girazoki, line 517: We can probably return a 0 here directly,

girazoki: Actually you can remove this struct I think its not used

girazoki:
We should add a chopsticks tests to test the upgradeability of it (you can take the existing dancebox one and modify it)

.github/workflows/release.yml
```diff
@@ -340,6 +341,40 @@ jobs:
...
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 20.x
+                  cache: "pnpm"
+            - name: "Generate typescript api"
+              run: |
+                  chmod uog+x target/release/tanssi-node
+                  cd typescript-api
+                  pnpm install
+                  pnpm create-local-interfaces
```
tmpolaczyk, line 371: How does typescript-api support multiple runtimes? As I understand it, it spawns a node and gets the list of pallets and extrinsics. So I guess it will use the dancebox typescript interface for flashbox?

runtime/flashbox/src/lib.rs
```diff
@@ -0,0 +1,1536 @@
...
+                RuntimeOrigin::root(),
+                *para_id,
+                credits,
+            ));
+        }
+    }
+}
+
+// TODO remove Period?
+pub struct NeverRotateCollators(PhantomData<u32>);
```
tmpolaczyk, line 578: Yes
```suggestion
pub struct NeverRotateCollators;
```

node/src/chain_spec/flashbox.rs
```diff
@@ -0,0 +1,319 @@
...
+                get_account_id_from_seed::<sr25519::Public>("Alice"),
+                &container_chains,
+                &mock_container_chains,
+                pallet_configuration::GenesisConfig {
+                    config: HostConfiguration {
+                        max_collators: 100u32,
+                        min_orchestrator_collators: 1u32,
+                        max_orchestrator_collators: 1u32,
+                        collators_per_container: 2u32,
+                        full_rotation_period: prod_or_fast!(24u32, 5u32),
```
tmpolaczyk, line 94: This field is ignored in flashbox, not sure if there is an easy way to remove it here but keep it in dancebox. Anyway 0 means no rotation so it's a better value.
```suggestion
                        full_rotation_period: 0,
```

.github/workflows/release.yml
fgamundi: Yeah, as it is now it's using dancebox's TS interface. We could [modify the generate script](https://github.com/moondance-labs/tanssi/blob/master/typescript-api/scripts/runtime-upgrade.ts#L35) to accept an arg so it know what interface to generate, but I think that using the dancebox one should work for what we need. I might be wrong though

tmpolaczyk: So moonbeam has a separate typescript-api for each runtime, but I don't know how they import that in tests: https://github.com/moonbeam-foundation/moonbeam/tree/master/typescript-api/src

Adding flashbox to the typescript-api is a one line change here:
https://github.com/moondance-labs/tanssi/blob/c43bce59e2258fe244ccb55002156f62e047f2f6/typescript-api/scripts/runtime-upgrade.ts#L9

But I don't know how to ensure that it is imported properly in tests...

girazoki: yes I will make sure I generate a separate one for flashbox, will do in a separate PR

girazoki:
Any idea why the flashbox tests fail? it does not look like a timeout issue

tmpolaczyk:
> Any idea why the flashbox tests fail? it does not look like a timeout issue

```
RpcError: 1014: Priority is too low: (9262886844 vs 9262886844): The transaction has too low priority to replace another transaction already in the pool.
```

There are 2 transactions with the same nonce in the mempool. I bet it's because of the empty blocks on session change, try to add an empty `createBlock()` before this error.

.github/workflows/release.yml
fgamundi: I was also looking into it and arrived at the same question. I think there's no easy way of choosing what interface to import on tests.
I guess it depends on what the future of flashbox looks like. Will it at some point differ enough from dancebox that it require its own interface? Or is the plan to keep it as close as posible to dancebox (including the same pallet indices)?
(cc @girazoki)

tmpolaczyk: I agree with @girazoki that we can handle this in a separate pull request. The way we're doing things now will be fine until we decide to add a pallet to flashbox but not to dancebox, and that might not even happen.

fgamundi: I've gone and added the flashbox interface. Still not sure how to use it on the test environment

girazoki: since the tests are common, let's use the dancebox one

test/moonwall.config.json
```diff
@@ -346,6 +346,54 @@
...
+            "name": "chopsticks_flashbox_upgrade",
+            "testFileDir": ["suites/rt-upgrade-chopsticks"],
+            "foundation": {
+                "type": "chopsticks",
+                "rtUpgradePath": "../target/release/wbuild/flashbox-runtime/flashbox_runtime.compact.compressed.wasm",
+                "launchSpec": [
+                    {
+                        "name": "db",
+                        "type": "parachain",
+                        "configPath": "./configs/stagenet-dancebox.yml"
```
girazoki, line 360: @fgamundi you need to create a new yml for flashbox, identical to the stagenet one, but with this url:

wss://fraa-flashbox-rpc.a.stagenet.tanssi.network

runtime/flashbox/src/lib.rs
```diff
@@ -0,0 +1,1535 @@
...
+}
+
+impl pallet_collator_assignment::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type HostConfiguration = Configuration;
+    type ContainerChains = Registrar;
+    type SessionIndex = u32;
+    type SelfParaId = ParachainInfo;
+    type ShouldRotateAllCollators = NeverRotateCollators;
+    type GetRandomnessForNextBlock = ();
```
tmpolaczyk, line 592: Since flashbox doesn't use randomness, container chains with a lower para id will be given priority when assigning collators.

girazoki: yes, that is fine, it will only happen if we dont have sufficient collators to serve demand

tmpolaczyk (approved):

girazoki (approved):
let's make sure the tests pass and we did not see something weird in previous commits

tmpolaczyk:
```
pnpm moonwall test dev_flashbox
```

This suite always fails locally for me, but with different errors than in CI. It's probably related to the javascript environment, but I already removed node_modules so I don't know what more to do. I hope the next moonwall version fixes this.

I did 4 tries in CI and only the first one failed, so I guess the tests work...

PR #371, opened 2023-12-21
Missing dev_flashbox flags

fgamundi:
Fixes the unexpected behavior when running Moonwall test suite `dev_flashbox`, addressed by `--reserved-only`

### DIFF
--- a/test/moonwall.config.json
+++ b/test/moonwall.config.json
@@ -387,7 +387,11 @@
                             "--alice",
                             "--force-authoring",
                             "--rpc-cors=all", 
-                            "--tmp"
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
For the record, the issue was that moonwall spawns multiple nodes at once, and without the `--reserved-only` flag those nodes connected to each other. `--dev` already implies this flag, that's why the other tests work. Thank you for the detective work @fgamundi! I already tried this locally and it works. (I got a timeout error once but that's a known issue)

tmpolaczyk (approved):

PR #373, opened 2023-12-22
Packages Update

timbrinded:
Updating Moonwall and other dependencies


girazoki (approved):

tmpolaczyk (approved):
I saw a timeout error in dev_flashbox in CI but I cannot reproduce it locally, all suites seem to work better than before. Good job!

timbrinded:
> I saw a timeout error in dev_flashbox in CI but I cannot reproduce it locally, all suites seem to work better than before. Good job!

I'll  keep an eye out for this

PR #374, opened 2023-12-22
Moonwall minor fixes

timbrinded:
Couple of small fixes after moonwall update:

- Default test timeout increase 1 -> 2 min
- Fixed version of moonwall `4.5.0` -> `4.5.1`
- Renamed smoke test providers to "para" (to get RT stats added to env vars)


test/pnpm-lock.yaml
```diff
@@ -1,6487 +0,0 @@
-lockfileVersion: '6.0'
```
tmpolaczyk, line 1: It looks like CI needs this file, not sure why

tmpolaczyk: Or maybe not

tmpolaczyk (approved):

girazoki (approved):

PR #376, opened 2023-12-26
Prepare runtimes and dev services for future async backing support

Agusrodri:
Adds the basic runtimes changes for async backing support in Dancebox, Flashbox and templates. It also adapts the corresponding dev services for the nodes to inject the required `para_head` and `relay_slot` key values.

TODO

- [x] Use proper branches
- [x] Fix templates dev services


Agusrodri:
Both simple and frontier template nodes are not working properly with the new changes, they are panicking with the following error:

```bash
panicked at 'Invalid orchestrator chain state proof: RootMismatch', /root/.cargo/git/checkouts/dancekit-b6c418692b12ca11/74eb41d/container-chain-pallets/authorities-noting/src/lib.rs:190:26
```
It's weird because dancebox and flashbox nodes run without issues. I'm still investigating this.

girazoki:
Pushed the commit that fixes the dev services. We might find a way to make it cleaner in the future, but for now I think this is ok

girazoki:
Questions to be answered:

- do we need a migration from highestSlotSeen to the new pallet async-backing pallet?
- Do we need any other migrations?

girazoki:
I am able to reproduce container-chain blocks not being produce in 2000 in zombienet rotation tests locally. I am going to investigate a bit..

girazoki:
the only thing left is to point to the appropriate branches @Agusrodri . I think after that we can merge

fgamundi (approved):
LGTM

Agusrodri:
Thanks @girazoki! I updated the branches and also my comment above explaining why using the `additional_keys` field was not working locally.

girazoki (approved):

PR #382, opened 2024-01-05
fix relay inface

girazoki:
Fixes the way the relay interface if built, which was not properly doing state-pruning


girazoki:
Closing as it is irrelevant

PR #384, opened 2024-01-08
Change default para id for templates

Agusrodri:
Prior to this change, the local execution of templates nodes was failing as there was a mismatch with the default `para-id` for the `orchestrator-chain` (1000) and the default `para-id` for `container-chains` (also 1000 and now changed to 2000 as we have in moonwall).


girazoki:
@Agusrodri remove the para-id in the moonwall config for the templates

tmpolaczyk (approved):
To reproduce:

```
./target/release/container-chain-template-frontier-node --dev --sealing 1000
```

Does not produce blocks and shows errors in master, but it works in this pr

girazoki (approved):

PR #385, opened 2024-01-11
Rewrite collator assignment to support parathreads

tmpolaczyk:
The existing algorithm was getting too complex to support parathreads. This new algorithm will first calculate how many collators we need in total, removing some container chains if we don't have enough collators, and then assign collators giving priority to those that were already assigned to the same chains.

In most cases this will result in the same assignment, expect that now all the orchestrator collators will be added all at the same time instead of first min and then max (see test `assign_collators_set_zero_per_container`).

Additionally, it allows each container chain to have a different number of collators, and even a min/max similar to the orchestrator chain. This is not used yet but it will be needed to support parathreads, which will have a different number of collators than container chains.

This fixes some bugs related to the case where there are not enough collators, see tests `assign_collators_remove_from_orchestator_when_all_assigned` and `assign_collators_invulnerables_priority_orchestrator_reassigned`.

This PR will not add parathread support, that will come in future PRs.


tmpolaczyk:
There are some unwraps/panics that I would like to remove, but aside from that this should be ready for review

pallets/collator-assignment/src/assignment.rs
```diff
@@ -0,0 +1,404 @@
...
+        available_collators = available_collators.saturating_sub(min_orchestrator_collators);
+
+        let mut container_chains_with_collators = vec![container_chains[0]];
+        // Skipping orchestrator chain because it was handled above
+        for cc in container_chains.iter().skip(1) {
+            if available_collators >= cc.min_collators {
+                available_collators -= cc.min_collators;
+                container_chains_with_collators.push(*cc);
+            } else {
+                // Do not break here because we want to push all the remaining para_ids
```
girazoki, line 143: here you mean parathreads?

pallets/collator-assignment/src/assignment.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// * `old_assigned` must not have duplicate collators.
+    ///
+    /// # Returns
+    ///
+    /// The number of invulnerables assigned to the orchestrator chain, capped to `min_collators`.
+    ///
+    /// # Panics
+    ///
+    /// * If `container_chains` is empty, or if the first element of `container_chains` does not have `para_id == SelfParaId::get()`.
+    pub fn prioritize_invulnerables(
```
girazoki, line 269: I love this refactor BTW. Much cleaner than having it in the pallet

test/scripts/build-spec-warp-sync.sh
```diff
@@ -8,4 +8,4 @@ cd $(dirname $0)/..
 
 mkdir -p specs
 ../target/release/container-chain-template-simple-node build-spec --disable-default-bootnode --add-bootnode "/ip4/127.0.0.1/tcp/33049/ws/p2p/12D3KooWHVMhQDHBpj9vQmssgyfspYecgV6e3hH1dQVDUkUbCYC9" --parachain-id 2000 --raw > specs/warp-sync-template-container-2000.json
-../target/release/tanssi-node build-spec --chain dancebox-local --parachain-id 1000 --add-container-chain specs/warp-sync-template-container-2000.json --invulnerable "Collator1000-01" --invulnerable "Collator1000-02" --invulnerable "Collator2000-01" --invulnerable "Collator2000-02" --invulnerable "Collator1000-03" > specs/warp-sync-tanssi-1000.json
+../target/release/tanssi-node build-spec --chain dancebox-local --parachain-id 1000 --add-container-chain specs/warp-sync-template-container-2000.json --invulnerable "Collator1000-01" --invulnerable "Collator1000-02" --invulnerable "Collator1000-03" --invulnerable "Collator2000-01" --invulnerable "Collator2000-02" > specs/warp-sync-tanssi-1000.json
```
girazoki, line 11: what is the reason for this?

test/scripts/build-spec.sh
```diff
@@ -10,4 +10,4 @@ mkdir -p specs
...
 ../target/release/container-chain-template-simple-node build-spec --disable-default-bootnode --parachain-id 2002 --raw > specs/template-container-2002.json
-../target/release/tanssi-node build-spec --chain dancebox-local --parachain-id 1000 --add-container-chain specs/template-container-2000.json --add-container-chain specs/template-container-2001.json --invulnerable "Collator1000-01" --invulnerable "Collator1000-02" --invulnerable "Collator2000-01" --invulnerable "Collator2000-02" --invulnerable "Collator2001-01" --invulnerable "Collator2001-02" --invulnerable "Collator2002-01" --invulnerable "Collator2002-02" > specs/tanssi-1000.json
+../target/release/tanssi-node build-spec --chain dancebox-local --parachain-id 1000 --add-container-chain specs/template-container-2000.json --add-container-chain specs/template-container-2001.json --invulnerable "Collator1000-01" --invulnerable "Collator1000-02" --invulnerable "Collator2002-01" --invulnerable "Collator2002-02" --invulnerable "Collator2000-01" --invulnerable "Collator2000-02" --invulnerable "Collator2001-01" --invulnerable "Collator2001-02" > specs/tanssi-1000.json
```
girazoki, line 13: And for these, it only changes the order no?

test/scripts/build-spec-warp-sync.sh
tmpolaczyk: Intentional, it's the order change mentioned in the description (see test `assign_collators_set_zero_per_container` for an example)

test/scripts/build-spec.sh
tmpolaczyk: Yes, to assign collators to the same chains as before the refactor

pallets/collator-assignment/src/lib.rs
```diff
@@ -177,10 +182,43 @@ pub mod pallet {
...
                 container_chain_ids.shuffle(&mut rng);
+                parathreads.shuffle(&mut rng);
             }
 
             // We read current assigned collators
             let old_assigned = Self::read_assigned_collators();
+            // Initialize list of container chains as `[orchestrator, container1, container2, parathread1, parathread2]`.
+            // The order means priority: the first chain in the list will be the first one to get assigned collators.
+            // Chains will not be assigned less than `min_collators`, except the orchestrator chain.
+            // First all chains will be assigned `min_collators`, and then the first one will be assigned up to `max`,
```
girazoki, line 193:
```suggestion
            // First all chains will be assigned `min_collators`, and then the orchestrator will be assigned up to `max`,
```

pallets/collator-assignment/src/lib.rs
```diff
@@ -177,10 +182,43 @@ pub mod pallet {
...
+            // then the second one, and so on.
```
girazoki, line 194:
```suggestion
```

pallets/collator-assignment/src/lib.rs
```diff
@@ -177,10 +182,43 @@ pub mod pallet {
...
+                min_collators: T::HostConfiguration::min_collators_for_orchestrator(
+                    target_session_index,
+                ),
+                max_collators: T::HostConfiguration::max_collators_for_orchestrator(
+                    target_session_index,
+                ),
+            }];
+            let collators_per_container =
+                T::HostConfiguration::collators_per_container(target_session_index);
+            for para_id in &container_chain_ids {
```
girazoki, line 206: Can you put a TODO here indicating that this is provisional and that the order should be done based on tipping in the future?

pallets/collator-assignment/src/lib.rs
```diff
@@ -177,10 +182,43 @@ pub mod pallet {
...
             }
 
             // We read current assigned collators
             let old_assigned = Self::read_assigned_collators();
+            // Initialize list of container chains as `[orchestrator, container1, container2, parathread1, parathread2]`.
+            // The order means priority: the first chain in the list will be the first one to get assigned collators.
+            // Chains will not be assigned less than `min_collators`, except the orchestrator chain.
+            // First all chains will be assigned `min_collators`, and then the first one will be assigned up to `max`,
+            // then the second one, and so on.
+            let mut container_chains = vec![ContainerChain {
```
girazoki, line 195: I am not entirely sure I like calling this a containerChain when it really isnt :(. I would rather call it Chain, but I also dont like it too much, maybe we can have a dual enum here that for now have the same parameters?

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Sure! I imagine that the priority will be done instead of shuffling here, so the rest of the algorithm should not change when we add tipping.

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Renamed to `ChainNumCollators` and added docs saying that this could be any kind of chain

pallets/collator-assignment/src/assignment.rs
tmpolaczyk: Yes, clarified the comment

girazoki (approved):

girazoki:
LGTM!

fgamundi (approved):
LGTM

PR #389, opened 2024-01-17
Refactor of benchmarks to run all pallets for all runtimes

girazoki:
Fixes issues when benchmarking author inherent and timestamp from both container-chain and tanssi side, and adds these into both dancebox and simple-container-chain runtimes.

I also took the opportunity to add all benhmarkable pallets to dancebox and container-chain-simple runtimes.


tools/benchmarking.sh
```diff
@@ -1,13 +1,21 @@
...
 # in release mode.
 
 set -e
 
-BINARY="./target/production/tanssi-node"
+# By default we use the tanssi-node release binary
+# However we can use any binary by running the benchmark tool with
+# BINARY=./target/release/container-chain-template-simple-node ./tools/benchmarking.sh
+if [[ -z "${BINARY}" ]]; then
+    BINARY="./target/release/tanssi-node"
```
tmpolaczyk, line 14: The default was production instead of release, why this change?

girazoki: I figured it's easier for us to use in this case to make sure benchmarks run correctly. We can set production in the benchmarking server only

tools/benchmarking.sh
```diff
@@ -16,6 +16,13 @@ else
     BINARY="${BINARY}"
 fi
 
+if [[ -z "${CHAIN}" ]]; then
+    mkdir -p tmp
```
tmpolaczyk, line 20:
```suggestion
```

tmpolaczyk (approved):

tools/benchmarking.sh
```diff
@@ -1,13 +1,34 @@
 #!/usr/bin/env bash
 
-# This script can be used for running moonbeam's benchmarks.
+# This script can be used for running tanssi's benchmarks.
 #
-# The moonbeam binary is required to be compiled with --features=runtime-benchmarks
+# The tamsso binary is required to be compiled with --features=runtime-benchmarks
```
fgamundi, line 5:
```suggestion
# The tanssi binary is required to be compiled with --features=runtime-benchmarks
```

fgamundi (approved):

PR #391, opened 2024-01-18
Stream Payment pallet

nanocryk:
[Rendered pallet README](https://github.com/moondance-labs/tanssi/blob/jeremy-stream-payment-pallet/pallets/stream-payment/README.md)


pallets/stream-payment/src/mock.rs
```diff
@@ -0,0 +1,422 @@
...
+    Timestamp,
+    Never,
+}
+
+pub struct TimeProvider;
+impl pallet_stream_payment::TimeProvider<TimeUnit, Balance> for TimeProvider {
+    fn now(unit: &TimeUnit) -> Option<Balance> {
+        match unit {
+            &TimeUnit::BlockNumber => Some(System::block_number().into()),
+            &TimeUnit::Timestamp => todo!(),
```
tmpolaczyk, line 178: This is missing tests. I guess setting the timestamp to `block_number * 12` will be good enough.

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+        QueryKind = OptionQuery,
+    >;
+
+    /// Lookup for all streams with given target.
+    /// To avoid maintaining a growing list of stream ids, they are stored in
+    /// the form of an entry (AccountId, StreamId). If such entry exists then
+    /// this AccountId is a target in StreamId. One can iterate over all storage
+    /// keys starting with the AccountId to find all StreamIds.
+    #[pallet::storage]
+    pub type LookupStreamsWithTarget<T: Config> = StorageDoubleMap<
```
tmpolaczyk, line 169: These double maps are not used by the pallet, it would probably be better to remove them and calculate them offchain. Unless they should be used somewhere?

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+
+            let delta = now
+                .checked_sub(&stream.last_time_updated)
+                .ok_or(Error::<T>::TimeOverflow)?;
+
+            // We compute the amount due to the target according to the rate, which may be
+            // lowered if the stream deposit is lower.
+            let mut payment = delta
+                .checked_mul(&stream.rate_per_time_unit)
+                .ok_or(Error::<T>::CurrencyOverflow)?;
```
tmpolaczyk, line 440: Here we could saturate to Balance::MAX, if someone sets a rate that will overflow I would expect that stream to be depleted immediately.

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+
+            // We compute the new amount of locked funds. If it underflows it
+            // means that there is more to pay that what is left, in which case
+            // we pay all that is left.
+            let (new_locked, drained) = match stream.deposit.checked_sub(&payment) {
+                Some(v) => (v, false),
```
tmpolaczyk, line 446: Edge case: if `v == 0` the stream will be drained but `drained == false`. I guess it's not important.

pallets/stream-payment/src/lib.rs
tmpolaczyk: Also, add some tests setting the rate to `0` and `u128::MAX`.

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+// (at your option) any later version.
+
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
```
tmpolaczyk, line 16: Can you add crate docs here? Copying the PR description should be enough

```suggestion

//! # Stream Payment Pallet
```

pallets/stream-payment/src/lib.rs
nanocryk: They are here for the UI to find the list of streams one is the source/target.

pallets/stream-payment/src/lib.rs
nanocryk: I don't think it is too important. The idea is that the payment was not complete, but with 0 you'll know that future payments will not complete (unless refill).

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+    sp_runtime::traits::{AtLeast32BitUnsigned, CheckedAdd, CheckedMul, CheckedSub, One, Zero},
+    sp_std::{fmt::Debug, marker::PhantomData},
+};
+
+pub use pallet::*;
+
+/// Type able to provide the current time for given unit.
+/// For each unit the returned number should monotonically increase and not
+/// overflow.
+pub trait TimeProvider<Unit, Number> {
```
girazoki, line 48: What about using the Time trait? https://github.com/paritytech/polkadot-sdk/blob/757ae372f7f07d9a7a4f62f646baa5e388ec5531/substrate/frame/support/src/traits/misc.rs#L723

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    /// Freeze reason to use if needed.
+    #[pallet::composite_enum]
+    pub enum FreezeReason {
+        StreamPayment,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
```
girazoki, line 228: Benches are needed

pallets/stream-payment/src/lib.rs
nanocryk: With `Time` trait the user can't choose between differents units.

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+            });
+
+            // Update stream info in storage.
+            Streams::<T>::insert(stream_id, stream);
+
+            Ok(().into())
+        }
+
+        #[pallet::call_index(4)]
+        pub fn change_stream_rate(
```
girazoki, line 359: Maybe it would also make sense to allow changing the time unit too? what do you think?

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,477 @@
...
+                Error::<T>::UnauthorizedOrigin
+            );
+
+            // Noop
+            if new_rate_per_time_unit == stream.rate_per_time_unit {
+                return Ok(().into());
+            }
+
+            // Ensure rate change is fair.
+            if origin == stream.source && new_rate_per_time_unit < stream.rate_per_time_unit {
```
girazoki, line 379: I think I would allow origin and target to both increase or decrease but I would make it such that:
- if your new rate is higher and you are the source, you need confirmation from the target
- if your new rate is smaller and you are the target, you need confirmation from the source

It probably is the same as what you have but at least two parties get involved in such case

girazoki:
I think it looks good overall, just noted down a couple of observations. Also let's try to configure this in Dancebox and Flashbox, and add some integration tests to it

pallets/stream-payment/src/lib.rs
girazoki: I thought the units would be more of Mulltiples of the same unit. I think we can use seconds and I guess the rest can be handled by a dapp.

girazoki: Ok I see that the expectation is to make an enum, looks good!

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,907 @@
...
+        ) -> DispatchResultWithPostInfo {
+            match change {
+                DepositChange::Absolute(amount) => {
+                    if let Some(increase) = amount.checked_sub(&stream.deposit) {
+                        T::Assets::increase_deposit(
+                            &stream.config.asset_id,
+                            &stream.source,
+                            increase,
+                        )?;
+                    } else if let Some(decrease) = stream.deposit.checked_sub(&amount) {
```
tmpolaczyk, line 879: Could you rewrite this in a way that returns an error if both calls to `checked_sub` fail? I know that's impossible but it makes me a bit nervous. For example:

```suggestion
                    } else {
                        let decrease = stream.deposit.checked_sub(&amount).ok_or(Error)?;
```

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,907 @@
...
+            );
+
+            if stream.config == new_config && deposit_change.is_none() {
+                return Ok(().into());
+            }
+
+            // If asset id and time unit are the same, we allow to make the change
+            // immediatly if the origin is at a disadvantage.
+            // We allow this event if there is already a pending request.
+            'immediate: {
```
tmpolaczyk, line 528: You can avoid this weird control flow by refactoring the condition into a function:

```suggestion
            if allow_immediate_config_change(&stream, &new_config, &requester) {
```

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,907 @@
...
+        /// Steam config (time unit, asset id, rate)
+        pub config: StreamConfig<Unit, AssetId, Balance>,
+        /// How much is deposited to fund this stream.
+        pub deposit: Balance,
+        /// Last time the stream was updated in `config.time_unit`.
+        pub last_time_updated: Balance,
+        /// Nonce for requests. This prevents a request to make a first request
+        /// then change it to another request to frontrun the other party
+        /// accepting.
+        pub request_nonce: RequestNonce,
```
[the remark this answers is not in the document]
nanocryk: You suggest removing the nonce parameter from the `accept_requested_change` call, and preventing accepting a request if one have been proposed before in the same block? What about delays and congestions? Approval extrinsic might be delayed by some blocks, which the requester can frontrun by paying high fees and go throught in a block before. I think the nonce is a simple solution to avoid frontrunning, and the dapp can hide that easily.

pallets/stream-payment/src/lib.rs
nanocryk: Well if the requested amount equals the current amount, there is no issue and no deposit change should be made. It may occur if the request is made, then an immediate request top change the deposit is made.

pallets/stream-payment/src/lib.rs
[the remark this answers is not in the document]
tmpolaczyk: I mean using the block number as the nonce. So when the user wants to approve a request, they would need to read the proposed config which would include the block number at which it was proposed. If a new change is proposed in the same block, it will be rejected. And if a new change is proposed in a later block, it will overwrite the old one but the user will not accidentally accept it, as it will have a different block number.

The main benefit is that we don't need to explain the concept of "request nonce", it's just the block number, but the drawback is that it only allows one change per block (per stream).

pallets/stream-payment/src/lib.rs
tmpolaczyk: Yes please

pallets/stream-payment/src/lib.rs
tmpolaczyk: There is no issue here, just that I don't like seeing `if / else if` without a final `else`. Feel free to ignore me

tmpolaczyk:
What happened to the `refill_stream` extrinsic?

pallets/stream-payment/src/lib.rs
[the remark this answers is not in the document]
nanocryk: In that case it may even not be necessary to forbid changes in the same block. As long as the other party see the proposal in an already proposed block and use that block number afterward to accept it should be safe, right?

nanocryk:
> What happened to the `refill_stream` extrinsic?

It got merged into the immediate part of `request_change`, as I needed to add deposit change anyway in case the source makes a request (with a possible deposit change); and the target can accept it (without being able to set a deposit change, as it is not their funds)

pallets/stream-payment/src/lib.rs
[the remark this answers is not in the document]
girazoki: You still need to provide the block number at which the request was made to accept it right? I am not sure it is more intuitive tbh

pallets/collator-assignment/src/assignment.rs
```diff
@@ -20,7 +20,7 @@ use {
         cmp,
         collections::{btree_map::BTreeMap, btree_set::BTreeSet},
         marker::PhantomData,
-        mem, vec,
+        mem,
```
girazoki, line 23: @nanocryk remember this breaks the CI

pallets/stream-payment/src/benchmarking.rs
```diff
@@ -0,0 +1,127 @@
...
+                stream_id: 0u32.into(),
+            }
+            .into(),
+        );
+
+        Ok(())
+    }
+
+    #[benchmark]
+    fn close_stream() -> Result<(), BenchmarkError> {
```
girazoki, line 81: Reminder to add benchmarks for the rest

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,936 @@
...
+    fn decrease_deposit(asset_id: &AssetId, account: &AccountId, amount: Balance)
+        -> DispatchResult;
+
+    /// Return the deposit for given asset and account.
+    fn get_deposit(asset_id: &AssetId, account: &AccountId) -> Balance;
+
+    /// Benchmarks: should return the asset id which has the worst performance when interacting
+    /// with it.
+    #[cfg(feature = "runtime-benchmarks")]
+    fn bench_asset_id() -> AssetId;
```
girazoki, line 148:
```suggestion
    fn bench_worst_case_asset_id() -> AssetId;
```

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,936 @@
...
+        Mandatory { deadline: Time },
+    }
+
+    /// Describe how the deposit should change.
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    #[derive(RuntimeDebug, PartialEq, Eq, Encode, Decode, Copy, Clone, TypeInfo)]
+    pub enum DepositChange<Balance> {
+        Increase(Balance),
+        Decrease(Balance),
+        Absolute(Balance),
```
girazoki, line 294: I think it Absolute means that such balance will be applied as a rate, but it would be good small comments in each enum variant

girazoki:
Yes I think we should bring back the `refill_stream` even if it means we need to specify the assetId (which I guess makes sense)

pallets/stream-payment/src/tests.rs
```diff
@@ -0,0 +1,1213 @@
...
+    }
+
+    #[test]
+    fn source_can_immediately_increase_deposit() {
+        source_can_immediately_change_deposit(DepositChange::Increase(100))
+    }
+
+    #[test]
+    fn source_can_immediately_decrease_deposit() {
+        source_can_immediately_change_deposit(DepositChange::Decrease(100))
```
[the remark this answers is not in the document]
girazoki: I think it's ok, sometimes source might have put too much in the deposit. The target as well can close the stream if the amount is not being paid, so I think it's ok

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+            }
+
+            // If deposit is zero the stream is fully drained and there is nothing to transfer.
+            if stream.deposit.is_zero() {
+                return Ok(0u32.into());
+            }
+
+            let delta = now
+                .checked_sub(&last_time_updated)
+                .ok_or(Error::<T>::TimeMustBeIncreasing)?;
```
tmpolaczyk, line 840: Making a mandatary change with `deadline = 0` can result in this error. Can be fixed by enforcing the deadline to be `>= last_time_updated`, either in this function or when calling `request_change`.

pallets/stream-payment/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+    /// `bench_asset_id` and `bench_asset_id2`.
+    #[cfg(feature = "runtime-benchmarks")]
+    fn bench_asset_id2() -> AssetId;
+
+    /// Benchmarks: should set the balance.
+    #[cfg(feature = "runtime-benchmarks")]
+    fn bench_set_balance(asset_id: &AssetId, account: &AccountId, amount: Balance);
+}
+
+#[pallet(dev_mode)]
```
tmpolaczyk, line 162: Remember to remove dev_mode and set weights

```suggestion
```

pallets/stream-payment/src/tests.rs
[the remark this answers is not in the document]
nanocryk: That's indeed the idea. Since the source can close the stream at any time and recover unspent deposit, I think we can allow to decrease the deposit without closing the stream. One could first deposit a big amount because they want a long-term service, but at some point temporarily need some funds for something else. What matters for the target is that past time is guaranteed to be paid, and by looking at the events they can compute up to which time the stream is funded.

pallets/stream-payment/src/lib.rs
```diff
@@ -14,57 +14,7 @@
...
-//! For UIs the pallet provides the following storages:
-//! - `Streams: StreamId => Stream`: stream data indexed by stream id.
-//! - `LookupStreamsWithSource: AccountId => StreamId => ()`: allows to list allow the streams with
-//! a given source by iterating over all storage keys with the key prefix corresponding to the
-//! account.
-//! - `LookupStreamsWithTarget: AccountId => StreamId => ()`: same but for the target.
-//! Those last 2 storages are solely for UIs to list incoming and outgoing streams. Key prefix is
-//! used to reduce the POV cost that would require a single Vec of StreamId.
-
+#![doc = include_str!("../README.md")]
```
tmpolaczyk, line 17: Nice!

girazoki: didnt now this!

tmpolaczyk (approved):

girazoki (approved):

PR #392, opened 2024-01-23
Girazoki change payment to tank model

girazoki:
CHanges the payment mechanism to send tokens to a so called `parachain_tank` from which later payments will be substracted.

We maintain the `BlockCredits` but we change their notion:

- before: `BlockCredits` meant both free+purchased credits
- now: `BlockCredits` mean just free credits. We will substract from these credits until we are left with None, then we start substracting from the `parachain_tank`

**To note:** this tank follows the existential deposit rules, meaning that if the tank goes below this value, it will simply be killed

In a future PR: Be able to determine a refund address to send those unspent tokens from the tank. As of today these tokens are burnt


runtime/dancebox/src/lib.rs
```diff
@@ -719,10 +718,14 @@ impl RemoveParaIdsWithNoCredits for RemoveParaIdsWithNoCreditsImpl {
...
-            let credits = pallet_services_payment::BlockProductionCredits::<Runtime>::get(para_id)
+            let free_credits = pallet_services_payment::BlockProductionCredits::<Runtime>::get(para_id)
                 .unwrap_or_default();
 
-            credits >= credits_for_2_sessions
+            let (block_production_costs, _) = <Runtime as pallet_services_payment::Config>::ProvideBlockProductionCost::block_cost(para_id);
+            let regular_credits = Balances::free_balance(pallet_services_payment::Pallet::<Runtime>::parachain_tank(para_id.clone()))
+                .saturating_div(block_production_costs);
+
+            (free_credits as u128).saturating_add(regular_credits) >= credits_for_2_sessions.into()
```
tmpolaczyk, line 728: This is to support both the old and the new kind of credits. Do we want to support both? Why not migrate old credits to the parachain tank address?

pallets/services-payment/src/lib.rs
```diff
@@ -344,14 +292,35 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
...
+                Err(e) => log::warn!(
+                    "Failed to withdraw credits for container chain {}: {:?}",
+                    u32::from(para_id),
+                    e
+                ),
+                Ok(imbalance) => {
+                    T::OnChargeForBlock::on_unbalanced(imbalance);
+                }
+            }
+            total_weight.saturating_add(weight);
```
tmpolaczyk, line 312: This branch is also missing the weight of `T::Currency::withdraw` and `T::OnChargeForBlock::on_unbalanced`. Not sure how I would benchmark this.

Also add a comment somewhere that says something like: "First try to burn a free credit, if there are no credits then use the parachain tank account"

girazoki: yes will do, and I am literally going to benchmark on_author_noted entirely

runtime/flashbox/src/lib.rs
```diff
@@ -761,6 +761,8 @@ impl RegistrarHooks for FlashboxRegistrarHooks {
         // Remove bootnodes from pallet_data_preservers
         DataPreservers::para_deregistered(para_id);
 
+        ServicesPayment::on_deregister(para_id);
```
tmpolaczyk, line 764: The call to `ServicesPayment::set_credits` from above can now be moved to inside this function

girazoki: yep true

pallets/services-payment/src/lib.rs
```diff
@@ -342,16 +290,54 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
...
+        let parachain_tank_balance = T::Currency::total_balance(&Self::parachain_tank(para_id));
+        if !parachain_tank_balance.is_zero() {
+            if let Ok(imbalance) = T::Currency::withdraw(
+                &Self::parachain_tank(para_id),
+                parachain_tank_balance,
+                WithdrawReasons::FEE,
+                ExistenceRequirement::AllowDeath,
+            ) {
+                // Burn for now, we might be able to pass something to do with this
+                drop(imbalance);
```
Agusrodri, line 336: Is there a specific reason why this imbalance is not handled using `OnChargeForBlock` type as the one it is inside `on_container_author_noted` hook function? I mean like `T::OnChargeForBlock::on_unbalanced(imbalance)`.

Agusrodri: I guess due to `OnChargeForBlock` is just for fees handling? But I wonder if we could use the same mechanism to deal with the unspent tokens from the tank.

tmpolaczyk: The plan is to sent them to a user-configured account, but that will be implemented in a future PR because it may require a refactor of `pallet_registrar`.

pallets/services-payment/src/lib.rs
```diff
@@ -342,16 +290,54 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
         _block_number: BlockNumber,
         para_id: ParaId,
     ) -> Weight {
-        let total_weight = T::DbWeight::get().reads_writes(1, 1);
+        if Pallet::<T>::burn_credit_for_para(&para_id).is_err() {
```
tmpolaczyk, line 293: We could rename that function to emphasize that it only burns free credits

```suggestion
        if Pallet::<T>::burn_free_credit_for_para(&para_id).is_err() {
```

pallets/services-payment/src/lib.rs
girazoki: yes exactly, I did not want to make this PR bigger than it already is (we also have an issue for the follow up where we refund the unspent tank tokens)

pallets/services-payment/src/lib.rs
```diff
@@ -342,16 +290,54 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
...
         para_id: ParaId,
     ) -> Weight {
-        let total_weight = T::DbWeight::get().reads_writes(1, 1);
+        if Pallet::<T>::burn_free_credit_for_para(&para_id).is_err() {
+            let (amount_to_charge, _weight) = T::ProvideBlockProductionCost::block_cost(&para_id);
+            match T::Currency::withdraw(
+                &Self::parachain_tank(para_id),
+                amount_to_charge,
+                WithdrawReasons::FEE,
+                ExistenceRequirement::AllowDeath,
```
tmpolaczyk, line 299: Shouldn't this be KeepAlive? In the runtime we check that this account has enough credit with keep alive true:

```
// The true refers to keepAlive
            Balances::can_withdraw(&pallet_services_payment::Pallet::<Runtime>::parachain_tank(*para_id), remaining_to_pay).into_result(true).is_ok()
```

girazoki: yeah it is pretty much irrelevant tbh as we should not enter this function if the account does not have enough money. In the runtime it is important that we know the account will pay for the credits and otherwise we dont assign collators, but once this funciton is reached, it is irrelevant if we do one or the other..

But I can put keepAlice to be consistent

Agusrodri (approved):

pallets/services-payment/src/tests.rs
```diff
@@ -232,129 +123,150 @@ fn burn_credit_fails_for_wrong_para() {
...
+            PaymentServices::on_container_author_noted(&1, 1, 1.into());
+
+            assert_eq!(
+                Balances::balance(&crate::Pallet::<Test>::parachain_tank(1.into())),
+                1u128
+            );
+            println!(
+                "address is {:?}",
+                crate::Pallet::<Test>::parachain_tank(2000.into())
+            )
```
tmpolaczyk, line 270:
```suggestion
```

runtime/dancebox/tests/integration_test.rs
```diff
@@ -4635,23 +4628,30 @@ fn test_can_buy_credits_before_registering_para() {
...
-                &ParaId::from(1001),
-            )
-            .unwrap_or_default();
-            assert_eq!(credits, dancebox_runtime::MaxCreditsStored::get());
+            // Now parachain tank should have this amount
+            println!("encoded {:?}", (b"modlpy/serpayment", 2000u32).encode());
+            println!(
+                "parachain tank address {:?}",
+                ServicesPayment::parachain_tank(2000.into())
+            );
```
tmpolaczyk, line 4644:
```suggestion
```

runtime/dancebox/tests/integration_test.rs
```diff
@@ -5035,3 +5046,243 @@ fn test_register_parathread() {
...
+            assert_ok!(DataPreservers::set_boot_nodes(
+                origin_of(ALICE.into()),
+                1001.into(),
+                dummy_boot_nodes()
+            ));
+            assert_ok!(Registrar::mark_valid_for_collating(
+                root_origin(),
+                1001.into()
+            ));
+            // Need to reset credits to 0 because now parachains are given free credits on register
```
tmpolaczyk, line 5259:
```suggestion
            // Set 1 session of free credits and purchase 1 session of credits
```

test/suites/common-tanssi/services-payment/test_services_payment credit_buying.ts
```diff
@@ -0,0 +1,88 @@
+import "@tanssi/api-augment";
```
tmpolaczyk, line 1: Remove this file, it is a duplicate of `test/suites/common-tanssi/services-payment/test_services_pament_credit_buying_free_combined.ts`

tmpolaczyk (approved):

PR #393, opened 2024-01-23
Support parathread registration and collator assignment

tmpolaczyk:
Adds a new extrinsic `register_parathread`, identical to `register` but allowing some parathread-specific params.

Currently the only param is the slot duration, which is not implemented in the client side, but it will allow parathreads to specify how often do they want blocks to be produced.


test/scripts/sudoRegisterPara.ts
```diff
@@ -42,7 +43,11 @@ yargs(hideBin(process.argv))
 
                 const containerChainGenesisData = chainSpecToContainerChainGenesisData(api, rawSpec);
                 const txs = [];
-                const tx1 = api.tx.registrar.register(rawSpec.para_id, containerChainGenesisData);
+                const slotDuration = api.createType("PalletRegistrarSlotDuration", {
+                    min: 1,
+                    max: 1,
+                });
+                const tx1 = api.tx.registrar.registerParathread(rawSpec.para_id, slotDuration, containerChainGenesisData);
```
tmpolaczyk, line 50: This is for testing, remove before merging

pallets/registrar/src/lib.rs
```diff
@@ -902,12 +988,22 @@ pub mod pallet {
...
-        }
+            let mut parachains = vec![];
+            let mut parathreads = vec![];
 
-        fn session_parathreads(_session_index: T::SessionIndex) -> Vec<ParaId> {
-            // FIXME(parathreads)
-            vec![]
+            for para_id in paras {
+                // TODO: sweet O(n) db reads
+                if let Some(parathread_params) = ParathreadParams::<T>::get(&para_id) {
```
tmpolaczyk, line 996: Here we need some way to distinguish parathreads from parachains. I am using the `ParathreadParams` storage map, but that will do N reads. Here are some alternatives, not sure what's the best one because I don't yet know if params will be used on-chain or not. They will probably be updated over time, so that's why I initially choose the StorageMap. Alternatives:

```
* StorageMap<ParaId, Params> (current)
* StorageValue<BTreeMap<ParaId, Params>> (all under one key)
* StorageValue<Vec<ParaId>> and StorageMap<ParaId, Params> (combination of both)
* Refactor currentContainerChains from `Vec<ParaId>` to `Vec<(ParaId, IsParathread)>`
```

primitives/traits/src/lib.rs
```diff
@@ -72,11 +72,42 @@ pub trait GetCurrentContainerChains {
     fn set_current_container_chains(container_chains: &[ParaId]);
 }
 
+/// How often should a parathread collator propose blocks.
+#[derive(Clone, Debug, Encode, Decode, scale_info::TypeInfo, PartialEq, Eq)]
+pub struct SlotDuration {
```
tmpolaczyk, line 77: Rename this, there is already a `SlotDuration` in aura.

pallets/collator-assignment/src/lib.rs
```diff
@@ -166,14 +166,21 @@ pub mod pallet {
...
+            let mut parathreads: Vec<_> = container_chains
+                .parathreads
+                .into_iter()
+                .map(|(para_id, _)| para_id)
+                .collect();
             // Remove the containerChains that do not have enough credits for block production
             T::RemoveParaIdsWithNoCredits::remove_para_ids_with_no_credits(
                 &mut container_chain_ids,
             );
+            T::RemoveParaIdsWithNoCredits::remove_para_ids_with_no_credits(&mut parathreads);
```
girazoki, line 183: We will probably change differently the parathread credit count and the parachian credit count. We should at least add a TODO here for that.

Ideally we should take into account the slot (something like blocksPerSession/propose_every)

pallets/registrar/src/lib.rs
girazoki: Ideally we should not need another storage item for this. It is very likely that every time we read `RegisteredParchains` we will need to know whether they are a `Parachain` or a `Parathread`, so we might just want to store a struct indicating this already in `RegisteredParchains`...

tmpolaczyk: We decided to keep the current StorageMap, but adding a `PendingParams = StorageValue<Vec<(ParaId, Params)>>` so that changes are only reflected in the next session.

We also would like to change `RegisteredParchains` into a `StorageValue<Vec<(ParaId, Option<Params>)>>`, where `Params` includes both `Option<ParathreadParams>` and also other things such as `paused: bool`. But that looks like a big change so it will be done in a future PR.

pallets/configuration/src/lib.rs
```diff
@@ -131,7 +131,7 @@ impl HostConfiguration {
         if self.max_orchestrator_collators < self.min_orchestrator_collators {
             return Err(InconsistentError::MaxCollatorsLowerThanMinCollators);
         }
-        if self.collators_per_parathread != 1 || self.parathreads_per_collator != 1 {
+        if self.parathreads_per_collator != 1 {
```
girazoki, line 134: here we are finally assuming that we always have one collator per parathread right?

pallets/registrar/src/lib.rs
```diff
@@ -581,6 +606,101 @@ pub mod pallet {
...
+
+            Self::deposit_event(Event::ParaIdRegistered { para_id });
+
+            Ok(())
+        }
+
+        /// Change parathread params
+        // TODO: weight
+        #[pallet::call_index(7)]
+        #[pallet::weight(T::WeightInfo::register(0u32, T::MaxLengthParaIds::get(), 0u32))]
```
girazoki, line 686:
```suggestion
        #[pallet::weight(T::WeightInfo::register(0u32, T::MaxLengthParaIds::get(), 0u32))]
```

Obviously, fix hehe

primitives/traits/src/lib.rs
```diff
@@ -72,11 +72,42 @@ pub trait GetCurrentContainerChains {
     fn set_current_container_chains(container_chains: &[ParaId]);
 }
 
+/// How often should a parathread collator propose blocks.
+#[derive(Clone, Debug, Encode, Decode, scale_info::TypeInfo, PartialEq, Eq)]
+pub struct SlotFrequency {
+    /// The parathread will produce at most 1 block every x slots. min=10 means that collators can produce 1 block
+    /// every `x >= 10` slots, but they are not enforced to. However they cannot produce a block after less than 10
+    /// slots, it will be invalid.
+    // TODO: invalid how, slashing collator? Not paying block reward?
```
girazoki, line 81: for now only not paying reward

girazoki (commented):
I think the layout looks good @tmpolaczyk good job! Now we only need to test it well

pallets/configuration/src/lib.rs
tmpolaczyk: No, here we allow more than one collator per parathread (just like with parachains), but we don't allow one collator to be assigned to more than one parathread (at least not yet).

pallets/registrar/src/lib.rs
tmpolaczyk: Oops, forgot

primitives/traits/src/lib.rs
tmpolaczyk: Also this struct is ignored, I will add some comment to reflect that.

girazoki (approved):
I think it looks good! A couple of tests that I have not seen (maybe I missed them?)
- A test where we change the parameters but we dont deregister. The params should get update after 2 sessions right?

- An additional typescript test where we register a parathread (I would rather have redundancy on tests)

Good job!

fgamundi (approved):
LGTM

PR #394, opened 2024-01-24
Avoid error log "ZeroCollators" when running cargo test

tmpolaczyk:
In #385 I added a log for error cases that should never happen, but I didn't consider that they happen in tests. This results in `cargo test` currently printing this error multiple times:

```
ERROR pallet_collator_assignment::pallet: Error in collator assignment, will keep previous assignment. ZeroCollators
```

So this PR changes the default collator assignment in tests to avoid that error, and also includes a helper function to silence logs inside a test to avoid this problem in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7633,6 +7633,8 @@ dependencies = [
  "sp-runtime",
  "sp-std",
  "tp-traits",
+ "tracing",
+ "tracing-subscriber",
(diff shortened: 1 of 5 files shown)


girazoki (approved):

girazoki:
@tmpolaczyk can we merge this¿?

PR #399, opened 2024-01-31
WIP: Adapt client to support asynchronous backing

Agusrodri:


node/src/service.rs
```diff
@@ -340,30 +377,69 @@ async fn start_node_impl(
...
+                announce_block_clone,
             )
-            .await?
-    } else {
-        node_builder.start_full_node(
-            para_id,
+            .expect("Start consensus should succeed");
+        });
+
+        start_consensus_orchestrator(
```
tmpolaczyk, line 424: If collate_on_tanssi will do exactly the same as this, you can call collate_on_tanssi here, this way avoiding all the clones

```suggestion
        let call = collate_on_tanssi.clone().unwrap();
        call().await;
```

girazoki:
all tests pass? wow

node/src/service.rs
```diff
@@ -636,51 +660,56 @@ fn build_manual_seal_import_queue(
     ))
 }
 
-fn build_consensus_container(
+async fn start_consensus_container(
```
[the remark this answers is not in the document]
Agusrodri: Yes but this async is needed if we want to keep calling the `warp` function with `start_consensus_container` inside it. This is because there is a constraint in the warp fn in which the function that we send via parameter must implement `Future`. We could also remove that constraint, but I don't know how that would affect logs, what do you think?

client/consensus/src/collators/basic.rs
```diff
@@ -0,0 +1,268 @@
...
+                para_id: params.para_id,
+                proposer: params.proposer,
+                collator_service: params.collator_service,
+            };
+
+            collator_util::Collator::<Block, P, _, _, _, _, _>::new(params)
+        };
+
+        while let Some(request) = collation_requests.next().await {
+            macro_rules! reject_with_error {
```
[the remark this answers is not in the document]
Agusrodri: Yeah probably it implies duplicating all the trait bounds, but I'll check to see if I can refactor it somehow.

node/src/service.rs
```diff
@@ -321,49 +350,38 @@ async fn start_node_impl(
...
+        let node_client = node_builder.client.clone();
+        let relay_interface = relay_chain_interface.clone();
+        let node_sync_service = node_builder.network.sync_service.clone();
+        let overseer = overseer_handle.clone();
 
-        // TODO: change for async backing
         collate_on_tanssi = Some(move || async move {
-            #[allow(deprecated)]
-            cumulus_client_collator::start_collator(params_generator()).await;
+            let _ = start_consensus_orchestrator(
```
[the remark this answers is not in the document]
Agusrodri: Actually the only line that could return error in that function, is the one that tries to fetch the slot duration, so I will handle that situation inside of `start_consensus_orchestrator` directly, thus no need for it to return a `Result`.

node/src/service.rs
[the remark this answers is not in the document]
tmpolaczyk: I see, let's add a TODO then. All the `collate_on` stuff was async in the past but does not need to be async anymore. But changing that will be quite a big refactor, so we can do it later. The logs issue can be fixed adding the attribute here as well, this way we can safely remove the wrap function in the future:

```suggestion
// TODO: this function does not need to be async
#[sc_tracing::logging::prefix_logs_with(container_log_str(para_id))]
async fn start_consensus_container(
```

client/consensus/src/consensus_orchestrator.rs
```diff
@@ -436,6 +436,7 @@ where
         }
     }
 
+    // TODO: Where to put these info in the refactor?
     fn pre_digest_data(&self, slot: Slot, claim: &Self::Claim) -> Vec<sp_runtime::DigestItem> {
         vec![
```
tmpolaczyk, line 441: You can make this function call the other one:
```suggestion
    // TODO: Where to put these info in the refactor?
    fn pre_digest_data(&self, slot: Slot, claim: &Self::Claim) -> Vec<sp_runtime::DigestItem> {
        crate::collators::pre_digest_data(slot, claim)
```

Agusrodri: Yeah, in that case we will need to modify this function again when we finally deprecate the old params.

Agusrodri: As the old parameters were removed, now there's no need to modify this one.

tmpolaczyk (approved):
Good job!

client/consensus/src/collators/basic.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub slot_duration: SlotDuration,
+    pub relay_chain_slot_duration: Duration,
+    pub proposer: Proposer,
+    pub collator_service: CS,
+    pub authoring_duration: Duration,
+    pub force_authoring: bool,
+    pub collation_request_receiver: Option<Receiver<CollationRequest>>,
+}
+
+/// Run bare Aura consensus as a relay-chain-driven collator.
```
girazoki, line 72:
```suggestion
/// Run tanssi Aura consensus as a relay-chain-driven collator.
```

client/consensus/src/collators/basic.rs
```diff
@@ -0,0 +1,248 @@
...
+                .relay_client
+                .header(RBlockId::hash(*request.relay_parent()))
+                .await
+            {
+                Err(e) => reject_with_error!(e),
+                Ok(None) => continue, // sanity: would be inconsistent to get `None` here
+                Ok(Some(h)) => h,
+            };
+
+            let authorities = match params
```
girazoki, line 178: Add a comment here of the sort: retrieve the authorities that are able to produce block

client/consensus/src/collators/basic.rs
```diff
@@ -0,0 +1,248 @@
...
+
+            let parent_header = try_request!(Block::Header::decode(
+                &mut &validation_data.parent_head.0[..]
+            ));
+
+            let parent_hash = parent_header.hash();
+
+            if !collator
+                .collator_service()
+                .check_block_status(parent_hash, &parent_header)
```
girazoki, line 163: Again a comment saying: it checks whether we can build upon this block

client/consensus/src/consensus_orchestrator.rs
```diff
@@ -21,562 +21,16 @@
...
 use {
-    cumulus_client_consensus_common::{ParachainCandidate, ParachainConsensus},
-    cumulus_primitives_core::{relay_chain::Hash as PHash, PersistedValidationData},
-    parity_scale_codec::{Decode, Encode},
+    sc_consensus_slots::{SimpleSlotWorker, SlotInfo, SlotResult},
+    sp_consensus::Proposer,
+    sp_runtime::traits::Block as BlockT,
 };
 
-use {
```
girazoki, line 29: Let's try to re-work this tests on a separate PR

client/consensus/src/collators.rs
```diff
@@ -0,0 +1,402 @@
...
+    let signature_digest_item = <DigestItem as NimbusCompatibleDigestItem>::nimbus_seal(signature);
+
+    // seal the block.
+    let block_import_params = {
+        let mut block_import_params = BlockImportParams::new(BlockOrigin::Own, pre_header);
+        block_import_params.post_digests.push(signature_digest_item);
+        block_import_params.body = Some(body.clone());
+        block_import_params.state_action =
+            StateAction::ApplyChanges(sc_consensus::StorageChanges::Changes(storage_changes));
+        block_import_params.fork_choice = Some(ForkChoiceStrategy::LongestChain);
```
girazoki, line 388: I am not sure about this one, specially looking at how polkadot-sdk hadnles this: https://github.com/paritytech/polkadot-sdk/blob/53f615de503c8936c2639f51b99019c64eab1018/cumulus/client/consensus/common/src/parachain_consensus.rs#L458

girazoki: Ah never mind, I see that they have longest chain too https://github.com/paritytech/polkadot-sdk/blob/53f615de503c8936c2639f51b99019c64eab1018/cumulus/client/consensus/aura/src/collator.rs#L373

girazoki (approved):
I think it looks good overall!

runtime/dancebox/src/lib.rs
```diff
@@ -199,6 +202,8 @@ pub mod opaque {
     pub type Block = generic::Block<Header, UncheckedExtrinsic>;
     /// Opaque block identifier type.
     pub type BlockId = generic::BlockId<Block>;
+    /// Opaque block hash type.
+    pub type Hash = <BlakeTwo256 as HashT>::Output;
```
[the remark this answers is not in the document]
Agusrodri: I needed this type to use it inside `service.rs`, and I added it directly on dancebox. I can add it to flashbox as well but I won't use it from there, and at the end it would be the same type tho.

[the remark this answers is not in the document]
girazoki: if you are not going to use it that is fine, dont add it

PR #401, opened 2024-02-05
Add collator assignment payment

girazoki:
Adds a second payment to the `parachain_tank` that will be applied every time a chain gets collators assigned. This is important to avoid cases where a container-chain creates a runtime for which its impossible to create blocks and gets collators blocked for free.

The new collator assignment payment will be fetched from the same `parachain_tank`, however, I also added `CollatorAssignmentCredits` which are free credits that can be consumed, similar to how the block production credits work.

Added a new extrinsic: `set_collator_assignment_credits` callable by root to set free credits for collator-assignment

In the config, there are three new elements 
- `ProvideCollatorAssignmentCost`: which provides the cost for collator_assignment
- `MaxCollatorAssignmentCreditsStored`: which provides the maximum number credits that can be stored for collator assignment
- `OnChargeForCollatorAssignment`: which decides with the tokens substracted from the tank for collator assignment.

We also dont assign collators if container-chains dont have credits/money in the tank for at least 2 session assignments.


pallets/registrar/src/lib.rs
```diff
@@ -995,6 +997,12 @@ pub mod pallet {
                 }
             }
 
+            // Call hook for newly assigned registered chains
+            if let Some(ref new_paras) = new_paras {
+                new_paras.iter().for_each(|para_id| {
+                    T::CollatorAssignmentHook::on_collators_assigned(*para_id);
```
tmpolaczyk, line 1003: This is the list of para ids that want collators to be assigned, before filtering out those that do not have credits. So it is possible for a container chain to get charged and get no collators. A better place for this hook would be in pallet_collator_assignment, after we know which container chains will be assigned collators.

girazoki: perfect thanks @tmpolaczyk

pallets/collator-assignment/src/lib.rs
```diff
@@ -285,6 +287,30 @@ pub mod pallet {
                 }
             };
 
+            // TODO: this probably is asking for a refactor
+            // only apply the onCollatorAssignedHook if sufficient collators
+            for para_id in &container_chain_ids {
```
girazoki, line 292: Let's try to do this in a cleaner manner

node/src/chain_spec/dancebox.rs
```diff
@@ -214,7 +214,7 @@ fn testnet_genesis(
     // Assign 1000 block credits to all container chains registered in genesis
     let para_id_credits: Vec<_> = para_ids
         .iter()
-        .map(|(para_id, _genesis_data, _boot_nodes)| (*para_id, 1000))
+        .map(|(para_id, _genesis_data, _boot_nodes)| (*para_id, 1000, 100))
```
Agusrodri, line 217: Might worth adding a comment here to not only explain the block production credits but also the collator assignment ones.

test/suites/common-tanssi/services-payment/test_services_pament_collator_credit_buying_free_combined.ts
```diff
@@ -0,0 +1,89 @@
+import "@tanssi/api-augment";
```
Agusrodri, line 1: Small typo in the name of this file.

runtime/dancebox/src/lib.rs
```diff
@@ -784,21 +798,35 @@ impl ProvideBlockProductionCost<Runtime> for BlockProductionCost<Runtime> {
...
+    pub const MaxBlockProductionCreditsStored: BlockNumber = 60 * DAYS;
+    // 60 days worth of blocks
+    pub const MaxCollatorAssignmentCreditsStored: u32 = MaxBlockProductionCreditsStored::get()/Period::get();
 }
 
 impl pallet_services_payment::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
     /// Handler for fees
     type OnChargeForBlock = ();
+    type OnChargeForCollatorAssignment = ();
```
Agusrodri, line 819: Have you already planned for a future implementation of this handler? Is it going to be something similar to what we want to do with unspent tokens in `para_deregistered` of `ServicesPayment`? I mean sending the unspent tokens to a user-configured account (I might be missing some context and this could be a totally different situation by the way, so correct me if I'm wrong).

Agusrodri (commented):
Some small comments so far. I think the overall configuration and tests (both rust and TS) look really good!

runtime/dancebox/src/lib.rs
girazoki: no, I think the idea would be:

- unspent tokens go to a container-chain manager dictated refund address.
- collator-assignment spent tokens might involve giving some part to collators, and some other to get burnt
- block production spent tokens might only involve burning

At least the hooks let us implement this whatever way we want in the future

pallets/services-payment/src/lib.rs
```diff
@@ -71,13 +71,22 @@ pub mod pallet {
...
+        /// Provider of a block cost which can adjust from block to block
+        type ProvideCollatorAssignmentCost: ProvideCollatorAssignmentCost<Self>;
+
+        /// The maximum number of block production credits that can be accumulated
+        type MaxBlockProductionCreditsStored: Get<BlockNumberFor<Self>>;
 
-        /// The maximum number of credits that can be accumulated
-        type MaxCreditsStored: Get<BlockNumberFor<Self>>;
+        /// The maximum number of collator assigment production credits that can be accumulated
+        type MaxCollatorAssignmentCreditsStored: Get<u32>;
```
tmpolaczyk, line 89: `MaxCreditsStored` was introduced to limit the number of credits that can be bought in advance, but now credits are not bought in advance so we don't need a limit.

Instead this is being used to set the initial number of free credits inside `give_free_credits`, so we could rename this to reflect that. `FreeAssignmentCredits` for example.

pallets/services-payment/src/lib.rs
```diff
@@ -191,11 +207,29 @@ pub mod pallet {
...
+            Self::set_free_collator_assignment_credits(&para_id, free_collator_assignment_credits);
+
+            Ok(().into())
+        }
     }
 
     impl<T: Config> Pallet<T> {
         /// Burn a credit for the given para. Deducts one credit if possible, errors otherwise.
-        pub fn burn_free_credit_for_para(para_id: &ParaId) -> DispatchResultWithPostInfo {
+        pub fn burn_block_production_free_credit_for_para(
```
tmpolaczyk, line 230: I love how on every PR this function name is getting longer and longer :rofl:

primitives/traits/src/lib.rs
```diff
@@ -43,6 +43,23 @@ pub trait AuthorNotingHook<AccountId> {
...
+impl CollatorAssignmentHook for Tuple {
+    fn on_collators_assigned(p: ParaId) -> Weight {
+        let mut weight: Weight = Default::default();
+        for_tuples!( #( weight.saturating_accrue(Tuple::on_collators_assigned(p)); )* );
+        weight
+    }
+}
+
 #[impl_trait_for_tuples::impl_for_tuples(5)]
 impl<AccountId> AuthorNotingHook<AccountId> for Tuple {
```
tmpolaczyk, line 64: This should be above `pub trait CollatorAssignmentHook`

runtime/dancebox/tests/common/mod.rs
```diff
@@ -231,12 +232,13 @@ pub struct ExtBuilder {
...
     // sudo key
     sudo: Option<AccountId>,
-    // list of registered para ids: para_id, genesis_data, boot_nodes, block_credits
+    // list of registered para ids: para_id, genesis_data, boot_nodes, block_credits, session_credits
     para_ids: Vec<(
         u32,
         ContainerChainGenesisData<MaxLengthTokenSymbol>,
         Vec<Vec<u8>>,
         u32,
+        u32,
```
tmpolaczyk, line 241: This should also be a struct, but don't change it in this PR.

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Yes but we can do it later, when parathreads and parachains are charged differently

pallets/services-payment/src/lib.rs
```diff
@@ -313,6 +420,31 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
...
+            let (amount_to_charge, _weight) =
+                T::ProvideCollatorAssignmentCost::collator_assignment_cost(&para_id);
+            match T::Currency::withdraw(
+                &Self::parachain_tank(para_id),
+                amount_to_charge,
+                WithdrawReasons::FEE,
+                ExistenceRequirement::KeepAlive,
+            ) {
+                Err(e) => log::warn!(
+                    "Failed to withdraw credits for container chain {}: {:?}",
```
tmpolaczyk, line 435: This warning should mention the kind of credits

pallets/services-payment/src/lib.rs
girazoki: hopefully we wont make it larger anymore!

runtime/dancebox/tests/common/mod.rs
girazoki: should not be too hard to change tbh

pallets/services-payment/src/lib.rs
```diff
@@ -239,11 +302,46 @@ pub mod pallet {
...
+                para_id: *para_id,
+                credits: free_collator_block_production_credits,
+            });
+        }
     }
 
     #[pallet::genesis_config]
     pub struct GenesisConfig<T: Config> {
-        pub para_id_credits: Vec<(ParaId, BlockNumberFor<T>)>,
+        pub para_id_credits: Vec<(ParaId, BlockNumberFor<T>, u32)>,
```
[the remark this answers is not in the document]
girazoki: I am thinking about making both u32, but anyway we can do it in another PR. It does not make sense to me that a potentially arbitrary unit (free credits) is tied to block number...what do you think?

runtime/dancebox/tests/common/mod.rs
girazoki: Done in  [4e983c0](https://github.com/moondance-labs/tanssi/pull/401/commits/4e983c0798ad377cd020273c666936a4dc888caf)

tmpolaczyk (approved):

Agusrodri (approved):

PR #402, opened 2024-02-05
Implement on demand block production

tmpolaczyk:
Allows registered parathreads to produce blocks every n slots, where n is defined on chain.

This is ready to be tested, but it may need a refactor (`get_authorities_from_orchestrator` will probably need to be renamed)

How to test this:

* `pnpm moonwall test zombie_tanssi_parathreads`

This is a new zombienet test suite that is not being run in CI.

Or, to test manually:

* `pnpm moonwall run zombie_tanssi`

* Manually call Registrar.deregister for para id 2000

* `pnpm run sudo-register-para register --parathread --url ws://127.0.0.1:9948 --account //Alice --chain specs/template-container-2000.json`

* Manually call Registrar.setParathreadParams to adjust slotFrequency. Set min=max=5 to produce 1 block every 5 slots (60 seconds)

* Observe how parachain 2000 produces 1 block every 60 seconds.

### DIFF
--- a/client/consensus/src/collators.rs
+++ b/client/consensus/src/collators.rs
@@ -29,7 +29,7 @@ use parity_scale_codec::{Codec, Encode};
 use polkadot_node_primitives::{Collation, MaybeCompressedPoV};
 use polkadot_primitives::Id as ParaId;
 
-use crate::AuthorityId;
+use crate::{find_pre_digest, AuthorityId, OrchestratorAuraWorkerAuxData};
 use futures::prelude::*;
 use nimbus_primitives::{CompatibleDigestItem as NimbusCompatibleDigestItem, NIMBUS_KEY_ID};
 use sc_consensus::{BlockImport, BlockImportParams, ForkChoiceStrategy, StateAction};
@@ -41,7 +41,7 @@ use sp_inherents::{CreateInherentDataProviders, InherentData, InherentDataProvid
 use sp_keystore::{Keystore, KeystorePtr};
 use sp_runtime::{
     generic::Digest,
-    traits::{Block as BlockT, HashingFor, Header as HeaderT, Member},
+    traits::{Block as BlockT, HashingFor, Header as HeaderT, Member, Zero},
 };
 use sp_state_machine::StorageChanges;
 use sp_timestamp::Timestamp;
@@ -87,7 +87,7 @@ where
     BI: BlockImport<Block> + ParachainBlockImportMarker + Send + Sync + 'static,
     Proposer: ProposerInterface<Block>,
     CS: CollatorServiceInterface<Block>,
-    P: Pair,
+    P: Pair + Send + Sync + 'static,
     P::Public: AppPublic + Member,
     P::Signature: TryFrom<Vec<u8>> + Member + Codec,
 {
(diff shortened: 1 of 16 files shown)


client/consensus/src/collators.rs
```diff
@@ -284,28 +284,66 @@ impl<Pub: Clone> SlotClaim<Pub> {
...
+{
+    if slot.is_zero() {
+        // Always produce on slot 0 (for tests)
+        return false;
+    }
+    if let Some((start, _end)) = aux_data.next_slot_range_inclusive {
+        if let Ok(chain_head_slot) = find_pre_digest::<B, P::Signature>(chain_head) {
+            let slot_diff = slot.saturating_sub(chain_head_slot);
+
+            // TODO: this ignores force authoring
```
girazoki, line 335: what does this sentence mean?

client/consensus/src/lib.rs
```diff
@@ -103,6 +106,35 @@ where
     authorities
 }
 
+/// Return the set of authorities assigned to the paraId where
```
girazoki, line 109: please update the variable names (you are returning something called authorities haha). Also the comment

pallets/registrar/rpc/runtime-api/src/lib.rs
```diff
@@ -36,3 +36,19 @@ sp_api::decl_runtime_apis! {
...
+        Slot: parity_scale_codec::Codec,
+    {
+        /// Inclusive range of slots during which collators can propose the next block.
+        ///
+        /// # Returns
+        ///
+        /// Range `Some((start, end))`, where the condition for the slot to be valid is
+        /// `(slot - parent_slot) >= start && (slot - parent_slot) <= end`.
+        /// `None` if the `para_id` is not a parathread.
+        fn next_slot_range_inclusive(para_id: ParaId) -> Option<(Slot, Slot)>;
```
girazoki, line 52: Shall we remove the end? I guess this is related to the max parameter, which we said we should probably skip?

client/consensus/src/collators.rs
tmpolaczyk: That even if we set force_authoring to true, this collator will not propose blocks on every slot

client/consensus/src/lib.rs
tmpolaczyk: True, I was also thinking about renaming `get_authorities_from_orchestrator` into something more generic, because now we fetch authorities and slot freq

pallets/registrar/rpc/runtime-api/src/lib.rs
tmpolaczyk: Yes, I think I will return only the `start`. But I need to experiment a bit on runtime apis to see how easy it will be to change the method signature in the future.

client/consensus/src/lib.rs
girazoki: yes we should, maybe something like get consensus data?

pallets/registrar/rpc/runtime-api/src/lib.rs
tmpolaczyk: So regarding runtime api versioning, it looks like changing a method signature is not possible. The only way to change it is to create a new method, which will return an error if called using an older runtime.

So I created a method that returns an `Option<start>`, which should cover all the possible user cases. Returning `None` will make clients create blocks immediately, and returning `Some(u32::MAX)` can be used to prevent outdated clients from creating blocks.

girazoki:
How much does it take for the new test to run? Because maybe it makes sense to add it to the CI?

tmpolaczyk:
> How much does it take for the new test to run? Because maybe it makes sense to add it to the CI?

It runs for 3 tanssi sessions, so I guess about 5-6 minutes. But I don't think it's very useful now, this test is going to be helpful when we start adding more parathread features. Also it would be a lot quicker, 2-3 minutes, if we supported parathread registration in genesis.

So how about we enable it the next time we touch parathreads?

girazoki:
I think we can add it now (I like the fact that you wrote it to be tested :)). I think 6 minutes running on the github servers its not too much because they can be parallelized

girazoki (approved):
Other than that looks good!

PR #406, opened 2024-02-09
Refund address to services-payment

girazoki:
Adds the possibility of adding a refund address (being the container-chain manager or root) to where the tokens will go upon deregistration.

If such address is not set, then the tokens are burnt


pallets/services-payment/src/lib.rs
```diff
@@ -332,8 +368,12 @@ impl<T: Config> Pallet<T> {
...
-                // Burn for now, we might be able to pass something to do with this
-                drop(imbalance);
+                if let Some(address) = RefundAddress::<T>::get(para_id) {
+                    T::Currency::resolve_creating(&address, imbalance);
+                } else {
+                    // Burn for now, we might be able to pass something to do with this
+                    drop(imbalance);
+                }
             }
         }
```
tmpolaczyk, line 378: This doesn't clear the refund address storage, add this outside of the if (if we want to clear it):

```
RefundAddress::<T>::remove(para_id);
```

girazoki: good catch, I will probably do a take

pallets/services-payment/src/lib.rs
```diff
@@ -332,8 +368,12 @@ impl<T: Config> Pallet<T> {
                 WithdrawReasons::FEE,
                 ExistenceRequirement::AllowDeath,
             ) {
-                // Burn for now, we might be able to pass something to do with this
-                drop(imbalance);
+                if let Some(address) = RefundAddress::<T>::take(para_id) {
```
tmpolaczyk, line 371: This only removes it if the `parachain_tank_balance` is not zero, if it is zero it will not be removed

girazoki: Added a take instead of get which should remove it

tmpolaczyk: Yes but this take is inside an if condition, which may not always be true

girazoki: yeah you are right, in case of credits being used only this might not hit. I will modify and add a test accordingly

tmpolaczyk (approved):

pallets/services-payment/src/lib.rs
```diff
@@ -191,6 +207,26 @@ pub mod pallet {
...
             Ok(().into())
         }
+
+        /// Call index to set the refund address for non-spent tokens
+        #[pallet::call_index(3)]
+        #[pallet::weight(T::WeightInfo::set_refund_address())]
+        pub fn set_refund_address(
+            origin: OriginFor<T>,
+            para_id: ParaId,
+            refund_address: T::AccountId,
```
nanocryk, line 217: No way to remove the refund address?

nanocryk (approved):

pallets/services-payment/src/lib.rs
girazoki: good catch

girazoki: I will add that

girazoki: So I added the refund_address as an optional parameter, if None is provided we remove it

PR #407, opened 2024-02-09
Re-work client tests

Agusrodri:
In the previous work for https://github.com/moondance-labs/tanssi/pull/399 we removed some tests in the client related to the old consensus model. This PR re-includes and adapts them to the new kind of collator params.


girazoki (approved):

client/consensus/Cargo.toml
```diff
@@ -39,10 +39,13 @@ cumulus-client-consensus-proposer = { workspace = true }
 cumulus-primitives-core = { workspace = true }
 cumulus-primitives-parachain-inherent = { workspace = true }
 cumulus-relay-chain-interface = { workspace = true }
+cumulus-test-relay-sproof-builder = { workspace = true }
 
 # Polkadot
+polkadot-core-primitives = { workspace = true }
 polkadot-node-primitives = { workspace = true }
 polkadot-overseer = { workspace = true }
+polkadot-parachain-primitives = { workspace = true }
```
tmpolaczyk, line 48: New dependencies that are only used in tests should be under `[dev-dependencies]`.

tmpolaczyk (approved):

PR #408, opened 2024-02-12
Remove collate_on closure

tmpolaczyk:
No longer being used because we always restart container chains.

TODO: fix unit tests

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -92,10 +92,6 @@ pub struct ContainerChainSpawnerState {
 }
 
 pub struct ContainerChainState {
-    /// Async callback that enables collation on this container chain
-    // We don't use it since we are always restarting container chains
-    #[allow(unused)]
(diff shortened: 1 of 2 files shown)


node/src/container_chain_spawner.rs
```diff
@@ -291,35 +287,22 @@ impl ContainerChainSpawner {
...
+                start_node_impl_container(
+                    container_chain_cli_config,
+                    orchestrator_client.clone(),
+                    relay_chain_interface.clone(),
+                    orchestrator_chain_interface.clone(),
+                    collator_key.clone(),
+                    sync_keystore.clone(),
+                    container_chain_para_id,
+                    orchestrator_para_id,
+                    validator && start_collation,
```
tmpolaczyk, line 300: Previously when we wanted to sync a container chain we started it with "collator = true", but we didn't call the `start_consensus_container` part. Now we simply start it with "collator = false". The only difference is that we no longer call `start_relay_chain_tasks`, which were not needed before when not running as collator.

girazoki (approved):

PR #411, opened 2024-02-13
Xcm assets into container runtimes

girazoki:
Brings all XCM asset handling to container-chains, and joins tests. In particular pallet-assets, pallet-asset-rate and foreign-asset-creator have been added, and the XCM configuration for them

TODO: some PRs left to be emrged in Polkadot...I can put TODOs on those things that need to get fixed


container-chains/templates/simple/runtime/src/xcm_config.rs
```diff
@@ -243,3 +253,185 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+    type Error = ();
+    fn to_asset_balance(balance: Balance, asset_id: AssetId) -> Result<Balance, Self::Error> {
+        let rate = pallet_asset_rate::ConversionRateToNative::<Runtime>::get(asset_id).ok_or(())?;
+        Ok(sp_runtime::FixedU128::from_u32(1)
+            .checked_div(&rate)
+            .ok_or(())?
+            .saturating_mul_int(balance))
+    }
+}
+
```
girazoki, line 379: This is going to dissappear once I convince parity to accept https://github.com/paritytech/polkadot-sdk/pull/3318

Hopefully this happens fast

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -244,3 +258,228 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+
+        if junctions != Here {
+            Some(MultiLocation::new(0, junctions))
+        } else {
+            None
+        }
+    }
+}
+
+pub struct AssetFeeAsExistentialDepositMultiplier<
```
girazoki, line 443: Should dissappear once https://github.com/paritytech/polkadot-sdk/pull/3303 lands

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -244,3 +258,228 @@ impl cumulus_pallet_dmp_queue::Config for Runtime {
...
+}
+#[cfg(feature = "runtime-benchmarks")]
+impl pallet_asset_rate::AssetKindFactory<AssetId> for ForeignAssetBenchmarkHelper {
+    fn create_asset_kind(id: u32) -> AssetId {
+        id.try_into()
+            .expect("number too large to create benchmarks")
+    }
+}
+
+pub type AssetId = u16;
```
Agusrodri, line 291: Is there a specific reason why we use u16 here instead of u128 as we do in Moonbeam?

girazoki: I think it might offer better protection to evm, so I am going to change it to u128

girazoki: Actually it might provide more protection if we set it to u16, as it is more difficult for an attacker to guess an evm contract that has more FFs fixed

girazoki: For context:
- What worries me the most is that someone deploys a contract in the precompile address. However with a u16 (2 bytes) they need to deploy a contract at (0xFFFF..FF (18 bytes)assetId). That is 2^144 probability already.

Agusrodri: What do you mean by FFs?

Agusrodri (approved):

PR #412, opened 2024-02-14
Always use wasm executor in container collators

tmpolaczyk:
We use a `NativeElseWasmExecutor` for all chains, but collators don't have the container chain's native runtime. So their executor was trying to use the native dancebox runtime as the container chain runtime, which fortunately resulted in a version mismatch and the wasm container chain runtime was always used.

This can be verified by starting zombienet with `export RUST_LOG=info,executor=trace`:

```
2024-02-13 18:33:48.286 TRACE tokio-runtime-worker executor: [Container-2000] Request for native execution failed native=dancebox-500 (dancebox-0.tx1.au1) chain=container-chain-template-500 (container-chain-template-0.tx1.au1)
```

As a bonus we now have two types of clients, `Arc<ContainerChainClient>` and `Arc<ParachainClient>`, and using one client in place of the other results in a compile error.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -60,7 +60,7 @@ use {
     },
     sc_consensus::BlockImport,
     sc_consensus::{BasicQueue, ImportQueue},
-    sc_executor::NativeElseWasmExecutor,
+    sc_executor::{NativeElseWasmExecutor, WasmExecutor},
     sc_network::NetworkBlock,
     sc_network_sync::SyncingService,
     sc_service::{Configuration, SpawnTaskHandle, TFullBackend, TFullClient, TaskManager},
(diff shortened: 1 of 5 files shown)


node/src/service.rs
```diff
@@ -101,19 +101,32 @@ pub struct NodeConfig;
...
     type RuntimeApi = RuntimeApi;
-    type ParachainNativeExecutor = ParachainNativeExecutor;
+    type ParachainExecutor = ParachainExecutor;
 }
 
-type ParachainExecutor = NativeElseWasmExecutor<ParachainNativeExecutor>;
+pub struct ContainerChainNodeConfig;
+impl NodeBuilderConfig for ContainerChainNodeConfig {
+    type Block = Block;
+    // TODO: RuntimeApi here should be the subset of runtime apis available for all containers
```
girazoki, line 110: Good first issue :)

girazoki (approved):

node/src/service.rs
girazoki: @tmpolaczyk let's create an issue for this so that we remember

tmpolaczyk: Created

PR #413, opened 2024-02-14
Polkadot v1.6.0 upgrade

fgamundi:
Upgrades polkadot-sdk dependencies to v1.6.0

Some relevant PRs
https://github.com/paritytech/polkadot-sdk/pull/1256 chain-spec: getting ready for native-runtime-free world
https://github.com/paritytech/polkadot-sdk/pull/1246 DmpQueue & XcmpQueue transition to MessageQueue
https://github.com/paritytech/polkadot-sdk/pull/2048 & https://github.com/paritytech/polkadot-sdk/pull/2651 Identity pallet improvements
https://github.com/paritytech/polkadot-sdk/pull/2803 `cumulus-primitives-parachain-inherent` split
https://github.com/paritytech/polkadot-sdk/pull/2331 `ExportGenesisStateCommand` to `ExportGenesisHeadCommand`
https://github.com/paritytech/polkadot-sdk/pull/2834 proposer: return optional block
https://github.com/paritytech/polkadot-sdk/pull/1343 `RuntimeTask`
https://github.com/paritytech/polkadot-sdk/pull/2767 `PartialComponents`


client/node-common/src/service.rs
```diff
@@ -521,10 +521,14 @@ where
             // Here you can check whether the hardware meets your chains' requirements. Putting a link
             // in there and swapping out the requirements for your own are probably a good idea. The
             // requirements for a para-chain are dictated by its relay-chain.
-            if collator && !SUBSTRATE_REFERENCE_HARDWARE.check_hardware(hwbench) {
-                log::warn!(
-                    "⚠️  The hardware does not meet the minimal requirements for role 'Authority'."
+            match SUBSTRATE_REFERENCE_HARDWARE.check_hardware(&hwbench) {
+                Err(err) if collator => {
```
tmpolaczyk, line 525: Is this intentional? The old code doesn't run `check_hardware` if the node is not a collator. The new code always runs it but only prints the warning if the node is a collator.

runtime/dancebox/tests/common/xcm/constants.rs
```diff
@@ -218,11 +218,12 @@ pub mod frontier_template {
 
     pub fn genesis() -> sp_core::storage::Storage {
         let genesis_config = container_chain_template_frontier_runtime::RuntimeGenesisConfig {
+            system: Default::default(),
             balances: container_chain_template_frontier_runtime::BalancesConfig {
                 balances: pre_funded_accounts()
                     .iter()
                     .cloned()
-                    .map(|k| (k, 1 << 80))
```
tmpolaczyk, line 231: Why this?

fgamundi: Seems to be a side effect of the new chain spec builder working with serde_json, as with anything bigger than a u64 it errors with with "number out of range"

tmpolaczyk: That has an easy fix, I saw it in the chain spec PR in polkadot-sdk, basically you need to add a feature to `serde_json` dependency.

node/src/cli.rs
```diff
@@ -323,38 +324,38 @@ impl ContainerChainCli {
...
+            &[],
             // TODO: what to do with extensions? We are hardcoding the relay_chain and the para_id, any
             // other extensions are being ignored
             extensions,
-        );
+        )
+        .with_name(&name)
+        .with_id(&id)
+        .with_chain_type(chain_type)
+        .with_genesis_config(serde_json::to_value(&raw_genesis_config).unwrap())
```
tmpolaczyk, line 347: This panics because "`RawGenesisConfigDummy` should never be serialized"

runtime/dancebox/tests/common/xcm/constants.rs
```diff
@@ -204,7 +196,11 @@ pub mod westend {
             ..Default::default()
         };
 
-        genesis_config.build_storage().unwrap()
+        ChainSpec::builder(westend_runtime::WASM_BINARY.unwrap(), None)
+            .with_genesis_config(serde_json::to_value(&genesis_config).unwrap())
+            .build()
+            .build_storage()
+            .unwrap()
```
tmpolaczyk, line 203: I believe this change makes CI take far longer for this test than it should because each test creates compiles the wasm once. Not sure if we can improve it however.

client/consensus/src/collators/basic.rs
```diff
@@ -240,11 +240,17 @@ where
                     .await
             );
 
-            let result_sender = Some(collator.collator_service().announce_with_barrier(post_hash));
-            request.complete(Some(CollationResult {
-                collation,
-                result_sender,
-            }));
+            if let Some((collation, _, post_hash)) = maybe_collation {
```
girazoki, line 243: Damn it,  I hate that they keep changing this part of the code

container-chains/templates/frontier/node/src/cli.rs
```diff
@@ -47,7 +47,8 @@ pub enum Subcommand {
     PurgeChain(cumulus_client_cli::PurgeChainCmd),
 
     /// Export the genesis state of the parachain.
-    ExportGenesisState(cumulus_client_cli::ExportGenesisStateCommand),
+    #[command(alias = "export-genesis-state")]
+    ExportGenesisHead(cumulus_client_cli::ExportGenesisHeadCommand),
```
girazoki, line 51: Worth mentioning to the dapp team

girazoki: although with the alias I hope nothing changes

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -163,7 +163,7 @@ pub type Executive = frame_executive::Executive<
     Block,
     frame_system::ChainContext<Runtime>,
     Runtime,
-    pallet_maintenance_mode::ExecutiveHooks<Runtime>,
+    AllPalletsWithSystem,
```
girazoki, line 166: Why this? this seems like a big change

girazoki: This is basically saying that maintenance mode is no-longer in charge of orchestrating hooks. Is this what we want?

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -706,89 +709,12 @@ impl Contains<RuntimeCall> for NormalFilter {
...
-    fn handle_dmp_messages(
-        iter: impl Iterator<Item = (RelayBlockNumber, Vec<u8>)>,
-        _limit: Weight,
-    ) -> Weight {
-        DmpQueue::handle_dmp_messages(iter, Weight::zero())
-    }
-}
-
-/// The hooks we want to run in Maintenance Mode
-pub struct MaintenanceHooks;
```
girazoki, line 738: Ah I see, so we only needed to override this because of the on_idle right? excluding on_idle (which we are changing) all others are `AllPalletsWithSystem`?

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -217,30 +220,61 @@ impl pallet_xcm::Config for Runtime {
...
     type RuntimeEvent = RuntimeEvent;
-    type XcmExecutor = XcmExecutor<XcmConfig>;
     type ChannelInfo = ParachainSystem;
     type VersionWrapper = PolkadotXcm;
-    type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
     type ControllerOrigin = EnsureRoot<AccountId>;
     type ControllerOriginConverter = XcmOriginToTransactDispatchOrigin;
     type WeightInfo = cumulus_pallet_xcmp_queue::weights::SubstrateWeight<Self>;
     type PriceForSiblingDelivery = NoPriceForMessageDelivery<ParaId>;
+    type XcmpQueue = TransformOrigin<MessageQueue, AggregateMessageOrigin, ParaId, ParaIdToSibling>;
```
girazoki, line 234: Is this the migration config?

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -217,30 +220,61 @@ impl pallet_xcm::Config for Runtime {
...
+    type WeightInfo = cumulus_pallet_dmp_queue::weights::SubstrateWeight<Runtime>;
     type RuntimeEvent = RuntimeEvent;
-    type XcmExecutor = XcmExecutor<XcmConfig>;
-    type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+    type DmpSink = frame_support::traits::EnqueueWithOrigin<MessageQueue, RelayOrigin>;
+}
+
+parameter_types! {
+    // TODO verify 35%
+    pub MessageQueueServiceWeight: Weight = Perbill::from_percent(35) * RuntimeBlockWeights::get().max_block;
```
girazoki, line 255: hmmm interesting limit. I think we had 25% before?

runtime/dancebox/src/migrations.rs
```diff
@@ -563,16 +354,40 @@ where
     }
 }
 
+const IDENTITY_MIGRATION_KEY_LIMIT: u64 = u64::MAX;
+
+pub struct IdentityMigration<T>(pub PhantomData<T>);
```
girazoki, line 359: I am not sure we need this, as I think the identity pallet is first introduced in this runtime. You can verify this, and if so, you can remove this migration I think.

Also, is there any other migration we need?

runtime/dancebox/tests/integration_test.rs
```diff
@@ -2410,43 +2409,6 @@ fn check_well_known_keys() {
     );
 }
 
-#[test]
-fn test_invulnerables_migration() {
```
girazoki, line 2414: Are we removing this migrations?

test/suites/common-xcm/test-maintenance/test-maintenance-dmp-queue.ts
```diff
@@ -128,9 +128,9 @@ describeSuite({
                 .as_v3();
 
             // In case of templates, we set a different Config for DmpQueue
-            if (["frontier-template", "container-chain-template"].includes(chain)) {
-                await setDmpConfigStorage(context, polkadotJs, alice);
-            }
+            // if (["frontier-template", "container-chain-template"].includes(chain)) {
```
girazoki, line 131: Remove if it is not needed anymore.

runtime/dancebox/src/migrations.rs
```diff
@@ -334,38 +153,10 @@ where
...
-    fn migrate(&self, _available_weight: Weight) -> Weight {
-        cumulus_pallet_xcmp_queue::migration::Migration::<T>::on_runtime_upgrade()
-    }
-}
-
 pub struct MigrateServicesPaymentAddCredits<T>(pub PhantomData<T>);
 impl<T> Migration for MigrateServicesPaymentAddCredits<T>
 where
-    T: cumulus_pallet_xcmp_queue::Config,
+    T: pallet_balances::Config,
```
girazoki, line 159:
```suggestion
    T: pallet_services_payment::Config,
```

runtime/dancebox/src/migrations.rs
```diff
@@ -536,7 +327,7 @@ pub type RegistrarBootNodesStorageMap<T> = StorageMap<
 pub struct MigrateBootNodes<T>(pub PhantomData<T>);
 impl<T> Migration for MigrateBootNodes<T>
 where
-    T: cumulus_pallet_xcmp_queue::Config,
+    T: pallet_balances::Config,
```
girazoki, line 330:
```suggestion
    T: pallet_registrar::Config,
```

client/node-common/src/service.rs
tmpolaczyk: I noticed that `check_hardware` does not run the benchmarks, it only compares the result. So benchmarks are being run in a function called `gather_hwbench`, which was already being run for all nodes before.

Anyway, I prefer the old version of this piece of code

container-chains/templates/frontier/runtime/src/lib.rs
fgamundi: That's it!

container-chains/templates/frontier/runtime/src/xcm_config.rs
fgamundi: It's both the normal XCMP message handling and migration config. It follows a similar logic to the DMP queue migration to lazily migrate any leftover items to the new `XcmpQueue` https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/pallets/xcmp-queue/src/lib.rs#L246

runtime/dancebox/tests/integration_test.rs
fgamundi: Yes, I removed migrations that were already applied and commented in `get_migrations` https://github.com/moondance-labs/tanssi/blob/runtime-400/runtime/dancebox/src/migrations.rs#L572 because there was a breaking change that affected one of them (don't remember exactly). Instead of trying to fix it I just removed them

container-chains/templates/frontier/runtime/src/xcm_config.rs
fgamundi: I grabbed that number from parity because I wasn't sure what was our previous limit. Where can I find it?

runtime/dancebox/src/migrations.rs
fgamundi: Not that I know of, other than the xcmp and dmp queues

container-chains/templates/frontier/runtime/src/xcm_config.rs
girazoki: https://github.com/moondance-labs/tanssi/blob/b018910b8234464df9da8cc65b642bd6f2fedcdc/runtime/dancebox/src/lib.rs#L479

container-chains/templates/simple/runtime/src/migrations.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn get_migrations() -> Vec<Box<dyn Migration>> {
+        let migrate_polkadot_xcm_v1 =
+            PolkadotXcmMigrationFixVersion::<Runtime, PolkadotXcm>(Default::default());
+        let migrate_xcmp_queue_v2 =
+            XcmpQueueMigrationFixVersion::<Runtime, XcmpQueue>(Default::default());
+        let migrate_xcmp_queue_v3 = XcmpQueueMigrationV3::<Runtime>(Default::default());
+        let migrate_xcmp_queue_v4 = XcmpQueueMigrationV4::<Runtime>(Default::default());
+        vec![
+            Box::new(migrate_polkadot_xcm_v1),
+            Box::new(migrate_xcmp_queue_v2),
```
girazoki, line 53: did we try this with tryruntime?

girazoki: I tried these migrations with chopsticks and they apply well!

girazoki (approved):
LGTM, good job @fgamundi ! But please remember to point to the right branch instead of the one with 'preview'

Agusrodri (approved):

tmpolaczyk (approved):

PR #414, opened 2024-02-14
Refactor collate_on_tanssi closure, make it not async

tmpolaczyk:
Follow-up to #408, but we can't remove this one.


girazoki (approved):
LGTM

PR #417, opened 2024-02-16
Apply current clippy lints

nanocryk:
Fix what clippy currently complains about.
In a future PR we could add more agressive clippy lints if we want, or agree on some coding patterns not enforced by clippy.


tmpolaczyk:
Can we park this until after the polkadot 1.6.0 upgrade? Because this may cause conflicts, and that PR also updates the rust version which will add new lints that will also need to be fixed.

PR #418, opened 2024-02-16
Update rust toolchain to 1.75.0

tmpolaczyk:


girazoki (approved):

tmpolaczyk:
It's not possible to upgrade to Rust 1.75.0 while still using polkadot-sdk 1.3.0, because of this compile error when running cargo test:

```
error[E0275]: overflow evaluating the requirement `<Runtime as pallet_grandpa::Config>::KeyOwnerProof == _`
```

So closing, the upgrade will be done in #413

PR #424, opened 2024-02-22
MD335 - adds pallet multisig to dancebox/flashbox

augustocollado:
This PR adds the pallet multisig to the dancebox, flashbox, substrate template, and frontier template runtimes.

It also includes moonwall unit tests to:
- create and cancel a multisig call
- create and approve a multisig call


girazoki:
Let's also add them to the container-chains!

container-chains/templates/frontier/runtime/src/lib.rs
```diff
@@ -165,6 +165,24 @@ pub type Executive = frame_executive::Executive<
     AllPalletsWithSystem,
 >;
 
+pub mod currency {
+    use super::Balance;
+
+    // Provide a common factor between runtimes based on a supply of 10_000_000 tokens.
+    pub const SUPPLY_FACTOR: Balance = 100;
```
girazoki, line 172: Did you copy this from moonbeam? I dont think this is needed here for now

girazoki: I mean the supply factor part

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -142,6 +142,24 @@ pub type Executive = frame_executive::Executive<
     AllPalletsWithSystem,
 >;
 
+pub mod currency {
+    use super::Balance;
+
+    // Provide a common factor between runtimes based on a supply of 10_000_000 tokens.
+    pub const SUPPLY_FACTOR: Balance = 100;
```
girazoki, line 149: Same thing here, not needed

girazoki (approved):
Besides small comments, looks good, great job!

container-chains/templates/frontier/runtime/src/lib.rs
augustocollado: I copied that from Tanssi actually
Should I delete it?

container-chains/templates/simple/runtime/src/lib.rs
```diff
@@ -608,6 +626,24 @@ impl tp_impl_tanssi_pallets_config::Config for Runtime {
     type AuthoritiesNotingWeights = pallet_cc_authorities_noting::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+    // One storage item; key size 32 + 20; value is size 4+4+16+20. Total = 1 * (52 + 44)
+    pub const DepositBase: Balance = currency::deposit(1, 96);
+    // Additional storage item size of 20 bytes.
+    pub const DepositFactor: Balance = currency::deposit(0, 20);
```
tmpolaczyk, line 633: These factors are assuming a 20 byte account id. So the are correct for the frontier template, but incorrect for simple template and tanssi, which use a 32 byte account id. You can copy the values from asset-hub: 

https://github.com/paritytech/polkadot-sdk/blob/6c5a42a690f96d59dbdf94640188f5d5b5cc47e2/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L415

tmpolaczyk (approved):

container-chains/templates/simple/runtime/src/lib.rs
tmpolaczyk: Same in runtime/dancebox/src/lib.rs and runtime/flashbox/src/lib.rs

girazoki:
Looks good Augusto!

PR #425, opened 2024-02-22
Clone ProposerFactory to fix issue with metrics

tmpolaczyk:


girazoki:
Tested it with zombienet and the duplicate metric registration dissappears. Good job @tmpolaczyk !

girazoki (approved):

PR #426, opened 2024-02-23
Use precompiled wasm in dev tests

tmpolaczyk:
`dev_tanssi` tests spawn a new `tanssi-node` process for each test. This node needs to compile the runtime wasm. That can take a long time. This pull request adds a flag to use a precompiled wasm, which means that now we only need to compile the wasm once.

Affects dev tests for tanssi, flashbox, and both templates.

Speeds up coverage job from 1h 20min to 1h 3m

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -433,7 +433,7 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "asset-test-utils"
 version = "1.0.0"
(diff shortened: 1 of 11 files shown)


tmpolaczyk:
Coverage report looks wrong,

[/pallets/collator-assignment/src/weights.rs](https://d3gkbsry1ehhqi.cloudfront.net/tanssi-coverage/pulls/426/html/pallets/collator-assignment/src/weights.rs.html) 	50.00% (-50.00%)

This already has 50% coverage in master, so it should be -0%: https://d3gkbsry1ehhqi.cloudfront.net/tanssi-coverage/branches/master/html/pallets/collator-assignment/src/weights.rs.html

girazoki (approved):
LGTM

PR #428, opened 2024-02-26
Use wasmtime-precompiled in zombienet tests

tmpolaczyk:
Compile wasm once before starting all the zombienet nodes, instead of having each node compile it. This also makes rotations a bit faster, we will see if the block delay on rotation is reduced.

Also disable hardware benchmarks, they were enabled for collators and full nodes.


tmpolaczyk:
Surprisingly this doesn't make the zombienet tests faster, but it will make them use less cpu, so hopefully that will make timeout errors less frequent...

girazoki (approved):

PR #436, opened 2024-02-29
Move clippy lints to Cargo.toml file and deny clippy warnings in CI

tmpolaczyk:
Move clippy lints to Cargo.toml file and deny warnings in CI.

~~And use the same lints as polkadot-sdk~~ actually we are using more restrictive ones

Includes a script that checks that all toml files contain `[lints] workspace = true`.


girazoki:
@nanocryk can we add this check as a parameter to toml-maid?

tmpolaczyk:
> @nanocryk can we add this check as a parameter to toml-maid?

I'm not sure if this fits the purpose of the tool, so I wrote a bash script that should be good enough.

Cargo.toml
```diff
@@ -17,6 +17,18 @@ resolver = "2"
 authors = [ "Moondance Labs" ]
 repository = "https://github.com/moondance-labs/tanssi"
 
+[workspace.lints.clippy]
+complexity = { level = "deny", priority = 1 }
+correctness = { level = "deny", priority = 1 }
+erasing_op = { level = "allow", priority = 2 }
+identity_op = { level = "allow", priority = 2 }
+suspicious = { level = "deny", priority = 1 }
+too-many-arguments = { level = "allow", priority = 2 }
```
nanocryk, line 26: Oh, that's how you deal with overriding parts of a group!

tmpolaczyk: Yes, the priority is used as the order in which the flags are passed to the compiler. Previously when using `RUSTFLAGS` we would add an exception for the `too_many_arguments` lint like this:

```
-D clippy::complexity -A clippy::too_many_arguments
```

Because this wouldn't work:

```
-A clippy::too_many_arguments -D clippy::complexity
```

nanocryk (approved):

PR #446, opened 2024-03-08
Store slot number in author-noting

girazoki:
Stores the latest slot number seeing by container-chains. This is useful in case some checks need to be done against the slot number, or to make rewarding dependant on certain slot conditions being met.

No required migration **until** we need to read that value at some point (right now it's not happening, but probably will in a follow-up)


pallets/author-noting/src/mock.rs
```diff
@@ -119,7 +119,7 @@ impl mock_data::Config for Test {}
 #[derive(Clone, Encode, Decode, PartialEq, sp_core::RuntimeDebug, scale_info::TypeInfo)]
 #[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
 pub struct Mocks {
-    pub container_chains: BoundedVec<ParaId, ConstU32<5>>,
+    pub container_chains: BoundedVec<ParaId, ConstU32<100>>,
```
girazoki, line 122: Benchmark tests were broken, hence this change

tmpolaczyk (approved):

PR #448, opened 2024-03-11
Add lookahead collator and enable async backing in all runtimes

Agusrodri:
### What does it do?
 Following https://github.com/moondance-labs/tanssi/pull/399, this PR adds an implementation for the `lookahead` collator client. 

With this collator, we now can enable asynchronous backing and leverage all its new functionalities, like the six seconds blocktime and also producing parachain blocks on top of other para-blocks that were not included in the relay yet (unincluded segment). Without the `lookahead` collator, these features could not be implemented as the other kind of collator (`basic`) allows to produce blocks only in the old way without async-backing enabled.


primitives/consensus/src/lib.rs
```diff
@@ -58,8 +58,14 @@ where
             "Slot duration cannot be zero."
         );
 
+        log::info!("ON_TIMESTAMP_SET: MOMENT: {}", moment);
```
Agusrodri, line 61: Reminder to remove these logs

girazoki:
@Agusrodri the zombienet tests do not work because you are pointing at a custom polkadot branch. Once you cherry-pick in the 1.6.0 branch all will be good

node/src/service.rs
```diff
@@ -73,10 +73,16 @@ use {
...
+        collators::lookahead::{
+            self as lookahead_tanssi_aura, Params as LookaheadTanssiAuraParams,
+        },
+        //collators::basic::{self as basic_tanssi_aura, Params as BasicTanssiAuraParams},
         OrchestratorAuraWorkerAuxData,
     },
-    tokio::sync::mpsc::{unbounded_channel, UnboundedSender},
+    tokio::sync::{
+        mpsc::{unbounded_channel, UnboundedSender},
+        watch::Receiver,
```
tmpolaczyk, line 84: Can we use `watch::Receiver<()>` instead of `Receiver<()>` in this file? Because there may be other kinds of `Receiver`s

```suggestion
        watch,
```

girazoki:
Let's not forget to modify the inflation rate to match the new expectations

girazoki:
@Agusrodri make sure you merge master in please, that should make your tests more robust against the CI

node/src/service.rs
```diff
@@ -73,10 +73,16 @@ use {
     std::{pin::Pin, sync::Arc, time::Duration},
     substrate_prometheus_endpoint::Registry,
     tc_consensus::{
-        collators::basic::{self as basic_tanssi_aura, Params as BasicTanssiAuraParams},
+        collators::lookahead::{
+            self as lookahead_tanssi_aura, Params as LookaheadTanssiAuraParams,
+        },
+        //collators::basic::{self as basic_tanssi_aura, Params as BasicTanssiAuraParams},
```
girazoki, line 79:
```suggestion
```

node/src/service.rs
```diff
@@ -398,6 +404,8 @@ async fn start_node_impl(
         sync_service: node_builder.network.sync_service.clone(),
     })?;
 
+    let (end_lookahead_sender, end_lookahead_receiver) = tokio::sync::watch::channel(());
```
girazoki, line 407:
```suggestion
    // This channel allows us to notify the lookahead collator when it should stop. Useful when rotating containers
    let (end_lookahead_sender, end_lookahead_receiver) = tokio::sync::watch::channel(());
```

girazoki (approved):

client/consensus/Cargo.toml
```diff
@@ -47,10 +47,12 @@ cumulus-relay-chain-interface = { workspace = true }
 
 # Polkadot
 polkadot-node-primitives = { workspace = true }
+polkadot-node-subsystem = { workspace = true }
```
girazoki, line 50: Where are we using this dependency?

test/configs/zombieTanssi.json
```diff
@@ -46,6 +61,7 @@
                 },
                 {
                     "name": "Collator2000-01",
+                    "ws_port": "9960",
```
girazoki, line 64: we dont need these

client/consensus/Cargo.toml
Agusrodri: It is used for some types inside the lookahead collator https://github.com/moondance-labs/tanssi/blob/agustin-lookahead-collator/client/consensus/src/collators/lookahead.rs#L45

test/configs/zombieTanssi.json
Agusrodri: Yes, I changed this for local testing only. I'll remove it.

client/consensus/Cargo.toml
girazoki: thanks

client/consensus/src/collators/lookahead.rs
```diff
@@ -0,0 +1,540 @@
...
+
+        // If we move forward without marking the value as unchanged,
+        // the channel will assume that the value has already changed
+        // in a different step than the one we want (inside containerChainSpawner)
+        // and will not kill the lookahead collator properly if it was already running.
+        if let Some(end_lookahead_receiver) = &mut params.end_lookahead_receiver {
+            end_lookahead_receiver.mark_unchanged();
+        }
+
+        while let Some(relay_parent_header) = import_notifications.next().await {
```
ParthDesai, line 189: Instead of using `tokio::sync::watch::Receiver`, you can use `CancellationToken` (this is common pattern in async programming.) 

https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html

This does exactly what we want to achieve and also has support for child tasks which can also be cancelled once parent token is cancelled. You can use it like this:

```rust
select! {
   _ = token.cancelled() => {
                // The token was cancelled
     }, 
   parent_header = import_notifications.next() => {
       // Call main collator function
   },
}
```

client/consensus/src/collators/lookahead.rs
```diff
@@ -0,0 +1,540 @@
...
+            if let Some(end_lookahead_receiver) = &mut params.end_lookahead_receiver {
```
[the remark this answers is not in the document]
tmpolaczyk: Yes, when rotating from one chain to another we need to restart the collator, because the params are different (client, para_id, slot_duration, ...)

client/consensus/src/collators/lookahead.rs
tmpolaczyk: Didn't know about this, it looks almost exactly what we need here. The only problem is that we need to cancel one task, then spawn a new one, and then later cancel the new task again. I'm not sure if we can implement that using a single `CancellationToken` (we would need some kind of token factory?).

I think the current solution is good enough because it allows use to use the same `Sender` to cancel all the tasks, and both the sender and the receiver are pretty clean. But it's good to know about this primitive, I didn't find it when looking for solutions for this problem so thanks.

ParthDesai: You can just create a new version by doing this:
```
let token = CancellationToken::new();
```

Plus, you can technically use child tokens and if you cancel main token the main and the child tasks are cancelled. 

Problem with current solution is that it is more error prone as you need to mark changed/unchanged in two different places, basically write some wrapper code to tailer it to our specific need.  But, with cancellation token, it is exactly our use case and we do not need any wrapper there.

client/consensus/src/collators/lookahead.rs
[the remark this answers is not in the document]
ParthDesai: Make sense.

client/consensus/src/collators/lookahead.rs
Agusrodri: We could test the new approach using `CancellationToken` yes, but I would only implement it if we get the same results (mostly in logs) as we are getting with the current way. I'm willing to implement what we decide best, what's your opinion on this @girazoki?

tmpolaczyk: I would do it in a separate PR tbh, as I expect it to not be a small refactor

tmpolaczyk (approved):

runtime/dancebox/src/lib.rs
```diff
@@ -1325,12 +1325,12 @@ parameter_types! {
...
     // The equation to solve is:
-    // initial_supply * (1.05) = initial_supply * (1+x)^2_629_800
-    // we should solve for x = (1.05)^(1/2_629_800) -1 -> 0.000000019 per block or 19/1_000_000_000
+    // initial_supply * (1.05) = initial_supply * (1+x)^5_259_600
+    // we should solve for x = (1.05)^(1/5_259_600) -1 -> 0.000000009 per block or 9/1_000_000_000
     // 1% in the case of dev mode
     // TODO: check if we can put the prod inflation for tests too
     // TODO: better calculus for going from annual to block inflation (if it can be done)
-    pub const InflationRate: Perbill = prod_or_fast!(Perbill::from_parts(19), Perbill::from_percent(1));
+    pub const InflationRate: Perbill = prod_or_fast!(Perbill::from_parts(9), Perbill::from_percent(1));
```
ParthDesai, line 1333: Why did we change inflation rate?

runtime/dancebox/tests/common/mod.rs
```diff
@@ -150,7 +150,7 @@ pub fn set_parachain_inherent_data() {
     let relay_sproof = RelayStateSproofBuilder {
         para_id: 100u32.into(),
         included_para_head: Some(HeadData(vec![1, 2, 3])),
-        current_slot: (current_slot() * 2).into(),
+        current_slot: (current_slot()).into(),
```
ParthDesai, line 153: What is the reasoning behind this change?

ParthDesai (approved):
Looks good to me.  As discussed, we will handle cancellation token in different PR.

Some minor questions, which are not blocker.

runtime/dancebox/src/lib.rs
Agusrodri: We must change the inflation rate as we now have the double amount of blocks per year, thus need to calculate the new proper amount of inflation per-block.

runtime/dancebox/tests/common/mod.rs
Agusrodri: As we now have the same clocktime as the relay (6 seconds), we don't need to double the slot that we mock. Previously we were forced to do this because for each para-block (with 12 seconds blocktime), we were receiving a relay slot that was the double of the previous one.

PR #450, opened 2024-03-11
Add compile time tests for pallet_balances consts

tmpolaczyk:
Add compile time checks for some constants used in pallet_balances.

For example, this will make it impossible to add a new pallet that declares a freeze reason when `MaxFreezes` is set to 0, because then the runtime will not compile.

Also fixes that some runtimes had an invalid value for `FreezeIdentifier`. Shouldn't matter because we don't use freezes in those runtimes yet:

```diff
-type FreezeIdentifier = [u8; 8];
+type FreezeIdentifier = RuntimeFreezeReason;
```

TODO: move this tests to a separate module, and perhaps use a macro to call them.


tmpolaczyk:
Closing this, will reopen a PR with regular tests, mainly because `assert_eq` doesn't work in `const fn`, so I think it will be better long term to use a regular integration test.

PR #451, opened 2024-03-12
Set correct FreezeIdentifier in all runtimes

tmpolaczyk:
This won't break anything because we don't use freezes in any of the runtimes.

### DIFF
--- a/pallets/stream-payment/src/lib.rs
+++ b/pallets/stream-payment/src/lib.rs
@@ -361,6 +361,12 @@ pub mod pallet {
         },
     }
 
+    /// Freeze reason to use if needed.
+    #[pallet::composite_enum]
+    pub enum FreezeReason {
+        StreamPayment,
+    }
+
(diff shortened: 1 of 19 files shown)


girazoki (approved):
Looks good to me

pallets/stream-payment/src/lib.rs
```diff
@@ -361,12 +361,6 @@ pub mod pallet {
         },
     }
 
-    /// Freeze reason to use if needed.
```
nanocryk, line 364: Since the actual lock/freeze implementation is done outside of the pallet, I left both FreezeReason and HoldReason if one need it.

tmpolaczyk: Ok, should I revert this change then?

girazoki: oh I see, that is interesting. Then we can leave it yes

PR #452, opened 2024-03-12
Implement pallet_xcm_core_buyer

tmpolaczyk:
This pallet allows collators to buy parathread cores on demand.

It uses XCM to send a message to the relay chain, to the extrinsic `OnDemandAssignmentProvider.place_order_allow_death`. The core and the fees are paid using the parathread tank account in the relay chain, which is an account derived through XCM where parathread managers are supposed to send DOT tokens to.

It provides 3 extrinsics:

* buy_core

(won't be implemented in this PR) Any collators that are currently assigned to a parathread can call this extrinsic to send an XCM message to the relay chain. We use an `ensure_none` origin to allow collators to call this extrinsic without paying any fees by providing a proof (proof validation is not implemented, so currently all calls will fail). Only 1 XCM message is allowed per tanssi block per parathread.

* force_buy_core

Same as `buy_core` but can only be called by root. Instead of a proof, we simply check that at least one collator is assigned to the para_id. We need this check because it doesn't make sense to buy a core for a parathread that doesn't have any assigned collators, because nobody will produce the block. This is mainly used in tests because the other extrinsic is not implemented yet.

* set_xcm_weights

Only callable by root, this is used to set a storage item which contains the withdraw amount needed to `BuyExecution`, and the `weight_at_most` of the `place_order_allow_death` call.  This is a storage item because relay chain weights can change, so we need to be able to adjust them without doing a runtime upgrade.


runtime/dancebox/tests/common/xcm/core_buyer.rs
```diff
@@ -0,0 +1,410 @@
...
+        assert_eq!(balance_before, balance_after);
+    });
+}
+
+#[test]
+fn xcm_core_buyer_only_enough_balance_for_buy_execution() {
+    let parathread_tank_in_relay = get_parathread_tank_relay_address();
+    let balance_before = BUY_EXECUTION_COST;
+
+    // Invariant: if balance_before >= BUY_EXECUTION_COST then balance_after <= (balance_before - BUY_EXECUTION_COST)
```
tmpolaczyk, line 207:
```suggestion
    // Invariant: if balance_before >= BUY_EXECUTION_COST then balance_after <= (balance_before + BUY_EXECUTION_REFUND - BUY_EXECUTION_COST)
```

In this test the final balance is 0 because `BUY_EXECUTION_REFUND < ROCOCO_ED`, so the account gets the refund but it is immediatelly killed.

primitives/traits/src/lib.rs
```diff
@@ -23,6 +23,7 @@ pub use cumulus_primitives_core::{
     relay_chain::{BlockNumber, Slot},
     ParaId,
 };
+use frame_support::__private::serde;
```
tmpolaczyk, line 26:
```suggestion
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+    staging_xcm::prelude::*,
+    staging_xcm::v3::{InteriorMultiLocation, MultiAsset, MultiAssets, Xcm},
+    tp_traits::ParathreadParams,
+};
+
+#[frame_support::pallet]
+pub mod pallet {
+    use super::*;
+
+    /// Data preservers pallet.
```
girazoki, line 55:
```suggestion
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        type GetBlockNumber: Get<u32>;
+        type GetParathreadAccountId: Convert<ParaId, [u8; 32]>;
+        type SelfParaId: Get<ParaId>;
+        type MaxParathreads: Get<u32>;
+        type GetParathreadParams: GetParathreadParams;
+        type GetAssignedCollators: GetParathreadCollators<Self::AccountId>;
+        /// A configuration for base priority of unsigned transactions.
+        ///
+        /// This is exposed so that it can be tuned for particular runtime, when
+        /// multiple pallets send unsigned transactions.
```
girazoki, line 77: Had no idea about this, but it's very nice

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        type Currency: Inspect<Self::AccountId> + Balanced<Self::AccountId>;
+
+        type XcmSender: SendXcm;
+        type GetPurchaseCoreCall: GetPurchaseCoreCall;
+        type GetBlockNumber: Get<u32>;
+        type GetParathreadAccountId: Convert<ParaId, [u8; 32]>;
+        type SelfParaId: Get<ParaId>;
+        type MaxParathreads: Get<u32>;
+        type GetParathreadParams: GetParathreadParams;
+        type GetAssignedCollators: GetParathreadCollators<Self::AccountId>;
```
girazoki, line 73: can we add a comment mentioning why we need this? also the upper variables

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+    #[pallet::storage]
+    //pub type InFlightOrders<T: Config> = StorageMap<_, Blake2_128Concat, ParaId, (), OptionQuery>;
+    pub type InFlightOrders<T: Config> =
+        StorageValue<_, BoundedBTreeSet<ParaId, T::MaxParathreads>, ValueQuery>;
+
+    /// This must be set by root with the value of the relay chain xcm call weight and extrinsic
+    /// weight limit. This is a storage item because relay chain weights can change, so we need to
+    /// be able to adjust them without doing a runtime upgrade.
+    #[pallet::storage]
+    pub type XcmWeights<T: Config> = StorageValue<_, XcmWeightsTy<T>, OptionQuery>;
```
girazoki, line 133: I dont particularly like this naming as they are not all weights (we have expenses plus weigth). But I cant only think of too generic names like "XcmParams"

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+
+            // TODO: also compare the latest slot from pallet_author_noting with parathread_params.slot_frequency
+
+            let xcm_weights_storage =
+                XcmWeights::<T>::get().ok_or(Error::<T>::XcmWeightStorageNotSet)?;
+
+            let withdraw_amount = xcm_weights_storage.buy_execution_cost;
+
+            // Send xcm to the relay
+            // Use a derivative account from the sovereign account based on the paraId
```
girazoki, line 264:
```suggestion
            // Use the account derived from the multilocation composed with DescendOrigin
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+            // Buy on-demand cores
+            // Any failure should return everything to the derivative account
+
+            // Don't use utility::as_derivative because that will make the tanssi sovereign account
+            // pay for fees, instead use `DescendOrigin` to make the container chain sovereign account
+            // pay for fees. The container chain sovereign account is derived from the tanssi sovereign
```
girazoki, line 270:
```suggestion
            // pay for fees. The container chain sovereign account is derived from the descended MultiLocation by appending the a 32 byte junction containing the para-id appended by zeros
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+            // account.
+            // TODO: when coretime is implemented, buy core there instead of buying on-demand cores
```
girazoki, line 272:
```suggestion
            // TODO: when coretime is implemented, use coretime instantaneous credits instead of buying on-demand cores at the price defined by the relay
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+            let origin = OriginKind::SovereignAccount;
+            // TODO: max_amount is the max price of a core that this parathread is willing to pay
+            // It should be defined in a storage item somewhere, controllable by the container chain
+            // manager.
+            let max_amount = u128::MAX;
```
girazoki, line 277: good point. Where do you envision this to be? services-payment maybe?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+            let call = T::GetPurchaseCoreCall::get_encoded(max_amount, para_id);
+            let weight_at_most = xcm_weights_storage.weight_at_most;
+
+            // Assumption: derived account already has DOT
+            // The balance should be enough to cover
+            // TODO: we could make this be part of the proof, so collators cannot call this if the
+            // derived account does not have enough balance
```
girazoki, line 284: I strongly oppose to this idea, for the reasons that we discussed already. We can check that the balance is correct in the block at which we are builiding, but have no guarantees that when the XCM reaches the relay this is true

girazoki: Ah I see you already say that in the next line :)

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+            let relay_asset_total: MultiAsset = (Here, withdraw_amount).into();
+            let refund_asset_filter: MultiAssetFilter =
+                MultiAssetFilter::Wild(WildMultiAsset::AllCounted(1));
+            // TODO: need better names for this methods.
+            //  interior_multilocation is the one used in DescendOrigin
+            //  absolute_multilocation is the one used in DepositAsset
+            // They can be easily converted from one another, the difference is that absolute_multilocation
+            // has an extra "Parachain" junction in the front, using SelfParaId::get()
+            let interior_multilocation = Self::interior_multilocation(para_id);
+            let derived_account = Self::absolute_multilocation(interior_multilocation);
```
girazoki, line 297: There should be an easier way to do this, I think there is a location inverter at some point that we could incorporate

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Yes, I also need a better name for `XcmWeightsTy`

runtime/dancebox/src/xcm_config.rs
```diff
@@ -456,3 +457,153 @@ impl pallet_message_queue::Config for Runtime {
...
+                para_id,
+            },
+        );
+
+        call.encode()
+    }
+}
+
+// TODO: create a crate for rococo_calls
+pub mod rococo_calls {
```
girazoki, line 564: Yep, crate for this

girazoki:
Overall looks good. I think we are missing typescript tests as well for this

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        account: T::AccountId,
+        // TODO
+        _signature: (),
+    }
+
+    /// Set of parathreads that have already sent an XCM message to buy a core recently.
+    /// Used to avoid 2 collators buying a core at the same time, because it is only possible to buy
+    /// 1 core in 1 relay block.
+    #[pallet::storage]
+    //pub type InFlightOrders<T: Config> = StorageMap<_, Blake2_128Concat, ParaId, (), OptionQuery>;
```
Agusrodri, line 125: Are we using the `StorageValue` instead of the commented `StorageMap` at the end? If so, we can remove this line.
```suggestion
```

pallets/xcm-core-buyer/src/lib.rs
Agusrodri: Maybe replace:

- `XcmWeights` by `RelayXcmWeightConfig`?
- `XcmWeightsTy` by `RelayXcmWeightConfigInner`?

(just a suggestion)

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+    impl<T: Config> Pallet<T> {
+        /// Buy a core for this parathread id.
+        /// Collators should call this to indicate that they intend to produce a block, but they
+        /// cannot do it because this para id has no available cores.
+        /// The purchase is automatic using XCM, and collators do not need to do anything.
+        // Note that the collators that will be calling this function are parathread collators, not
+        // tanssi collators. So we cannot force them to provide a complex proof, e.g. against relay
+        // state.
+        #[pallet::call_index(0)]
+        // TODO: weight
```
Agusrodri, line 153:
```suggestion
```

tmpolaczyk: Actually this is still using the wrong weight, because I expect it to be more expensive than `force_buy_core` because of the proof validation. But proof validation is not being implemented in this PR, so it is not possible to write a benchmark for this.

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// Overarching event type.
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+        type Currency: Inspect<Self::AccountId> + Balanced<Self::AccountId>;
+
+        type XcmSender: SendXcm;
+        type GetPurchaseCoreCall: GetPurchaseCoreCall;
+        type GetBlockNumber: Get<u32>;
+        type GetParathreadAccountId: Convert<ParaId, [u8; 32]>;
```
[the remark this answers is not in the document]
tmpolaczyk: Done, that is now available as the struct `ParaIdIntoAccountTruncating`

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Documented most config types

pallets/xcm-core-buyer/src/lib.rs
girazoki: yeah I think that should work

pallets/xcm-core-buyer/src/lib.rs
girazoki: it's the most idiomatic way to do it (it is meant for that). Comments should make sure we can understand it

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+        /// The `interior_multilocation` can be obtained using `Self::interior_multilocation`.
+        pub fn absolute_multilocation(
+            interior_multilocation: InteriorMultiLocation,
+        ) -> MultiLocation {
+            let relay_chain = MultiLocation::parent();
+            let context = Parachain(T::SelfParaId::get().into()).into();
+            let mut reanchored: MultiLocation = interior_multilocation.into();
+            reanchored
+                .reanchor(&relay_chain, context)
+                .expect("reanchor failed");
```
girazoki, line 248: Remove this expect, let's just return a pallet error

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+            let relay_asset_total: MultiAsset = (Here, withdraw_amount).into();
+            let refund_asset_filter: MultiAssetFilter =
+                MultiAssetFilter::Wild(WildMultiAsset::AllCounted(1));
+            // TODO: need better names for this methods.
+            //  interior_multilocation is the one used in DescendOrigin
+            //  absolute_multilocation is the one used in DepositAsset
+            // They can be easily converted from one another, the difference is that absolute_multilocation
+            // has an extra "Parachain" junction in the front, using SelfParaId::get()
+            let interior_multilocation = Self::interior_multilocation(para_id);
+            let derived_account = Self::absolute_multilocation(interior_multilocation);
```
girazoki, line 305:
```suggestion
            let derived_account = Self::relay_relative_multilocation(interior_multilocation);
```

Agusrodri: We could also rename `interior_multilocation` to something like `local_relative_multilocation` or `local_interior_multilocation`, to emphasize on the fact that the point of view of that multilocation is the local origin.

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Probably, created a new issue to track this

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,494 @@
...
+
+        type XcmSender: SendXcm;
+        /// Get encoded call to buy a core in the relay chain. This will be passed to the XCM
+        /// `Transact` instruction.
+        type GetPurchaseCoreCall: GetPurchaseCoreCall<Self::RelayChain>;
+        /// Get current block number, used in `validate_unsigned`.
+        type GetBlockNumber: Get<u32>;
+        /// How to convert a `ParaId` into an `AccountId32`. Used to derive the parathread tank
+        /// account in `interior_multilocation`.
+        type GetParathreadAccountId: Convert<ParaId, [u8; 32]>;
```
ParthDesai, line 73: Question. Who owns this account? Is it a chain identified by the corresponding `para_id`?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -0,0 +1,494 @@
...
+        ReanchorFailed,
+    }
+
+    /// Proof that I am a collator, assigned to a para_id, and I can buy a core for that para_id
+    #[derive(Encode, Decode, CloneNoBound, PartialEq, Eq, DebugNoBound, TypeInfo)]
+    #[scale_info(skip_type_params(T))]
+    pub struct BuyCoreCollatorProof<T: Config> {
+        account: T::AccountId,
+        // TODO
+        _signature: (),
```
ParthDesai, line 138: This signature is required, since we are spending tokens from the parathread tank account right?

pallets/xcm-core-buyer/src/lib.rs
girazoki: yes exactly, it's a chain-derived account based on the paraId.

pallets/xcm-core-buyer/src/lib.rs
girazoki: I think we decided not to make this PR bigger with the validation of the signature but @tmpolaczyk correct me if I am wrong. Also in that case let's update the TODO

pallets/xcm-core-buyer/src/lib.rs
girazoki: for now this address is only usable for this purpose. However, in the future we might want the container-chain manager to have some power over this

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Yes, the validation logic will be a separate PR so that it can be reviewed better

runtime/relay-encoder/src/rococo.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) Moondance Labs Ltd.
```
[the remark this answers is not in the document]
tmpolaczyk: Done

girazoki (approved):
For me the current state of what has to be done in this PR is an approval. Missing pieces that should come as follow-ups are:

- validation logic of the tx
- slot checking (possibly with cooldown) within this pallet

Good job @tmpolaczyk

PR #453, opened 2024-03-12
Use `CollatorId` instead of `AccountId` - Part 1 of N

ParthDesai:
## Description
Currently, we are using `AccountId` and `CollatorId` interchangeably in the runtime. This works right now, since our `CollatorId` config type is set to  `AccountId` in runtime.

To be more idiomatic, we need to replace direct use of `AccountId` with `CollatorId` both in pallets and runtime. With `CollatorId` currently only type alias of `AccountId`. This make sure that both types are equivalent for now, but can be changed just by setting different type while building runtime.

### What's changed?
This PR modifies `pallet_invulnerables` to use `CollatorId` instead of `AccountId` with some minor refactor involved.

In the runtime, we define type alias, and replace `AccountId` with `CollatorId` at required places.

## Note
This PR does not completely modifies runtime and all pallets to use `CollatorId`, only `pallet_invulnerables` and defining the core distinction between `AccountId` and `CollatorId` in runtime.

The changes involved to do it completely will require quite a few PR both in this repository and dancekit.

### DIFF
--- a/pallets/invulnerables/src/benchmarking.rs
+++ b/pallets/invulnerables/src/benchmarking.rs
@@ -77,22 +77,32 @@ fn keys<T: Config + session::Config>(c: u32) -> <T as session::Config>::Keys {
 
 fn invulnerable<T: Config + session::Config + pallet_balances::Config>(
     c: u32,
-) -> (T::AccountId, <T as session::Config>::Keys) {
-    (create_funded_user::<T>("candidate", c, 100), keys::<T>(c))
+) -> (T::AccountId, T::CollatorId, <T as session::Config>::Keys) {
+    let funded_user = create_funded_user::<T>("candidate", c, 100);
+    let collator_id = T::CollatorIdOf::convert(funded_user)
+        .expect("Converstion of account id of collator id failed.");
+    (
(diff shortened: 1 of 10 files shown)


pallets/invulnerables/src/benchmarking.rs
```diff
@@ -77,22 +77,32 @@ fn keys<T: Config + session::Config>(c: u32) -> <T as session::Config>::Keys {
...
 fn invulnerable<T: Config + session::Config + pallet_balances::Config>(
     c: u32,
-) -> (T::AccountId, <T as session::Config>::Keys) {
-    (create_funded_user::<T>("candidate", c, 100), keys::<T>(c))
+) -> (T::AccountId, T::CollatorId, <T as session::Config>::Keys) {
+    let funded_user = create_funded_user::<T>("candidate", c, 100);
+    let collator_id = T::CollatorIdOf::convert(funded_user)
+        .expect("Converstion of account id of collator id failed.");
+    (
+        create_funded_user::<T>("candidate", c, 100),
```
girazoki, line 85:
```suggestion
       funded_user,
```

girazoki (approved):
LGTM!

PR #463, opened 2024-03-18
Use `SlowAdjustingFeeUpdate` instead of `ConstFeeMultiplier`

ParthDesai:
## Description
Currently, we are using `ConstFeeMultiplier` for transaction payment, which mean that regardless of network congestion level, the transaction fee would be constant. This is not recommended since it can be DOS vector. 

Instead we need to use `SlowAdjustingFeeUpdate` which adjusts slowly, at the pace of long term tendencies, and uses tipping to give users the possibility of controlling waiting times at peak hours.

### DIFF
--- a/runtime/dancebox/tests/common/mod.rs
+++ b/runtime/dancebox/tests/common/mod.rs
@@ -103,7 +103,7 @@ pub fn insert_authorities_and_slot_digests(slot: u64) {
     );
 }
 
-pub fn run_block_with_operation(operation_fn: fn(slot: u64)) -> RunSummary {
+pub fn run_block_with_operation<F: FnOnce(u64)>(operation_fn: F) -> RunSummary {
     let slot = current_slot() + 1;
 
     insert_authorities_and_slot_digests(slot);
--- a/runtime/dancebox/tests/common/mod.rs
+++ b/runtime/dancebox/tests/common/mod.rs
@@ -42,6 +42,7 @@ use {
 
 mod xcm;
 
+use dancebox_runtime::TransactionPayment;
 pub use dancebox_runtime::{
     AccountId, AssetRate, AuthorNoting, AuthorityAssignment, AuthorityMapping, Balance, Balances,
     CollatorAssignment, Configuration, DataPreservers, ForeignAssets, ForeignAssetsCreator,
@@ -102,7 +103,7 @@ pub fn insert_authorities_and_slot_digests(slot: u64) {
     );
 }
 
-pub fn run_block() -> RunSummary {
+pub fn run_block_with_operation(operation_fn: fn(slot: u64)) -> RunSummary {
     let slot = current_slot() + 1;
 
     insert_authorities_and_slot_digests(slot);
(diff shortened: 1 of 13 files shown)


ParthDesai:
Some typescript tests are failing. Looking into it.

girazoki:
Yes I expected that. Also make sure you have rust integration tests and ts-tests for this

girazoki:
And also, if the container-chain template runtimes dont have this change, make sure they include it

ParthDesai:
@girazoki I have fixed ts-tests and added both rust and integration ts-tests. Let me know what you think.

test/suites/common-tanssi/fees/test_fee_balance_transfer.ts
```diff
@@ -91,13 +91,35 @@ describeSuite({
                     signedTx.encodedLength
                 );
 
+                const feeMultiplier = (await polkadotJs.query.transactionPayment.nextFeeMultiplier()).toBigInt();
```
girazoki, line 94: I am wondering whether we should move this to common instead of common-tanssi. I guess all runtimes should behave equally with respect to this. @ParthDesai can you try it?

runtime/dancebox/tests/common/mod.rs
```diff
@@ -102,7 +103,7 @@ pub fn insert_authorities_and_slot_digests(slot: u64) {
     );
 }
 
-pub fn run_block() -> RunSummary {
+pub fn run_block_with_operation(operation_fn: fn(slot: u64)) -> RunSummary {
```
girazoki, line 106: nice!

test/suites/common-tanssi/fees/test_fee_balance_transfer.ts
girazoki: thank you!

girazoki (approved):
Looks good tom me!

runtime/dancebox/tests/common/mod.rs
tmpolaczyk, line 106:
```suggestion
pub fn run_block_with_operation<F: FnOnce(u64)>(operation_fn: F) -> RunSummary {
```

This should be generic to avoid compile errors when the closure captures some state

runtime/dancebox/tests/integration_test.rs
```diff
@@ -5525,3 +5527,45 @@ fn test_migration_services_collator_assignment_payment() {
...
+                    &ConsumedWeight::new(|class| {
+                        max_block_weights
+                            .get(class)
+                            .max_total
+                            .unwrap_or(Weight::MAX)
+                    }),
+                );
+            });
+            let current_multiplier = TransactionPayment::next_fee_multiplier();
+            assert!(current_multiplier.gt(&before_multiplier));
```
tmpolaczyk, line 5558:
```suggestion
            assert!(current_multiplier > before_multiplier);
```

Does this work?

ParthDesai:
@tmpolaczyk I have applied your suggestions. Please review and if there is nothing else, please approve it.

tmpolaczyk (approved):

PR #465, opened 2024-03-18
MD364 - Adds session key management proxy type

augustocollado:
This PR allows the delegating of key management
It adds:

- New proxy type 'SessionKeyManagement' in Dancebox runtime
- It adds unit testing for this new type

### DIFF
--- a/runtime/dancebox/src/lib.rs
+++ b/runtime/dancebox/src/lib.rs
@@ -1057,7 +1057,7 @@ pub enum ProxyType {
     Registrar = 6,
     /// Allow extrinsics related to Registrar that needs to be called through Sudo
     SudoRegistrar = 7,
-    /// Allow extrinsics from the Session pallet for key management. 
+    /// Allow extrinsics from the Session pallet for key management.
     SessionKeyManagement = 8,
 }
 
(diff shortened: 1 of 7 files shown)


girazoki:
@augustocollado you should not need any cargo.lock changes to make this PR, make sure there are not please, I think your tests are not passing because of these

girazoki (approved):
LGTM

runtime/dancebox/src/lib.rs
```diff
@@ -1057,6 +1057,8 @@ pub enum ProxyType {
     Registrar = 6,
     /// Allow extrinsics related to Registrar that needs to be called through Sudo
     SudoRegistrar = 7,
+    /// Allow extrinsics from the Session pallet for key management.
+    SessionKeyManagement = 8,
```
tmpolaczyk, line 1061: We don't need this in flashbox?

augustocollado: I don't think so. 
This came up as a requirement for node operators to decouple staking from key management, and there will be no external node operators on flashbox.

girazoki: Yeah exactly we could put it but no-one is going to use it

tmpolaczyk (approved):

PR #466, opened 2024-03-19
Tests reserve transfers

girazoki:
Some tests regarding the `transfer_assets` extrinsic in `PolkadotXcm`. I cherry-picked https://github.com/paritytech/polkadot-sdk/pull/3450. so that delivery fees work in the latest-case


Agusrodri (approved):
Looks good! ~Regarding the tests for delivery fees, I think it's fine if we add them in another PR~.

Edit: I see you already opened #479, so we could implement those tests there or also in a separate PR would be fine too.

PR #469, opened 2024-03-21
chore(ci): improve dev test duration by not compiling wasm on startup

RomarQ:
This change make the pipeline a bit faster by removing the wasm compilation from every node startup. It generates a raw chain spec at the beginning and reuses it during the test run.

The `coverage` workflow on master takes around `~57 minutes`, and with this change it goes to `~37 minutes`.


girazoki (approved):

girazoki:
thank you @RomarQ !

girazoki (approved):

test/moonwall.config.json
```diff
@@ -19,7 +19,9 @@
                         "name": "tanssi",
                         "binPath": "../target/release/tanssi-node",
                         "options": [
-                            "--dev",
+                            "--alice",
+                            "--tmp",
+                            "--chain=tmp/dancebox_dev-raw.json",
```
tmpolaczyk, line 24: This can introduce race conditions to tests because `--dev` was also enabling some additional options.

The most important one is `--reserved-only` which prevents two nodes spawned at once from connecting to one another, as that will result in them running in a single network instead of each node being a separate network.

So I believe the args should be this:

```
                            "--chain=tmp/dancebox_dev-raw.json",
                            "--sealing=manual",
                            "--alice",
                            "--force-authoring",
                            "--rpc-cors=all",
                            "--tmp",
                            "--no-hardware-benchmarks",
                            "--no-prometheus",
                            "--no-telemetry",
                            "--reserved-only",
                            "--wasmtime-precompiled=wasm"
```

tmpolaczyk (approved):
Thanks a lot!

PR #470, opened 2024-03-22
Implement custom slot duration runtime api for runtimes

tmpolaczyk:
Technically only container chains need this, but if we don't implement it for tanssi then the code does not compile, because we assume that the ContainerChainClient has the same set of RuntimeApi as the tanssi Client

### DIFF
--- a/runtime/flashbox/src/lib.rs
+++ b/runtime/flashbox/src/lib.rs
@@ -1828,6 +1828,12 @@ impl_runtime_apis! {
             TransactionPayment::length_to_fee(length)
         }
     }
+
+    impl dp_slot_duration_runtime_api::TanssiSlotDurationApi<Block> for Runtime {
+        fn slot_duration() -> u64 {
+            SLOT_DURATION
+        }
+    }
 }
 
 struct CheckInherents;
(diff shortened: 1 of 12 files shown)


runtime/flashbox/src/lib.rs
```diff
@@ -1828,6 +1828,12 @@ impl_runtime_apis! {
             TransactionPayment::length_to_fee(length)
         }
     }
+
```
girazoki, line 1831: AFAIK we can now remove the aura-runtime-api and all its requirements from our runtime with this

girazoki: but we can leave it for a separate PR

girazoki:
I tested using this client against runtime 500 containers, and works producing every 12secs. After the upgrade to runtime600, they produce every 6 seconds without any downtime. Good job @tmpolaczyk

girazoki (approved):

runtime/flashbox/src/lib.rs
tmpolaczyk: I prefer not to

PR #471, opened 2024-03-25
Shuffle only top N collators to respect staking priority

tmpolaczyk:
This will give more priority to collators with more stake.

Previously when the number of collators was much greater than the required, all collators were equally likely to be selected. Now, the first ones on the list are more likely to be selected.

Example with max_collators = 100, invulnerables = 5, min_orchestrator_collators = 2, max_orchestrator_collators = 5, and one container chain with 4 collators

* We take the 5 invulnerables, then 95 more collators from staking. These are the 95 with most stake.
* We assign 2 invulnerables to the orchestrator chain (prioritize_invulnerables)
* Now we need 3 collators more for orchestrator chain and 4 collators more for the container chain, 7 in total

Previously: we always shuffle the full list of collators, so we will take 7 collators randomly out of 98 (3 invulnerables + 95 from staking, all equally likely).

After this PR: we first truncate the list to have length = 7 (so 3 invulnerables + 4 from staking, the ones with most stake), and then we shuffle that. It is still possible that a collator with less stake is given priority but only if that collator was previously assigned to that chain, and only until the next rotation. So after a full rotation all the collators with most stake will be assigned to chains. 

We shuffle to avoid always assigning the collators with more stake to the orchestrator chain, and to avoid collators with similar stake to be assigned to the same container chain.

### DIFF
--- a/pallets/collator-assignment/src/tests/assign_full.rs
+++ b/pallets/collator-assignment/src/tests/assign_full.rs
@@ -24,8 +24,9 @@ use {
     sp_std::collections::btree_map::BTreeMap,
 };
 
+#[allow(unused_assignments)]
 fn no_shuffle() -> Option<impl FnOnce(&mut Vec<u64>)> {
-    let mut shuffle = Some(move |collators: &mut Vec<u64>| {
+    let mut shuffle = Some(move |_collators: &mut Vec<u64>| {
         // Create an empty closure to allow the compiler to infer the return type
     });
     // But don't return that closure, instead return `None`
--- a/pallets/collator-assignment/src/tests/assign_full.rs
+++ b/pallets/collator-assignment/src/tests/assign_full.rs
@@ -19,8 +19,20 @@ use {
         assignment::{Assignment, AssignmentError},
         tests::Test,
     },
+    rand::{seq::SliceRandom, SeedableRng},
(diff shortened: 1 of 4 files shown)


girazoki (approved):

girazoki:
LGTM!

pallets/collator-assignment/src/tests/assign_full.rs
```diff
@@ -19,9 +19,22 @@ use {
...
     sp_std::collections::btree_map::BTreeMap,
 };
 
+#[allow(unused_assignments)]
+fn no_shuffle() -> Option<impl FnOnce(&mut Vec<u64>)> {
+    let mut shuffle = Some(move |_collators: &mut Vec<u64>| {
+        // Create an empty closure to allow the compiler to infer the return type
+    });
+    // But don't return that closure, instead return `None`
+    shuffle = None;
```
tmpolaczyk, line 33: I feel that I am doing something wrong here, but I can't just `return None::<_>` because the compiler doesn't know the type of `_`.

tmpolaczyk: Improved this, thanks @nanocryk

ParthDesai (approved):

PR #475, opened 2024-03-26
Collator assignment prioritization on congestion based on tip

fgamundi:


primitives/traits/src/lib.rs
```diff
@@ -31,23 +31,35 @@ use {
...
 
-/// The collator-assignment hook to react to collators beign assigned to container chains.
-pub trait CollatorAssignmentHook {
+/// The collator-assignment hook to react to collators being assigned to container chains.
+pub trait CollatorAssignmentHook<Balance> {
     /// This hook is called when collators are assigned to a container
     ///
     /// The hook should never panic and is required to return the weight consumed.
-    fn on_collators_assigned(para_id: ParaId) -> Weight;
+    fn on_collators_assigned(para_id: ParaId, maybe_tip: &Option<Balance>) -> Weight;
```
tmpolaczyk, line 39:
```suggestion
    fn on_collators_assigned(para_id: ParaId, maybe_tip: Option<&Balance>) -> Weight;
```

This is always better, use `.as_ref()` to convert from `&Option` to `Option<&`

pallets/collator-assignment/src/lib.rs
```diff
@@ -233,6 +233,14 @@ pub mod pallet {
                     max_collators: collators_per_parathread,
                 });
             }
+
+            // Prioritize paras by tip
+            // This doesn't distinguish between parachains and parathreads
+            chains.sort_by(|a, b| {
+                T::CollatorAssignmentTip::get_para_tip(b.para_id)
+                    .cmp(&T::CollatorAssignmentTip::get_para_tip(a.para_id))
+            });
```
tmpolaczyk, line 242: Can the fee be `None` here? Because I believe the order will be `Some(1) > Some(0) > None`. Which I guess is ok? Add a test for this case if you haven't already.

runtime/dancebox/tests/integration_test.rs
```diff
@@ -5527,3 +5527,311 @@ fn test_migration_services_collator_assignment_payment() {
...
+            ));
+
+            // Only set tip for 1003
+            assert_ok!(ServicesPayment::set_max_tip(
+                root_origin(),
+                para_id_with_tip.into(),
+                max_tip,
+            ));
+
+            run_to_session(2);
```
fgamundi, line 5733: In this and the following test, for some reason, if it runs to the 1st session it seems that containers are not getting prioritized (but it does on the prior tests). But running to the 2nd session withdraws the tips corresponding to two sessions. It might have something to do for having the two additional collators (eve, ferdie)

pallets/collator-assignment/src/lib.rs
fgamundi: Yes, by default it's `None`, and the order will be as you described. If all are `None`, then the order shouldn't change

pallets/collator-assignment/src/lib.rs
```diff
@@ -287,6 +295,23 @@ pub mod pallet {
...
+            assigned_containers.retain(|_, v| !v.is_empty());
+
+            // On congestion (not enough collators for all paras) prioritized chains need to pay
+            // the minimum tip of the prioritized chains
+            let maybe_tip: Option<BalanceOf<T>> =
+                if assigned_containers.len() < container_chain_ids.len() + parathreads.len() {
+                    assigned_containers
+                        .into_keys()
+                        .map(T::CollatorAssignmentTip::get_para_tip)
+                        .filter(|o| o.is_some())
```
tmpolaczyk, line 308:
```suggestion
                        .filter_map(T::CollatorAssignmentTip::get_para_tip)
```

Pretty sure there's a clippy lint for this

runtime/dancebox/tests/integration_test.rs
tmpolaczyk: This is related to how assignment is calculated one session in advance. Changes to the list of para ids are only visible after 2 sessions. This explains why we only see the chain 1003 being assigned in the current assignment in session 2: when session 1 starts, the assignment for session 2 will be calculated. That assignment includes the changes to `max_tip`, so chain 1003 will be given priority. But that assignment will be applied in session 2.

The tip is being deducted twice: in session 1 (for the assignment in session 2), and in session 2 (for the assignment in session 3). In the previous tests the fee is only deduced once because we only `run_to_session(1)`, not 2 like here. You can verify this by adding this to the test:

```rust
run_to_session(1);
assert_eq!(
    Balances::usable_balance(ServicesPayment::parachain_tank(para_id_with_tip.into())),
    tank_funds - max_tip,
);
```

pallets/collator-assignment/src/lib.rs
fgamundi: Much better! Clippy didn't warn about this, though https://github.com/moondance-labs/tanssi/actions/runs/8434396122/job/23097544135

pallets/collator-assignment/src/lib.rs
```diff
@@ -236,6 +236,14 @@ pub mod pallet {
                     max_collators: collators_per_parathread,
                 });
             }
+
+            // Prioritize paras by tip
+            // This doesn't distinguish between parachains and parathreads
```
girazoki, line 241: We should put a TODO here. It's likely we want to treat parathread tips differently in the future (as they have different collator needs)

pallets/collator-assignment/src/mock.rs
```diff
@@ -241,6 +241,8 @@ impl pallet_collator_assignment::Config for Test {
     type RemoveInvulnerables = RemoveAccountIdsAbove100;
     type RemoveParaIdsWithNoCredits = RemoveParaIdsAbove5000;
     type CollatorAssignmentHook = ();
+    type CollatorAssignmentTip = ();
```
girazoki, line 244: Shouldnt we have unitests here? I think it would make sense

pallets/collator-assignment/src/lib.rs
```diff
@@ -312,7 +335,7 @@ pub mod pallet {
                     .unwrap_or(&vec![])
                     .is_empty()
                 {
-                    T::CollatorAssignmentHook::on_collators_assigned(*para_id);
+                    T::CollatorAssignmentHook::on_collators_assigned(*para_id, maybe_tip.as_ref());
```
girazoki, line 338: Maybe we can add `is_parathread` to `on_collators_assigned`?

runtime/dancebox/src/lib.rs
```diff
@@ -794,9 +794,13 @@ impl RemoveParaIdsWithNoCredits for RemoveParaIdsWithNoCreditsImpl {
             let free_session_credits = pallet_services_payment::CollatorAssignmentCredits::<Runtime>::get(para_id)
                 .unwrap_or_default();
 
+            // Should have enough tip for 2 sessions
+            let max_tip = pallet_services_payment::MaxTip::<Runtime>::get(para_id).unwrap_or_default() * 2 ;
```
girazoki, line 798: We probably need to revisit this because of the following assumptions:
- We are charging immediately this value (and therefore, we dont have to check 2 sessions but just one).
- It does not even make sense to check for couple of sessions as the container owner can change the maxTip in the middle to something higher or lower
- I think the same thing applies to the session payment itself.

What do you think?

girazoki:
I am missing unitest in the collator-assignment-pallet and some typescript integration tests. For the latter I think we are already having one of the container-chains without collators (in dev service, we register 2000 and 2001 but I think the latter does not get collators). We could implement a higher tip for the latter and see how suddenly 2000 stops advancing while 2001 starts advancing

runtime/dancebox/src/lib.rs
fgamundi: I have gone ahead and changed the logic a bit. Now it only checks if the tank has enough balance for only 1 session for both blocks and assignment + tip, unless the para has been assigned collators on the current session. If so, it checks that it has enough to pay for two sessions of block production

pallets/collator-assignment/src/lib.rs
```diff
@@ -292,6 +308,23 @@ pub mod pallet {
                 }
             };
 
+            let mut assigned_containers = new_assigned.container_chains.clone();
+            assigned_containers.retain(|_, v| !v.is_empty());
+
+            // On congestion (not enough collators for all paras) prioritized chains need to pay
+            // the minimum tip of the prioritized chains
+            let maybe_tip: Option<BalanceOf<T>> =
+                if assigned_containers.len() < container_chain_ids.len() + parathreads.len() {
```
ParthDesai, line 317: Shouldn't this calculation take into account min collators required per chain? And congestion can happen if sum of min collators required for all chains is greater than assigned containers?

Right now it is set as `collators_per_container` for all chains, so it can simply be `container_chain_ids.len() * collators_per_container + parathreads.len()`

pallets/collator-assignment/src/lib.rs
```diff
@@ -236,6 +243,15 @@ pub mod pallet {
                     max_collators: collators_per_parathread,
                 });
             }
+
+            // Prioritize paras by tip
+            // As of now this doesn't distinguish between parachains and parathreads
+            // TODO apply different logic to parathreads
+            chains.sort_by(|a, b| {
```
ParthDesai, line 250: I think it would be better if we only sort the chains by tip when we detect congestion as tip is only being paid if we detect congestion.

Reason being that, the first chains on the list gets more collators if there are surplus collators as per `select_chains_with_collators`. So, if there is no congestion in the network, the chains which declared the max tip do not pay fee and based on the max tip setting alone, could be given priority in getting surplus collators.  Even for chains with tip declared, the chain with less max tip will be given lower priority than chain with more tip declared. 

Alternatively, you could always charge the tip and give them more collators even in case there is no congestion. That way there is a tangible cost to be ahead in the list.

girazoki: I think we should not charge tip if there is no congestion. So I agree on sorting only if congestion is detected

pallets/collator-assignment/src/lib.rs
girazoki: I agree here, my impression is that first we try to assign, then if we detect that the assignation does not cover all the demand we sort by tip. So I think it's better to do it as @ParthDesai describes, and assign only once sorted

pallets/collator-assignment/src/lib.rs
fgamundi: You're right that it's better to only sort on congestion. Just pushed a new commit with that check

pallets/collator-assignment/src/lib.rs
fgamundi: At this point it doesn't matter the min collators required because that was already takin into account in the assignment function (`assign_collators_rotate_all` and `assign_collators_always_keep_old`). So if the amount of assigned chains is less than the demand then it's because the assignment wasn't able to satisfy the minimum demand.
Nevertheless, I have [changed the congestion condition in the latest commit](https://github.com/moondance-labs/tanssi/pull/475/files#diff-8686f26c396cd2f0d9520d4e109f7f2f163dd0635de8cd2c37c84e4b98e53976R248) and the tip check is using that.

pallets/collator-assignment/src/lib.rs
```diff
@@ -177,13 +179,22 @@ pub mod pallet {
                 .into_iter()
                 .map(|(para_id, _)| para_id)
                 .collect();
+
+            // We read current assigned collators
+            let old_assigned = Self::read_assigned_collators();
+            let old_assigned_para_ids = old_assigned.container_chains.clone().into_keys().collect();
```
tmpolaczyk, line 185: This looks very similar but here you are only cloning the keys, while using `clone` clones both keys and values.

```suggestion
            let old_assigned_para_ids = old_assigned.container_chains.keys().cloned().collect();
```

pallets/collator-assignment/src/lib.rs
```diff
@@ -236,6 +243,24 @@ pub mod pallet {
                     max_collators: collators_per_parathread,
                 });
             }
+
+            // Are there enough collators to satisfy the minimum demand?
+            let congestion =
+                T::HostConfiguration::min_collators_for_orchestrator(target_session_index)
+                    + collators_per_container * container_chain_ids.len() as u32
+                    + collators_per_parathread * parathreads.len() as u32
+                    > collators.len() as u32;
```
tmpolaczyk, line 252: Auditors will be happier if we use `.saturating_mul` and `.saturating_add`. Also the variable name could be improved, something like `enough_collators_for_all_chains` or `more_collators_than_chains` or `need_chain_priority` or `use_chain_tip`.

pallets/collator-assignment/src/lib.rs
```diff
@@ -301,7 +339,19 @@ pub mod pallet {
...
+                        maybe_tip.as_ref(),
+                        false,
+                    ) {
+                        // On error remove para from assignment
+                        log::warn!(
+                            "CollatorAssignmentHook error! Removing para {} from assignment: {:?}",
+                            u32::from(*para_id),
+                            e
+                        );
+                        new_assigned.container_chains.remove(para_id);
```
tmpolaczyk, line 353: Can this branch actually happen or is it unreachable?

fgamundi: In practice it is currently unreachable given dancebox's and flashbox's current `remove_para_ids_with_no_credits` implementation that discards any container that might reach this.
Nevertheless it was suggested for the case to be correctly handled by auditors

tmpolaczyk: So basically `T::CollatorAssignmentHook::on_collators_assigned` should never fail, and we currently ensure that by using a separate trait `remove_para_ids_with_no_credits`.

Maybe a better design would be try have only 1 trait, and first call it using `transactional` magic so that the changes to storage are not applied, and then after the assignment is calculated we call it same as we do now. This would decrease the amount of code we have to maintain, because I guess now `remove_para_ids_with_no_credits` is just a duplicate. But this is out of scope for this PR.

And I agree with the auditors that we should handle the error case here.

primitives/traits/src/lib.rs
```diff
@@ -172,7 +192,7 @@ pub trait RemoveInvulnerables<AccountId> {
 pub trait RemoveParaIdsWithNoCredits {
     /// Remove para ids with not enough credits. The resulting order will affect priority: the first para id in the list
     /// will be the first one to get collators.
-    fn remove_para_ids_with_no_credits(para_ids: &mut Vec<ParaId>);
+    fn remove_para_ids_with_no_credits(para_ids: &mut Vec<ParaId>, old_assigned: &Vec<ParaId>);
```
tmpolaczyk, line 195: This should make `.contains()` a bit faster when we have many para ids

```suggestion
    fn remove_para_ids_with_no_credits(para_ids: &mut Vec<ParaId>, old_assigned: &BTreeSet<ParaId>);
```

pallets/collator-assignment/src/lib.rs
fgamundi: It's still not quite a duplicate because it also checks that the tank has enough to pay for block production.
I've created a ticket to look further into this

girazoki (approved):
LGTM now @fgamundi good job! Wait for one more approval from any of the other reviews though please

pallets/collator-assignment/src/lib.rs
```diff
@@ -236,6 +244,27 @@ pub mod pallet {
                     max_collators: collators_per_parathread,
                 });
             }
+
+            // Are there enough collators to satisfy the minimum demand?
+            let enough_collators_for_all_chain = collators.len() as u32
+                > T::HostConfiguration::min_collators_for_orchestrator(target_session_index)
```
tmpolaczyk, line 250:
```suggestion
                >= T::HostConfiguration::min_collators_for_orchestrator(target_session_index)
```

If the number of collators is equal to the required we also don't want to apply the congestion fee. Not sure if this was correct before the last refactor.

fgamundi: Good catch!

pallets/services-payment/src/lib.rs
```diff
@@ -503,28 +528,71 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
...
+            T::OnChargeForCollatorAssignment::on_unbalanced(assignment_imbalance);
+        }
+
+        Ok(T::WeightInfo::on_collators_assigned())
+    }
+}
+
+impl<T: Config> CollatorAssignmentTip<BalanceOf<T>> for Pallet<T> {
+    fn get_para_tip(para_id: ParaId) -> Option<BalanceOf<T>> {
+        MaxTip::<T>::get(para_id)
```
tmpolaczyk, line 595: In this file the `para_deregistered` hook should clean all the storage items. It's missing `MaxTip` and also `MaxCorePrice` because I forgot to add it

fgamundi: Added both

test/suites/common-tanssi/services-payment/test_services_payment_collator_tip.ts
```diff
@@ -0,0 +1,48 @@
...
+                await context.createBlock([await tx.signAsync(alice)]);
+
+                const txMaxTip = polkadotJs.tx.servicesPayment.setMaxTip(paraId, tip);
+                await context.createBlock([await polkadotJs.tx.sudo.sudo(txMaxTip).signAsync(alice)]);
+                await jumpSessions(context, 2);
+
+                const collators = await polkadotJs.query.collatorAssignment.collatorContainerChain();
+                expect(
+                    collators.toJSON().containerChains[paraId].length,
+                    `Container chain ${paraId} should have 0 collators`
```
tmpolaczyk, line 39:
```suggestion
                    `Container chain ${paraId} should have 2 collators`
```

test/suites/common-tanssi/services-payment/test_services_payment_block_credit_buying_free_combined.ts
```diff
@@ -11,7 +11,7 @@ describeSuite({
     testCases: ({ it, context }) => {
         let polkadotJs: ApiPromise;
         let alice: KeyringPair;
-        const blocksPerSession = 10n;
+        const blocksPerSession = 5n;
```
tmpolaczyk, line 14: We have `blocksPerSession = 10` because of the 6s block time, right? If this test fails we should divide by 2 where needed, but not here.

tmpolaczyk (approved):
Good job! I left some comments for the typescript tests but aside from that it looks good!

tmpolaczyk:
It seems CI will fail until you merge master

PR #476, opened 2024-03-26
Allow parathread manager to set max core price

tmpolaczyk:
Allows the parathread manager to specify the max price for an on-demand core, in units of relay chain currency. The default value is unlimited. This is the price limit, not the price, the actual price depends on usage but that's calculated in the relay chain.


pallets/services-payment/src/lib.rs
```diff
@@ -249,6 +257,31 @@ pub mod pallet {
 
             Ok(().into())
         }
+
+        /// Max core price for parathread in relay chain currency
+        #[pallet::call_index(5)]
+        // TODO: weight
+        #[pallet::weight(T::WeightInfo::set_refund_address())]
```
girazoki, line 264: UPdate bench

runtime/dancebox/src/xcm_config.rs
```diff
@@ -570,3 +571,11 @@ impl GetPurchaseCoreCall<RelayChain> for EncodedCallToBuyCore {
         }
     }
 }
+
+pub struct GetMaxCorePriceFromServicesPayment;
+
+impl Convert<ParaId, Option<u128>> for GetMaxCorePriceFromServicesPayment {
+    fn convert(para_id: ParaId) -> Option<u128> {
```
girazoki, line 578: Let's use a different trait to convert

fgamundi (approved):

pallets/services-payment/src/lib.rs
```diff
@@ -249,6 +257,30 @@ pub mod pallet {
...
+
+        /// Max core price for parathread in relay chain currency
+        #[pallet::call_index(5)]
+        #[pallet::weight(T::WeightInfo::set_max_core_price())]
+        pub fn set_max_core_price(
+            origin: OriginFor<T>,
+            para_id: ParaId,
+            max_core_price: Option<u128>,
+        ) -> DispatchResultWithPostInfo {
+            T::SetRefundAddressOrigin::ensure_origin(origin, &para_id)?;
```
fgamundi, line 269: Might be good to rename `SetRefundAddressOrigin` to `ManagerOrigin` or similar if it's going to be used in multiple calls

girazoki: Yeah I would do that

girazoki (commented):
I would just add an integration test if possible (in the xcm-core-buyer with theh emulator)

girazoki (approved):

PR #477, opened 2024-03-26
Fix use of current max_collators instead of pending

tmpolaczyk:
In `pallet_collator_assignment`, we use the pending config values to ensure that the current assignment always matches the current configuration state. So, if we change the value of `collators_per_container`, when the change is reflected in `configuration.config`, we should see that the number of collators per container in `pallet_collator_assignment` has also changed.

However we were not doing that with the value of `max_collators`, we were using the current value instead of the pending one. This PR fixes that.


girazoki (approved):
Lookgs good, thank you @tmpolaczyk

PR #478, opened 2024-03-27
Add dummy precompile hook with rust and typescript tests

ParthDesai:
## Description 
Follow up PR for https://github.com/Moonsong-Labs/moonkit/pull/32 

This PR utilizes the hooks introduced in above PR and adds revert bytecode at the evm account derived from the asset id specified in the hook invocation.

### DIFF
--- a/container-chains/templates/frontier/runtime/src/xcm_config.rs
+++ b/container-chains/templates/frontier/runtime/src/xcm_config.rs
@@ -14,6 +14,9 @@
 // You should have received a copy of the GNU General Public License
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
+use pallet_foreign_asset_creator::{
+    AssetBalance, AssetId as AssetIdOf, ForeignAssetCreatedHook, ForeignAssetDestroyedHook,
+};
 use {
     super::{
         precompiles::FOREIGN_ASSET_PRECOMPILE_ADDRESS_PREFIX, AccountId, AllPalletsWithSystem,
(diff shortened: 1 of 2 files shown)


test/suites/dev-frontier-template/test-eth-asset-address/test-eth-asset-address-creation.ts
```diff
@@ -0,0 +1,51 @@
...
+import { alith } from "@moonwall/util";
+
+describeSuite({
+    id: "DF0901",
+    title: "Ethereum asset dummy precompile address creation",
+    foundationMethods: "dev",
+    testCases: ({ context, it }) => {
+        it({
+            id: "T01",
+            title: "dummy precompile address is created when",
```
girazoki, line 12:
```suggestion
            title: "dummy precompile address is created when creating the asset and removed when destroyed",
```

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -376,6 +379,33 @@ impl pallet_assets::Config<ForeignAssetsInstance> for Runtime {
     type BenchmarkHelper = ForeignAssetBenchmarkHelper;
 }
 
+pub struct DummyPrecompileHook;
```
girazoki, line 382:
```suggestion
pub struct RevertCodePrecompileHook;
```

girazoki (approved):
Small comments, but otherwise looks good!

nanocryk (approved):

container-chains/templates/frontier/runtime/src/xcm_config.rs
```diff
@@ -439,3 +469,17 @@ pub type ForeignFungiblesTransactor = FungiblesAdapter<
 /// Multiplier used for dedicated `TakeFirstAssetTrader` with `ForeignAssets` instance.
 pub type AssetRateAsMultiplier =
     AssetFeeAsExistentialDepositMultiplier<Runtime, WeightToFee, AssetRate, ForeignAssetsInstance>;
+
+#[test]
+fn test_asset_id_to_account_conversion() {
```
tmpolaczyk, line 474: I think we don't need this test

ParthDesai: Currently `asset_id_to_account` takes `prefix` as slice, instead of hard bound array. Then it does invoke `copy_from_slice`, which can panic, if the slice's size is not exact. 

If the implementation changes tomorrow (i.e the expected size of the slice in the implementation), it need to be captured by the unit test (otherwise you would see panic during integration test or worse in dev environment). Which this test prevents.

ParthDesai: Let's discuss this offline. I am merging this as it is not blocker as of now.

PR #479, opened 2024-03-27
Add delivery fees

girazoki:
Add delivery fees for XCM routers in all runtimes (except Flashbox). We use `ExpontentialPrice` from polkadot-runtime-common and we charge in the native asset for each runtime.

Added a couple function in TS so that we can check that indeed the messages are correctly charged. These take the latest fees from the sent messages available in storage


runtime/dancebox/src/xcm_config.rs
```diff
@@ -245,14 +249,20 @@ impl pallet_xcm::Config for Runtime {
     type AdminOrigin = EnsureRoot<AccountId>;
 }
 
+pub type PriceForSiblingParachainDelivery =
+    ExponentialPrice<SelfReserve, BaseDeliveryFee, TransactionByteFee, XcmpQueue>;
+
+pub type PriceForParentDelivery =
+    ExponentialPrice<FeeAssetId, BaseDeliveryFee, TransactionByteFee, ParachainSystem>;
```
Agusrodri, line 256: Why do we need the conversion to `FeeAssetId` type? Couldn't we use the same `SelfReserve` as in `PriceForSiblingParachainDelivery`?

girazoki: yep I think so

test/util/xcm.ts
```diff
@@ -919,3 +919,43 @@ export const expectXcmEventMessage = async (context: DevModeContext, message: st
...
+    const records = await context.polkadotJs().query.system.events();
+
+    const filteredEvents = records
+        .map(({ event }) => (context.polkadotJs().events.polkadotXcm.FeesPaid.is(event) ? event : undefined))
+        .filter((event) => event);
+
+    return filteredEvents[0]!.data[1][0].fun.asFungible.toBigInt();
+};
+
+export const getLastSentUmpMessageFee = async (context: DevModeContext, baseDelivery: bigint, txByteFee: bigint) => {
```
Agusrodri, line 933: Really useful functions :)

Agusrodri (approved):
Looks great! Specially the typescript helpers/tests addition 🥇

fgamundi (approved):

test/util/xcm.ts
```diff
@@ -919,3 +919,43 @@ export const expectXcmEventMessage = async (context: DevModeContext, message: st
...
+    const sentXcm = (await context.polkadotJs().query.parachainSystem.upwardMessages())[0];
```
ParthDesai, line 934: Could this be a runtime error if the message list is empty?

ParthDesai (approved):
Minor comment in test. Apart from that everything else look good.

test/util/xcm.ts
girazoki: If the message list is empty it will simply throw a typescript error.

girazoki: the runtime will return an empty vec and you will try to get the item 0, which does not exist

ParthDesai: yes, I mean typescript error. Do we safeguard against that? Or it is fine?

PR #481, opened 2024-04-03
Use cancellation token and exit notifications instead of watch

ParthDesai:
## Description
This PR changes `call_collate_on` to use cancellation token and exit notification to stop it gracefully as well as wait for it to exit.

### DIFF
--- a/client/consensus/src/collators/lookahead.rs
+++ b/client/consensus/src/collators/lookahead.rs
@@ -144,8 +144,6 @@ where
     // rules specified by the parachain's runtime and thus will never be too deep.
     const PARENT_SEARCH_DEPTH: usize = 10;
 
-    log::info!("LOOKAHEAD COLLATOR RUNNING...");
-
     let (exit_notification_sender, exit_notification_receiver) = oneshot::channel();
 
     let aura_fut = async move {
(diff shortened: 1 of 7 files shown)


girazoki (approved):
It looks to me but I would rather have @tmpolaczyk take a look before merging

client/consensus/src/collators/lookahead.rs
```diff
@@ -178,262 +184,256 @@ where
...
-                Ok(None) => continue,
-                Ok(Some(pvd)) => pvd.max_pov_size,
-                Err(err) => {
-                    tracing::error!(target: crate::LOG_TARGET, ?err, "Failed to gather information from relay-client");
-                    continue;
-                }
-            };
+        loop {
+            select! {
+                Some(relay_parent_header) = import_notifications.next() => {
```
tmpolaczyk, line 189: The previous code exited the loop if `import_notifications.next()` returned `None`, what will happen now?

tmpolaczyk (commented):
Some comments but overall it looks better than what we had before

client/consensus/src/collators/lookahead.rs
```diff
@@ -142,7 +146,9 @@ where
 
     log::info!("LOOKAHEAD COLLATOR RUNNING...");
```
tmpolaczyk, line 147:
```suggestion
```

This was probably meant for debugging

tmpolaczyk (approved):

node/src/container_chain_spawner.rs
```diff
@@ -135,8 +139,21 @@ impl ContainerChainSpawner {
...
+            self.chain_type.clone(),
+            self.relay_chain.clone(),
+            self.relay_chain_interface.clone(),
+            self.collator_key.clone(),
+            self.sync_keystore.clone(),
+            self.orchestrator_para_id,
+            self.validator,
+            self.spawn_handle.clone(),
+            self.state.clone(),
+        );
```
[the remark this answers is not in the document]
ParthDesai: I tried to find more idiomatic way than that and at last left it as it is. Will open an issue and improve this as part of next PR.

PR #483, opened 2024-04-04
Fix calculation of Auto compounding delegation of candidate

ParthDesai:
## Description
Currently, we are calculating delegation amount of candidate (referred to as delegator in this scenario) in the auto compounding pool by simply querying `AutoCompoundingSharesHeldStake` which is not updated against rewards earned by that delegator as the rewards simply increases value of each shares. 

This PR attempts to correct this by calculating delegation of delegator in auto compounding pool by:
 ```
(shares of delegator in auto pool * (total staked amount in auto pool / total supply of shares in auto pool);
```

### DIFF
--- a/test/suites/smoke-test-dancebox/test-staking-consistency.ts
+++ b/test/suites/smoke-test-dancebox/test-staking-consistency.ts
@@ -49,13 +49,31 @@ describeSuite({
                         })
                     ).toBigInt();
 
-                    const auto = (
+                    const autoCompoundingSharesTotalStaked = (
                         await api.query.pooledStaking.pools(candidate, {
-                            AutoCompoundingSharesHeldStake: {
+                            AutoCompoundingSharesTotalStaked: {},
+                        })
+                    ).toBigInt();
+
+                    const autoCompoundingSharesSupply = (
+                        await api.query.pooledStaking.pools(candidate, {
+                            AutoCompoundingSharesSupply: {},
+                        })
+                    ).toBigInt();
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by girazoki — APPROVED** [review]
Looks good to me, @nanocryk this should fix the smoke tests. @ParthDesai did you check whether the alert gets cleared?

**Comment by ParthDesai** [comment]
> Looks good to me, @nanocryk this should fix the smoke tests. @ParthDesai did you check whether the alert gets cleared?

Yes. I checked in my local, without this fix the test was failing, with this fix the test started passing.

File: test/suites/smoke-test-dancebox/test-staking-consistency.ts [hunk 1]
```diff
@@ -49,14 +49,32 @@ describeSuite({
...
+                    const autoCompoundingSharesOfCandidate = (
+                        await api.query.pooledStaking.pools(candidate, {
+                            AutoCompoundingShares: {
                                 delegator: candidate,
                             },
                         })
                     ).toBigInt();
 
+                    // auto stake is calculated using this method as the AutoCompoundingSharesHeldStake is not updated with rewards received
+                    // by the candidate, rather the number of shares of candidate increases.
```
nanocryk, line 73:
```suggestion
                    // auto stake is calculated using this method as the AutoCompoundingSharesHeldStake is not updated with rewards received
                    // by the candidate, rather the value of each share of candidate increases.
```

nanocryk (approved):

PR #492, opened 2024-04-10
Stop orchestrator collator as soon as the node is unassigned

tmpolaczyk:
Fixes a bug where the collator process for the orchestrator chain will always keep running in the background, trying to produce blocks but not being able to because it is not eligible.

Consider a collator that goes from orchestrator to a container and then back, like 1000 -> 2001 -> 1000.
Before the PR, the collator process for 1000 was always running, being restarted during the transition to 2001 -> 1000.
After the PR, the collator process for 1000 is stopped during the first transition 1000 -> 2001, and then started again in the transition 2001 -> 1000.

Also includes a refactor on how the orchestrator collator is started, which shouldn't change the behavior but allows to remove some TODOs from unit tests, and a cleanup of a zombienet test that was always unreliable, which should reduce the number of false positive CI errors related to timeouts. But this is not related to the core issue of this PR.

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -81,6 +81,8 @@ pub struct ContainerChainSpawner {
     // Async callback that enables collation on the orchestrator chain
     pub collate_on_tanssi:
         Arc<dyn Fn() -> (CancellationToken, futures::channel::oneshot::Receiver<()>) + Send + Sync>,
+    // Stores the cancellation token used to stop the orchestrator chain collator process.
+    // When this is None, the orchestrator collator is not running.
     pub collation_cancellation_constructs:
         Option<(CancellationToken, futures::channel::oneshot::Receiver<()>)>,
 }
@@ -538,7 +540,8 @@ fn handle_update_assignment_state_change(
 
     running_chains_before.extend(state.assigned_para_id);
     running_chains_before.extend(state.next_assigned_para_id);
-    // Ignore orchestrator_para_id because it cannot be stopped or started, it is always running
+    // Ignore orchestrator_para_id because it is handled in a special way, as it does not need to
+    // start one session before in order to sync.
     running_chains_before.remove(&orchestrator_para_id);
 
     running_chains_after.extend(current);
@@ -548,13 +551,10 @@ fn handle_update_assignment_state_change(
     let mut need_to_restart_next = false;
 
     if state.assigned_para_id != current {
-        // If the assigned container chain was already running but not collating, we need to call collate_on
         if let Some(para_id) = current {
-            // Check if we get assigned to orchestrator chain
-            if para_id == orchestrator_para_id {
-            } else {
-                // When we get assigned to a different container chain, we don't need to call collate_on because
-                // we will restart that container chain in collation mode.
+            // If the assigned container chain has changed, we may need to
+            // restart it in collation mode, unless it is the orchestrator chain.
+            if para_id != orchestrator_para_id {
                 need_to_restart_current = true;
             }
         }
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -428,9 +428,16 @@ impl ContainerChainSpawner {
     }
 
     /// Receive and process `CcSpawnMsg`s indefinitely
-    pub async fn rx_loop(mut self, mut rx: mpsc::UnboundedReceiver<CcSpawnMsg>) {
-        // The node always starts as an orchestrator chain collator
-        self.state.lock().unwrap().assigned_para_id = Some(self.orchestrator_para_id);
+    pub async fn rx_loop(mut self, mut rx: mpsc::UnboundedReceiver<CcSpawnMsg>, validator: bool) {
+        // The node always starts as an orchestrator chain collator.
+        // This is because the assignment is detected after importing a new block, so if all
+        // collators stop at the same time, when they start again nobody will produce the new block.
+        // So all nodes start as orchestrator chain collators, until the first block is imported,
+        // then the real assignment is used.
+        if validator {
+            self.handle_update_assignment(Some(self.orchestrator_para_id), None)
+                .await;
+        }
 
         while let Some(msg) = rx.recv().await {
             match msg {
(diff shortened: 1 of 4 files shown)


test/suites/para/test_tanssi_containers.ts
```diff
@@ -341,35 +344,88 @@ describeSuite({
...
-                // TODO: this passes if only 2 authors are creating blocks, think a way to test that case
-                await countUniqueBlockAuthors(paraApi, blockNumber, blockNumber2002Start - 1, 4);
-
-                expect(sessionPeriod * 5 < blockNumber2002End, "2002 should have deregistered after first rotation");
-                expect(sessionPeriod * 10 > blockNumber2002End, "2002 should have deregistered before second rotation");
-
-                // While 2002 is live: 2 authors (the other 2 went to container chain 2002)
-                // We take from the first block that rotates, otherwise rotation kicks in
-                await countUniqueBlockAuthors(paraApi, sessionPeriod * 10, blockNumber2002End - 1, 2);
+                const fullRotationBlock = 50;
```
girazoki, line 355: I think we can do this in a more parametric manner. Meaning that we can always do `blockNumber2002Start%blocksPerSesssion == 0`

node/src/container_chain_spawner.rs
```diff
@@ -428,7 +428,17 @@ impl ContainerChainSpawner {
     }
 
     /// Receive and process `CcSpawnMsg`s indefinitely
-    pub async fn rx_loop(mut self, mut rx: mpsc::UnboundedReceiver<CcSpawnMsg>) {
+    pub async fn rx_loop(mut self, mut rx: mpsc::UnboundedReceiver<CcSpawnMsg>, validator: bool) {
```
girazoki, line 431: Is this task even necessary to be spawned if validator==false?

node/src/container_chain_spawner.rs
```diff
@@ -894,20 +891,30 @@ mod tests {
...
                 &mut self.state.lock().unwrap(),
                 self.orchestrator_para_id,
-                self.collate_on_tanssi.clone(),
                 current,
                 next,
             );
 
+            if current != Some(self.orchestrator_para_id) {
+                // If not assigned to orchestrator chain anymore, we need to stop the collator process
+                let mut cco = self.currently_collating_on.lock().unwrap();
```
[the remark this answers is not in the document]
ParthDesai: Yes. If possible avoid unwrap or expect, if not use expect with the reason saying why it won't fail.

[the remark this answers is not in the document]
girazoki: nevermind this is a test, I did not realize

node/src/container_chain_spawner.rs
girazoki: let's consider in the future not spawning this rx-loop if validator == false

test/suites/para/test_tanssi_containers.ts
tmpolaczyk: What do you mean?

node/src/container_chain_spawner.rs
```diff
@@ -429,6 +429,9 @@ impl ContainerChainSpawner {
 
     /// Receive and process `CcSpawnMsg`s indefinitely
     pub async fn rx_loop(mut self, mut rx: mpsc::UnboundedReceiver<CcSpawnMsg>) {
+        // The node always starts as an orchestrator chain collator
+        self.state.lock().unwrap().assigned_para_id = Some(self.orchestrator_para_id);
```
ParthDesai: If possible, avoid unwrap/expect. If not use expect instead of unwrap.

node/src/container_chain_spawner.rs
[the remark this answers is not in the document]
ParthDesai: For test it is okay.

test/suites/para/test_tanssi_containers.ts
girazoki: here you are comparing your tests cases against a hardcoded block. But maybe it would be good to compare it against the next full rotation block (regadless if its 50 or 100)

girazoki: it's not very important as 99.5% of the times this block willl be 50, but if it's easy to change it maybe it's worth doing

girazoki (approved):
I have only minor comments that should not hold back merging this PR. Good job @tmpolaczyk

test/suites/para/test_tanssi_containers.ts
tmpolaczyk: I'm a bit scared but will see if it can be done

node/src/container_chain_spawner.rs
```diff
@@ -472,8 +481,9 @@ impl ContainerChainSpawner {
             if let Some(exit_notification_receiver) = maybe_exit_notification_receiver {
                 let _ = exit_notification_receiver.await;
             }
-
-            self.collation_cancellation_constructs = Some(f());
+        } else if self.collation_cancellation_constructs.is_none() {
```
ParthDesai, line 484: If you are going to use `collation_cancellation_constructs` to determine if collator process is not running, maybe we should rename it to indicate that.

node/src/container_chain_spawner.rs
```diff
@@ -550,7 +552,6 @@ fn handle_update_assignment_state_change(
         if let Some(para_id) = current {
             // Check if we get assigned to orchestrator chain
             if para_id == orchestrator_para_id {
-                call_collate_on = Some(collate_on_tanssi);
```
ParthDesai, line 553: So, is this just empty if?

tmpolaczyk: Good catch haha

node/src/container_chain_spawner.rs
tmpolaczyk: I can't think of a good name, but I added docs to the field definition.

node/src/container_chain_spawner.rs
tmpolaczyk: Also I believe the logic of `need_to_restart_current` and `need_to_restart_next` can be simplified, but I didn't want to touch it in case it breaks something.

test/suites/para/test_tanssi_containers.ts
girazoki: no worries, thanks for trying!

node/src/container_chain_spawner.rs
girazoki: let's leave it as a future improvement yes

PR #493, opened 2024-04-12
Stream payment deposit on opening a stream

fgamundi:
Adds a balance hold (to the origin) when opening a `pallet_stream_payment` Stream to prevent possible storage bloat.

### DIFF
--- a/pallets/stream-payment/src/lib.rs
+++ b/pallets/stream-payment/src/lib.rs
@@ -155,6 +155,8 @@ pub mod pallet {
         /// Provide interaction with assets.
         type Assets: Assets<Self::AccountId, Self::AssetId, Self::Balance>;
 
+        /// Currency for the opening balance hold for the storage used by the Stream.
+        /// NOT to be confused with Assets.
         type Currency: Inspect<Self::AccountId, Balance = Self::Balance>
             + MutateHold<Self::AccountId, Reason = Self::RuntimeHoldReason>;
 
@@ -189,8 +191,6 @@ pub mod pallet {
         pub target: AccountId,
         /// Steam config (time unit, asset id, rate)
         pub config: StreamConfig<Unit, AssetId, Balance>,
-        /// One-time opening deposit. Will be released on close.
-        pub opening_deposit: Balance,
         /// How much is deposited to fund this stream.
         pub deposit: Balance,
         /// Last time the stream was updated in `config.time_unit`.
@@ -201,6 +201,8 @@ pub mod pallet {
         pub request_nonce: RequestNonce,
         /// A pending change request if any.
         pub pending_request: Option<ChangeRequest<Unit, AssetId, Balance>>,
+        /// One-time opening deposit. Will be released on close.
+        pub opening_deposit: Balance,
     }
 
     impl<AccountId: PartialEq, Unit, AssetId, Balance> Stream<AccountId, Unit, AssetId, Balance> {
@@ -416,8 +418,15 @@ pub mod pallet {
             initial_deposit: T::Balance,
         ) -> DispatchResultWithPostInfo {
             let origin = ensure_signed(origin)?;
+            let opening_deposit = T::OpenStreamHoldAmount::get();
 
-            let _stream_id = Self::open_stream_returns_id(origin, target, config, initial_deposit)?;
+            let _stream_id = Self::open_stream_returns_id(
+                origin,
+                target,
+                config,
+                initial_deposit,
+                opening_deposit,
+            )?;
 
             Ok(().into())
         }
@@ -758,6 +767,7 @@ pub mod pallet {
             target: AccountIdOf<T>,
             config: StreamConfigOf<T>,
             initial_deposit: T::Balance,
+            opening_deposit: T::Balance,
         ) -> Result<T::StreamId, DispatchErrorWithPostInfo> {
             ensure!(origin != target, Error::<T>::CantBeBothSourceAndTarget);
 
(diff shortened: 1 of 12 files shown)


runtime/dancebox/src/lib.rs
```diff
@@ -1530,13 +1530,20 @@ impl pallet_stream_payment::TimeProvider<TimeUnit, Balance> for TimeProvider {
 
 type StreamId = u64;
 
+parameter_types! {
+    pub const OpenStreamHoldAmount: Balance = 100 * UNIT;
```
girazoki, line 1534: This is way to much haha.

girazoki: We probably should calculate how much size it takes to hold a deposit in storage. I think there is something similar done in pallet-assets

pallets/stream-payment/src/lib.rs
```diff
@@ -738,6 +777,11 @@ pub mod pallet {
                 .ok_or(Error::<T>::StreamIdOverflow)?;
             NextStreamId::<T>::set(next_stream_id);
 
+            // TODO save the holded amount to storage
+            let hold_amount = T::OpenStreamHoldAmount::get();
+            T::Currency::hold(&HoldReason::StreamOpened.into(), &origin, hold_amount)?;
```
girazoki, line 782: @nanocryk do we envision cases in which we dont apply this hold? (Maybe if the stream-opening comes from data-preservers, which already will do a hold per se

pallets/stream-payment/src/lib.rs
```diff
@@ -425,11 +455,20 @@ pub mod pallet {
             // Unfreeze funds left in the stream.
             T::Assets::decrease_deposit(&stream.config.asset_id, &stream.source, stream.deposit)?;
 
+            let hold_amount = StreamsHolds::<T>::take(stream_id);
```
girazoki, line 458: what happens if the hold does not exist?

pallets/stream-payment/src/lib.rs
```diff
@@ -289,6 +309,15 @@ pub mod pallet {
         QueryKind = OptionQuery,
     >;
 
+    /// Store each hold made to
+    #[pallet::storage]
+    pub type StreamsHolds<T: Config> = StorageMap<
```
tmpolaczyk, line 314: Why not make the hold a field of `pub struct Stream`? Then it will be impossible to create a stream without explicitly setting the hold, and we save 1 write when opening and closing a stream.

pallets/stream-payment/src/lib.rs
fgamundi: Yes, it returns 0. Sounds good

pallets/stream-payment/src/lib.rs
fgamundi: Yeah, I though of that also, but wanted to first have a working poc. If I add it to `Stream` that would also require a migration, right?

runtime/dancebox/src/lib.rs
fgamundi: You mean [`currency::deposit`](https://github.com/moondance-labs/tanssi/blob/master/runtime/dancebox/src/lib.rs#L160)? Like it's used for [pallet_identity](https://github.com/moondance-labs/tanssi/blob/master/runtime/dancebox/src/lib.rs#L1546)

pallets/stream-payment/src/lib.rs
girazoki: yeah but I am not a big fan of trusting no-one will use it till then haha

girazoki: but I guess it's ok yes

pallets/stream-payment/src/lib.rs
```diff
@@ -145,6 +155,13 @@ pub mod pallet {
         /// Provide interaction with assets.
         type Assets: Assets<Self::AccountId, Self::AssetId, Self::Balance>;
 
+        type Currency: Inspect<Self::AccountId, Balance = Self::Balance>
```
nanocryk, line 158: Add comment that explain this is only for the deposit and should not be confused with Assets, which are assets the streams can pay in.

runtime/flashbox/src/lib.rs
```diff
@@ -1284,13 +1284,21 @@ impl pallet_stream_payment::TimeProvider<TimeUnit, Balance> for TimeProvider {
 
 type StreamId = u64;
 
+parameter_types! {
+    // 1 entry, storing 173 bytes on-chain
+    pub const OpenStreamHoldAmount: Balance = currency::deposit(1, 173);
```
tmpolaczyk, line 1289: Just curious, how did you get this value?

pallets/stream-payment/src/lib.rs
fgamundi: I've changed the `open_stream_returns_id` function to accept an `opening_deposit` https://github.com/moondance-labs/tanssi/pull/493/files#diff-6a7c7b4db06d7b82cddd52c77c149e7ee2e502bdc12e7291b9f3743349ec377fR763.
When it's called by the extrinsic this is set to the value in the runtime, and it can also be called by a future trait to be used by other pallets to set their own deposit if needed.
Would this suffice?

runtime/flashbox/src/lib.rs
fgamundi: Did a raw query of a Stream that had a `pending_request` and used that size

pallets/stream-payment/src/lib.rs
girazoki: yes, thanks!

pallets/stream-payment/src/lib.rs
```diff
@@ -145,6 +148,15 @@ pub mod pallet {
...
         type Assets: Assets<Self::AccountId, Self::AssetId, Self::Balance>;
 
+        /// Currency for the opening balance hold for the storage used by the Stream.
+        /// NOT to be confused with Assets.
+        type Currency: Inspect<Self::AccountId, Balance = Self::Balance>
+            + MutateHold<Self::AccountId, Reason = Self::RuntimeHoldReason>;
+
+        type RuntimeHoldReason: From<HoldReason>;
+
+        type OpenStreamHoldAmount: Get<Self::Balance>;
```
girazoki, line 158:
```suggestion
        #[constant]
        type OpenStreamHoldAmount: Get<Self::Balance>;
```

girazoki (approved):
good job frnacisco! one more additional comment from my side

tmpolaczyk (approved):

PR #494, opened 2024-04-12
Do not panic when container chain fails to start

tmpolaczyk:
Allows collators to keep running if there is any error when trying to start a container chain, or if the container chain has an essential tasks that stops unexpectedly. Note that any panics will still stop the node, regardless if they happen in the container chain task manager or in the orchestrator/relay. The only exception is panics that happen inside the runtime, those will never cause the node to stop.

Can be tested by reverting the commit 7af640409c82df81ab53b7baa0dda4232e2f0fcf and running `pnpm moonwall test zombie_tanssi`, the test will fail but inspecting the logs will show that the nodes did not crash.


girazoki:
let's try to make a separate PR @tmpolaczyk with a proper testing framework for the CC spawner

girazoki (approved):
LGTM

PR #496, opened 2024-04-12
Fix zombienet test by undoing a change

tmpolaczyk:
It can sometimes fail when we expect 4 different authors in 10 blocks, and we only see 3 authors. That's because the 4th author has just rotated from a container chain, and has not had enough time to propose a block for the orchestrator chain.

This test was broken by #492, I was too optimistic.


girazoki (approved):
so many edge cases...thank you @tmpolaczyk

PR #497, opened 2024-04-15
Improve more timeouts in zombienet tests

tmpolaczyk:
Increase some test timeouts.

With the change from 12 second block time to 6 second, some timeouts were not increased. Previously we waited for 3 tanssi block to assume finality, but finality depends on the relay chain, so we should have been waiting for 6 relay blocks. With this change we wait for 6 tanssi blocks, which should be the same.

Also I merged 2 batches of transactions into 1 batch, because using `signAndSendAndInclude` twice in the same test can make it time out depending on finality.


girazoki (approved):
looks good, thank you @tmpolaczyk

PR #500, opened 2024-04-16
Store tanssi slot number in author noting pallet

tmpolaczyk:
We want this slot number to be the tanssi slot number, not the container chain slot number. Doesn't need a migration because this is not yet deployed.

### DIFF
--- a/pallets/author-noting/src/lib.rs
+++ b/pallets/author-noting/src/lib.rs
@@ -40,6 +40,7 @@ use {
     dp_core::well_known_keys::PARAS_HEADS_INDEX,
     frame_support::{dispatch::PostDispatchInfo, pallet_prelude::*, Hashable},
     frame_system::pallet_prelude::*,
+    nimbus_primitives::SlotBeacon,
     parity_scale_codec::{Decode, Encode},
     sp_consensus_aura::{inherents::InherentType, Slot, AURA_ENGINE_ID},
     sp_inherents::{InherentIdentifier, IsFatalError},
@@ -76,6 +77,7 @@ pub mod pallet {
         type ContainerChains: GetCurrentContainerChains;
 
         type SelfParaId: Get<ParaId>;
+        type SlotBeacon: SlotBeacon;
 
         type ContainerChainAuthor: GetContainerChainAuthor<Self::AccountId>;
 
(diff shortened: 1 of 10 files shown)


pallets/author-noting/src/lib.rs
```diff
@@ -388,7 +388,9 @@ impl<T: Config> Pallet<T> {
             Ok(ContainerChainBlockInfo {
                 block_number: author_header.number,
                 author,
-                latest_slot_number: slot,
+                // TODO: this needs to be the tanssi slot, not the container chain slot,
+                // unimplemented for now, will always be 0
+                latest_slot_number: 0.into(),
```
girazoki, line 393: how difficult is it to put already the tanssi-slot?

tmpolaczyk: Done in next commit

runtime/dancebox/tests/integration_test.rs
```diff
@@ -1791,7 +1791,7 @@ fn test_author_noting_not_self_para() {
                 Some(ContainerChainBlockInfo {
                     block_number: 1,
                     author: AccountId::from(DAVE),
-                    latest_slot_number: 5.into(),
+                    latest_slot_number: 0.into(),
```
girazoki, line 1794: we dont advance the slots in the integration tests?

tmpolaczyk: I believe it's because we are in block 0, let me check what happens if I set the block number to 1.

girazoki: I see there are some tests in which we have a 1. I think it is fair.

girazoki (approved):

runtime/dancebox/tests/integration_test.rs
tmpolaczyk: After `run_to_block(10)`, the slot number is 9, I expected 10 but I guess it doesn't matter?

tmpolaczyk: I guess it's a problem with the way we do integration tests, in zombienet the slot number stored in author noting at block 25 matches the slot found in the header of block 25.

PR #502, opened 2024-04-18
increase rust toolchain

girazoki:


tmpolaczyk:
If we wait 1 week more we can upgrade directly to 1.78 :)

girazoki:
> If we wait 1 week more we can upgrade directly to 1.78 :)

One more week it is :)

girazoki:
> > If we wait 1 week more we can upgrade directly to 1.78 :)
> 
> One more week it is :)

@tmpolaczyk I updated it to 1.78 :)

pallets/invulnerables/src/benchmarking.rs
```diff
@@ -112,6 +112,7 @@ pub(crate) fn currency_issue<T: Config + frame_system::Config>(
     <<T as crate::Config>::Currency as Balanced<T::AccountId>>::issue(amount)
 }
 
+#[allow(clippy::multiple_bound_locations)]
```
girazoki, line 115: For this one, I guess it's part of the macro and we cannot do much

tmpolaczyk: Yeah, it's the one downside of upgrading Rust before upgrading polkadot-sdk, we see clippy warnings from code generated by macros...

tmpolaczyk (approved):

PR #508, opened 2024-04-18
Fix actions

girazoki:
I think these are the main ones to be fixed


.github/workflows/client-release-issue.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.2
       - name: Upload tools
-        uses: actions/upload-artifact@v3
```
ParthDesai, line 18: Why `v4` for client release and runtime release issue?

girazoki: I only want to downgrade those that fail.

tmpolaczyk (approved):

PR #511, opened 2024-04-19
Implement Default for ExtBuilder and simplify integration tests

tmpolaczyk:
By default set ALICE and BOB as collators and give them some initial balance.

### DIFF
--- a/runtime/dancebox/tests/common/mod.rs
+++ b/runtime/dancebox/tests/common/mod.rs
@@ -15,6 +15,7 @@
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
 use {
+    crate::UNIT,
     cumulus_primitives_core::{ParaId, PersistedValidationData},
     cumulus_primitives_parachain_inherent::ParachainInherentData,
     dancebox_runtime::{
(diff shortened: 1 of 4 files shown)


runtime/dancebox/tests/integration_test.rs
```diff
@@ -67,6 +67,13 @@ fn genesis_balances() {
             // Alice gets 10k extra tokens for her mapping deposit
             (AccountId::from(ALICE), 210_000 * UNIT),
             (AccountId::from(BOB), 100_000 * UNIT),
+            (AccountId::from(CHARLIE), 100_000 * UNIT),
+            (AccountId::from(DAVE), 100_000 * UNIT),
+        ])
+        .with_collators(vec![
+            // Remove ALICE and BOB from collators
+            (AccountId::from(CHARLIE), 100 * UNIT),
+            (AccountId::from(DAVE), 100 * UNIT),
```
tmpolaczyk, line 76: This was needed because if ALICE is a collator, her free balance is 1 unit less than expected and this test fails.

runtime/dancebox/tests/common/mod.rs
```diff
@@ -288,6 +272,53 @@ impl
...
+    // list of registered para ids: para_id, genesis_data, boot_nodes, block_credits, session_credits
+    para_ids: Vec<ParaRegistrationParams>,
+    // configuration to apply
+    config: pallet_configuration::HostConfiguration,
+    safe_xcm_version: Option<u32>,
+    own_para_id: Option<ParaId>,
+}
+
+impl Default for ExtBuilder {
+    fn default() -> Self {
```
girazoki, line 302: Interesting that we did not have a default :(

tmpolaczyk: We did, but it had no collators, so any tests that needed collators had to explicitly define them.

girazoki:
Likely not feasible but is it possible to merge the extBuilder for dancebox & flashbox?

tmpolaczyk:
>Likely not feasible but is it possible to merge the extBuilder for dancebox & flashbox? I know they have different runtimes (with different assumptions in, e.g., collators) but just wondering

We can always make it generic over the runtime, similar to https://github.com/paritytech/polkadot-sdk/blob/eba3deca3e61855c237a33013e8a5e82c479e958/cumulus/parachains/runtimes/test-utils/src/lib.rs#L151

But I prefer to not do it in this PR.

girazoki:
Sounds good

girazoki (approved):

PR #512, opened 2024-04-19
Fix S09C02 and S08C03

ParthDesai:
## Description
This PR fixes `S09C02` and `S08C03` typescript test by conditionally changing `sessionLength` and `inflationIncrement` depending upon whether runtime is > 500 or not.

### DIFF
--- a/test/suites/smoke-test-common/test-inflation-rewards.ts
+++ b/test/suites/smoke-test-common/test-inflation-rewards.ts
@@ -93,8 +93,9 @@ describeSuite({
 
                 const supplyAfter = (await apiAtIssuanceAfter.query.balances.totalIssuance()).toBigInt();
 
-                // expected issuance block increment in prod is: 19n/1_000_000_000n
-                const expectedIssuanceIncrement = (supplyBefore * 19n) / 1_000_000_000n;
+                // expected issuance block increment in prod
+                const expectedIssuanceIncrement =
+                    runtimeVersion > 500 ? (supplyBefore * 9n) / 1_000_000_000n : (supplyBefore * 19n) / 1_000_000_000n;
 
                 // we know there might be rounding errors, so we always check it is in the range +-1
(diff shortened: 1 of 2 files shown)


girazoki:
@ParthDesai merge master back in to make the CI work

girazoki (approved):

test/suites/smoke-test-common/test-inflation-rewards.ts
```diff
@@ -93,8 +93,9 @@ describeSuite({
 
                 const supplyAfter = (await apiAtIssuanceAfter.query.balances.totalIssuance()).toBigInt();
 
-                // expected issuance block increment in prod is: 19n/1_000_000_000n
-                const expectedIssuanceIncrement = (supplyBefore * 19n) / 1_000_000_000n;
+                // expected issuance block increment in prod
+                const expectedIssuanceIncrement =
+                    runtimeVersion > 500 ? (supplyBefore * 9n) / 1_000_000_000n : (supplyBefore * 19n) / 1_000_000_000n;
```
tmpolaczyk, line 98: This could have `const inflationRate = ...` as a separate variable, to make it easier to change again in the future.
```suggestion
                    (supplyBefore * inflationRate) / 1_000_000_000n;
```

tmpolaczyk (approved):

test/suites/smoke-test-common/test-inflation-rewards.ts
ParthDesai: Yes. For this I tried using `PerBill`, but problem is that it returns `BN` instead of `bigint` as result of the operation and I tried to find a way to convert (possibly I do not know enough typescript to do this). 

So, I left it as it is.

tmpolaczyk: Ok it's fine

PR #513, opened 2024-04-22
Refactor container chains folders

nanocryk:
The current folders in `container-chains` are misleading, as the nodes are not templates but data provider nodes that are not expected to be modified by individual projects. This PR thus reorganize crates into a (non template) `nodes` folder and a `runtime-templates` folders, each containing `simple` and `frontier` folders.


.github/workflow-templates/cargo-build/action.yml
```diff
@@ -73,5 +73,5 @@ runs:
       run: |
         mkdir -p binaries
         cp target/release/tanssi-node binaries/tanssi-node;
-        cp target/release/container-chain-template-simple-node binaries/container-chain-template-simple-node;
-        cp target/release/container-chain-template-frontier-node binaries/container-chain-template-frontier-node;
+        cp target/release/container-chain-simple-node binaries/container-chain-simple-node;
```
tmpolaczyk, line 76: This file is in the "workflow-templates" folder, let's rename it to "workflow-containers"?

girazoki: Actually this mean 2 very different things. Workflow-templates only refer to "templates for workflows to be run in the CI". That is the reason why now the coverage does not work. Let's stay with `workflow-template`

girazoki:
The docker jobs are failing because we assume we need to push to `moondance-labs/file-name` and now file-name was renamed from `container-chain-frontier-template-node` to `container-chain-frontier-node`. I see two options here:

- Either we create a new docker repository with the appropriate name or
- We change the publish-docker job to add a "docker_image" field that points to the appropriate repository being used now

Options @nanocryk ?

nanocryk:
The docker repository containing "template" also seems misleading to me, so I'm in favor of changing the name

girazoki:
Alright let me create the new docker repositories then. Luckily this have not been user-oriented for now :)

girazoki (approved):
It's possible that there are things we are missing but I guess we will find those with time

tmpolaczyk (approved):
I run some of the zombienet tests that are disabled in CI and they seem to work fine.

PR #516, opened 2024-04-22
Modifying xcm-core-buyer to add reporting of the core purchase

ParthDesai:
## Description
This PR  converts xcm-core-buyer to add reporting of the core purchase and only allow the purchase if there are no in flight orders. We add `ReportError` instruction in our xcm to report back result of the core purchase xcm instruction back in `xcm-core-buyer`'s `query_response` dispatchable. If we receive a response back, we consider the order to be done (later on we can change it to properly handle success and failure scenario.)

Since, there can be a scenario where we do not get the report back, there is additional `ttl` that will be triggered and the pending order will be cleared to allow the collator to place new buy core order.

I will cover proper benchmarks in follow up PR.


pallets/xcm-core-buyer/src/mock.rs
```diff
@@ -161,6 +164,18 @@ parameter_types! {
     pub const ParachainId: ParaId = ParaId::new(1000);
 }
 
+parameter_types! {
+    pub const CoreBuyingXCMQueryTtl: u32 = 100;
```
girazoki, line 168: what is the difference between these 2?

runtime/dancebox/src/xcm_config.rs
```diff
@@ -477,7 +479,25 @@ parameter_types! {
     pub const XcmBuyExecutionDotRococo: u128 = XCM_BUY_EXECUTION_COST_ROCOCO;
```
girazoki, line 479: this is not being used anywhere AFAICT, is it just for testing=?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -382,23 +595,55 @@ pub mod pallet {
...
-            // TODO: with async backing, it is possible that two tanssi blocks are included in the
-            // same relay block, so this kill is not correct, should only kill the storage if the
-            // relay block number has changed
-            // TODO: this allows collators to send N consecutive messages to buy 1 core for the same
-            // parathread, as long as the parathread block is not included in pallet_author_noting.
-            // So a malicious collator could drain the parathread tank account by buying cores on
-            // every tanssi block but not actually producing any block.
-            InFlightOrders::<T>::kill();
+        fn on_finalize(n: BlockNumberFor<T>) {
+            let mut ttl_queue = InFlightOrdersTtl::<T>::get();
```
girazoki, line 605: if we want to do stuff on_finalize, we have to take the weight into account on_initialize

girazoki: ah sorry we are counting it :+1:

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -382,23 +595,55 @@ pub mod pallet {
...
+
+            loop {
+                maybe_oldest_entry = ttl_queue.first().copied();
+                (oldest_block_number, query_id) = if let Some(oldest_entry) = maybe_oldest_entry {
+                    oldest_entry
+                } else {
+                    break;
+                };
+
+                // No entry at current block number found
```
girazoki, line 620: There is one thing I am thinking. Instead of "automatically" doing this `on_finalize`,  can't we do it such that this is automatically cleaned when a collator tries to send the XCM message? If the current block number - block_number_last_message_sent > ttl then we insert the new one, otherwise we error.

ParthDesai: Hmm. That could work.

pallets/xcm-core-buyer/src/mock.rs
ParthDesai: So, the first `CoreBuyingXCMQueryTtl ` is used in the xcm query itself. If the query response comes  from the relay chain after this `Ttl`, it will be ignored and notification will not be triggered.

The other `ttl` is the additional `ttl` on top of the `CoreBuyingXCMQueryTtl ` to indicate when to clear the in flight order and consider the result as unknown.

runtime/dancebox/src/xcm_config.rs
ParthDesai: Yes. But, I think we should have some limit imposed in proper runtime there.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: I was worried about the storage bloat, but I think this is doable.

ParthDesai: @girazoki We do both. On response it is cleared. And if response is not received within ttl it is cleaned as well as fallback.

girazoki: but only when a collator injects a new one right? or is it automatically done?

ParthDesai: @girazoki Only when collator tries to inject new one. But it is not para id specific. It will try to clean up as many expired entry it could.

runtime/dancebox/tests/common/xcm/expected_event_checker.rs
```diff
@@ -0,0 +1,103 @@
...
+
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+#[macro_export]
```
tmpolaczyk, line 17: Add a comment to why this exists. Like

```suggestion
// Copied from upstream to fix a bug. Remove when PR ... is merged.
#[macro_export]
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -261,6 +379,62 @@ pub mod pallet {
...
+                *para_id
+            } else {
+                // Most probably entry was expired or removed in some other way. Let's return early.
+                return Ok(());
+            };
+
+            let mut in_flight_orders = InFlightOrders::<T>::get();
+            let order = in_flight_orders.get(&para_id).expect("If the QueryId->ParaId mapping exists \
+                    then an entry must be present in the InFlightOrders, if not we should crash as that \
+                    indicates storage inconsistency; qed");
```
tmpolaczyk, line 405: How is `query_response` executed? If it is on `on_initialize`, or some other mandatory mechanism, we should never crash, even if the storage is inconsistent, because that will stall the chain.

A conservative apporach would be to ignore any messages when the state is inconsistent, which can be easily achieved by retuning an error here.

```suggestion
            let order = in_flight_orders.get(&para_id).ok_or(Error::<T>::InFlightOrdersInconsistentState)?;
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -291,15 +465,64 @@ pub mod pallet {
...
 
+        fn clean_up_expired_in_flight_orders(n: BlockNumberFor<T>) {
+            let mut ttl_queue = InFlightOrdersTtl::<T>::get();
+            let mut maybe_oldest_entry;
+            let mut oldest_block_number;
+            let mut query_id;
+
+            let mut query_ids_to_remove = vec![];
+
+            loop {
```
tmpolaczyk, line 476: Infinite loops make me nervous, please rewrite it as a while loop, maybe using

```rust
while let Some((oldest_block_number, query_id)) = ttl_queue.first() {
```

tmpolaczyk: Well that is still an infinite loop, but something along this lines

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -349,56 +572,89 @@ pub mod pallet {
             // We use `descend_origin` instead of wrapping the transact call in `utility.as_derivative`
             // because with `descend_origin` the parathread tank account will pay for fees, while
             // `utility.as_derivative` will make the tanssi sovereign account pay for fees.
+
+            let notify_call = <T as Config>::RuntimeCall::from(Call::<T>::query_response {
+                query_id: 0,
```
tmpolaczyk, line 577: I guess this gets automatically set to a value other than 0 later?

pallets/xcm-core-buyer/src/lib.rs
girazoki: humm I think this can be a bit dangreous if too many need to be cleaned. Is there a reason for this? I think we can only expire the one that the collator is trying to send

pallets/xcm-core-buyer/src/lib.rs
girazoki: it's part of the XCM message which now are executed on_initialize. Yes we should never crash, but rather error

pallets/xcm-core-buyer/src/lib.rs
girazoki: @tmpolaczyk what do you think about this?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -141,15 +228,39 @@ pub mod pallet {
...
+
+    /// Number of pending blocks
+    #[pallet::storage]
+    pub type PendingBlocks<T: Config> =
+        StorageValue<_, BoundedBTreeSet<ParaId, T::MaxNumberOfParaIds>, ValueQuery>;
+
+    /// Mapping of QueryId to ParaId
+    #[pallet::storage]
+    pub type QueryIdToParaId<T: Config> =
+        StorageValue<_, BoundedBTreeMap<QueryId, ParaId, T::MaxInFlightOrders>, ValueQuery>;
```
tmpolaczyk, line 263: I've been thinking about these data structures and I think we can do better. It all depends on the max number of requests that we expect to have in flight at one time.

If the number of requests is low (<100), then we don't need `InFlightOrdersTtl` or `QueryIdToParaId`, because we can just iterate over all the para ids in `InFlightOrders` until we find the one we want. The cost of iterating over a list is smaller than serializing a list, and we already serialize it anyway.

If the number of requests is high (>1000), it's complicated, we probably want to use a `StorageMap` instead of a `StorageValue`, but that also has drawbacks. We would need to write some benchmarks.

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -291,15 +465,64 @@ pub mod pallet {
...
+        }
+
         /// Send an XCM message to the relay chain to try to buy a core for this para_id.
         fn on_collator_instantaneous_core_requested(para_id: ParaId) -> DispatchResult {
+            Self::clean_up_expired_in_flight_orders(<frame_system::Pallet<T>>::block_number());
+
+            let pending_blocks = PendingBlocks::<T>::get();
+            if pending_blocks.contains(&para_id) {
+                return Err(Error::<T>::BlockProductionPending.into());
+            }
```
tmpolaczyk, line 520: This looks wrong, because when an xcm message is successful and the core is bought, the collator can still fail to produce a block. In that scenario we want to allow another collator to try to buy a core again after the timeout.

And also we cannot prevent a third party from manually buying a core in the relay chain, using their own account, so it is possible that a parathread produces a block when we haven't sent an xcm message.

So I would remove the `PendingBlocks` storage item.

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Also, why do we need a response? Wasn't xcm "fire and forget"?

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: So, this `RuntimeCall` is sent across the destination chain and parameter's value does not matter as it will automatically be set to proper query id by destination chain.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: I am not sure here. Reason I chose crash is because if there is no corresponding entry, it could indicate storage inconsistency. 

In that situation, lets the chain work as usual would mean that we are operating on inconsistent state. Instead shouldn't we try to fix it before we allow any child blocks to be executed?

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: Hmm. I think you may be right. If we have only one data structure that we can iterate, it would reduce the complexity. 

Will implement it.

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: I get what you mean, but since this is executed in on_initialize, we cannot fix it, because any extrinsic we send will be executed after the panic, so it won't be executed. You can use this substrate macro to warn in production but panic in tests, this way hopefully we will detect it if anything goes wrong:

https://paritytech.github.io/polkadot-sdk/master/frame_support/macro.defensive.html

ParthDesai: Hmm. If I only have one data structure, I think this problem is eliminated.

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -349,56 +571,81 @@ pub mod pallet {
...
+            InFlightOrders::<T>::set(
+                para_id,
+                Some(InFlightCoreBuyingOrder {
+                    para_id,
+                    query_id,
+                    ttl: in_flight_order_ttl,
+                }),
+            );
+
+            QueryIdToParaId::<T>::set(query_id, Some(para_id));
```
tmpolaczyk, line 634:
```suggestion
            QueryIdToParaId::<T>::insert(query_id, para_id);
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -349,56 +571,81 @@ pub mod pallet {
...
-            // 1 write in on_finalize
-            weight += T::DbWeight::get().writes(1);
 
-            weight
-        }
+        pub fn para_deregistered(para_id: ParaId) {
+            // If para is deregistered we need to clean up in flight order, query id mapping
+            let maybe_in_flight_order = InFlightOrders::<T>::get(para_id);
+            if let Some(in_flight_order) = maybe_in_flight_order {
+                InFlightOrders::<T>::remove(para_id);
```
tmpolaczyk, line 643:
```suggestion
            if let Some(in_flight_order) = InFlightOrders::<T>::take(para_id); {
```

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Also same above

client/consensus/src/collators/lookahead.rs
```diff
@@ -192,6 +197,19 @@ where
...
+                    //                   If we are scheduled, continue on the loop.
+                    //                   If we are not scheduled
+                    //                         If we haven't tried to buy it,  buy pay as you go blockspace in one of the parathread core.
+                    //                         Skip this iteration on the loop.
+                    // (This will replace below is_para_scheduled function call.)
+                    // Handle XCM failure
+
+                    // Parth: If it is parachain, check if there is a core with para id if yes return true.
+                    //  If it is parathread, we need to check if it is `Scheduled` otherwise return true.
+                    //  Else return false.
```
tmpolaczyk, line 212: I guess this comments will be implemented in a future PR? If so, they should be removed

tmpolaczyk (approved):

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -33,26 +34,71 @@ mod benchmarks;
...
+    ttl: BN,
+}
+
+impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
+    fn on_container_author_noted(
+        _author: &T::AccountId,
+        _block_number: BlockNumber,
+        para_id: ParaId,
+    ) -> Weight {
+        PendingBlocks::<T>::remove(para_id);
```
girazoki, line 92: Does this imply a write even if the pendingBlocks does not exist? I am saying this because it is likely that para_ids for **parachains** and not **parathreads** are not present here. Yet we are adding a write for every single one of them

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -349,56 +561,80 @@ pub mod pallet {
...
-            // TODO: with async backing, it is possible that two tanssi blocks are included in the
-            // same relay block, so this kill is not correct, should only kill the storage if the
-            // relay block number has changed
-            // TODO: this allows collators to send N consecutive messages to buy 1 core for the same
-            // parathread, as long as the parathread block is not included in pallet_author_noting.
-            // So a malicious collator could drain the parathread tank account by buying cores on
-            // every tanssi block but not actually producing any block.
-            InFlightOrders::<T>::kill();
+            // We need to clean the pending block entry if any
+            PendingBlocks::<T>::remove(para_id);
```
girazoki, line 637: Same comment as before, maybe worth checking whether it exists=

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: I would expect `remove()` to be more efficient than `if exists() { remove() }` in terms of collator cpu time, but not sure what happens to PoV if you remove a storage item that does not exist.

girazoki: the point is that you are going to call this probably for parachains and parathreads. And yes for parathreads is faster, but not for parachains.

ParthDesai: PoV should not increase in size. PoV is only consist of Markle trie nodes that are being accessed (read) during the block execution.

ParthDesai: @girazoki I went through the entire code path. 

1. For directly calling `remove()`  the change is first written to in-memory overlayed changes data structure  of the runtime api instance from which at the time of block building we extract storage changes. These storage changes  are then applied atomically to the current state of the chain to produce block. Here, advantage of direct removal is that PoV does not increase but we do have additional entry in the overlayed changes which is internal struct and only exists in memory between the time extrinsics are executed to the state is modified (i.e block is imported.).

2. Regarding checking if the key exists or not. This becomes part of the runtime api's `StorageProof` data structure which is used in building PoV.  I believe CPU/memory overhead of local node will be same in both cases but checking existence mean that we are reading something from the storage so it will be recorded by proof recorder and will become part of the storage proof and consequently of the `PoV`. Which I believe is much bigger concern than extra entry in overlaid changes.

So, my opinion would be to directly call `remove`

tmpolaczyk: I guess in the future we should add a `is_parathread: bool` parameter to `on_container_author_noted`, because it is used in pallets inflation-rewards and service-payment, and those will probably be different for parachains vs parathreads.

girazoki (approved):
makes sense thanks for the discussions, good job @ParthDesai !

PR #518, opened 2024-04-22
Change run_block function in integration tests

tmpolaczyk:
Now we do on_finalize then on_initialize, so that after calling this function any extrinsics will be executed as if they were inside a block. Most of the tests are not affected by this change.

TODO (for future PRs): maybe try to extract run_block to a common crate `test_utils`.


girazoki:
@tmpolaczyk did you merge master back for flashbox?

tmpolaczyk:
Not yet

PR #523, opened 2024-04-23
Add tests to tanssi-node

tmpolaczyk:
This new kind of tests that are very hacky, but also quite useful. They work by spawning a new process that behaves similarly to `tanssi-node --dev`, but they do not spawn the release binary, they use the test binary itself. This means that we could add subcommands and flags only for tests, and they will not exist in the final release.

The goal is to be able to test the container_chain_spawner module in a way that more closely resembles production. But this PR only adds some tests related to panics as a starting point, as well as the testing infrastructure.


girazoki (approved):
Looks good Tomasz, dont forget to add the copyright to the new file

PR #526, opened 2024-04-24
remove pallet-hotfix-sufficients from frontier template

girazoki:
Non-needed pallet that was inserted into the frontier-template


tmpolaczyk (approved):
I didn't find any storage related to this pallet in the deployed evm templates, so I guess it does not have any storage that needs to be removed

PR #527, opened 2024-04-24
Frontier template `isBlockFinalized` & `isTxFinalized` RPC methods

fgamundi:
Ports `isBlockFinalized` & `isTxFinalized` RPC methods from Moonbeam to Frontier container chain template.


container-chains/nodes/frontier/src/rpc/finality.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+fn is_block_finalized_inner<B: Block<Hash = H256>, C: HeaderBackend<B> + 'static>(
+    backend: &(dyn fc_api::Backend<B>),
+    client: &C,
+    raw_hash: H256,
+) -> RpcResult<bool> {
+    let substrate_hash =
+        match futures::executor::block_on(frontier_backend_client::load_hash::<B, C>(
```
tmpolaczyk, line 50: I don't understand why are we using an `async_trait` and `block_on` at the same time, but I see moonbeam does the same so I guess just leave it like this...

container-chains/nodes/frontier/src/rpc/finality.rs
```diff
@@ -0,0 +1,73 @@
...
+use sp_runtime::traits::Block;
+use std::{marker::PhantomData, sync::Arc};
+
+#[rpc(server)]
+#[async_trait::async_trait]
+pub trait FrontierFinalityApi {
+    /// Reports whether a Substrate or Ethereum block is finalized.
+    /// Returns false if the block is not found.
+    #[method(name = "frnt_isBlockFinalized")]
+    fn is_block_finalized(&self, block_hash: H256) -> RpcResult<bool>;
```
tmpolaczyk, line 14: Why are we not adding the tx finalized method?

       fn is_tx_finalized(&self, tx_hash: H256) -> RpcResult<bool>;

tmpolaczyk (approved):

container-chains/nodes/frontier/src/rpc/finality.rs
girazoki: yes let's add this, this is fairly important

container-chains/nodes/frontier/src/rpc/finality.rs
fgamundi: I was kinda wondering the same thing. It seems to work fine without the `async_trait` decorator. Maybe it's ok to remove it?

tmpolaczyk: Actually the proper solution is to turn the `fn` into `async fn`, and `block_on` into `await`, but maybe that doesn't work for some reason

fgamundi: Updated the code. Pls take a look

tmpolaczyk: Perfect, you can propose a similar change to moonbeam

tmpolaczyk (approved):

container-chains/nodes/frontier/src/rpc/finality.rs
fgamundi: https://github.com/moonbeam-foundation/moonbeam/pull/2784

PR #528, opened 2024-04-24
Migrate PendingVerification in pallet_registrar from value to map

tmpolaczyk:
This allows an unlimited number of para ids to be waiting for verification, by using an unbounded `StorageMap<ParaId, ()>` instead of a bounded `StorageValue<Vec<ParaId>>`.

Breaking change, includes a migration.

### DIFF
--- a/runtime/common/src/migrations.rs
+++ b/runtime/common/src/migrations.rs
@@ -38,6 +38,7 @@
 use frame_support::ensure;
 
 use {
+    cumulus_primitives_core::ParaId,
     frame_support::{
         pallet_prelude::GetStorageVersion,
         traits::{OnRuntimeUpgrade, PalletInfoAccess, StorageVersion},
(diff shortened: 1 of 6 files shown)


runtime/common/src/migrations.rs
```diff
@@ -377,6 +378,40 @@ where
     }
 }
 
+pub struct RegistrarPendingVerificationValueToMap<T>(pub PhantomData<T>);
```
girazoki, line 381: Can we add a test for the migration?

girazoki (commented):
The PR looks good but I would prefer if the migration was tested. Also we can use the pre and post upgrade hooks for this

tmpolaczyk:
> Also we can use the pre and post upgrade hooks for this

The problem is that this storage item can be empty if there are no pending chains, so I don't know what to test for. But I can add an integration test similar to `test_migration_config_full_rotation_period` to check that the key was calculated correctly.

girazoki:
> > Also we can use the pre and post upgrade hooks for this
> 
> The problem is that this storage item can be empty if there are no pending chains, so I don't know what to test for. But I can add an integration test similar to `test_migration_config_full_rotation_period` to check that the key was calculated correctly.

So I think:

- You manually insert in this storage item certain chains with the previous layout.
- You run on_runtime_upgrade.
- You check the new storage map is populated correctly

Yes similar to `test_migration_config_full_rotation_period`

tmpolaczyk:
Yes, that's already done and it works :)

girazoki (approved):

nanocryk (approved):

PR #530, opened 2024-04-25
forbid teleports and fix command in templates

girazoki:
Fixes the teleport policy as well as the command yield in the frontier template


tmpolaczyk:
Can we add tests?

fgamundi (approved):

girazoki:
> Can we add tests?

yes will do a ts test making sure teleport is forbidden

ParthDesai (approved):

PR #532, opened 2024-04-26
Fix BuildSpec for all configs

fgamundi:
Reworked `BuildSpec` subcommand handling to correctly generate all possible runtimes


node/src/command.rs
```diff
@@ -38,54 +38,62 @@ use {
...
-                "Bob".to_string(),
-                "Charlie".to_string(),
-                "Dave".to_string(),
-            ],
+            container_chains,
+            mock_container_chains,
+            invulnerables,
         )),
-        "" | "dancebox-local" => Box::new(chain_spec::dancebox::local_dancebox_config(
+        "" | "dancebox_local" => Box::new(chain_spec::dancebox::local_dancebox_config(
```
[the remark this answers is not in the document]
tmpolaczyk: Let's add both?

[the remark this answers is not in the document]
fgamundi: They [use dashes](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/cli/src/command.rs#L99), but I changed it to underscore mostly so that I can use [`load_spec` on `BuildSpec`](https://github.com/moondance-labs/tanssi/pull/532/files#diff-8ce03a5a2b79f3021ff2d6322d13448db8cfd5fb9267517ab4b96d0d299ed45fR280) directly, because that's how they are in the chainspecs.
Nevertheless, we can still have the CLI with dashes and have a simple str replace from underscore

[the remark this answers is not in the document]
fgamundi: @tmpolaczyk that's a good idea. We do need to change something, though, as right now it's very confusing that `dancebox-local` uses dash and `dancebox_dev` underscore

test/moonwall.config.json
```diff
@@ -513,7 +513,7 @@
             "name": "dev_flashbox",
             "envVars": ["DEBUG_COLORS=1"],
             "testFileDir": ["suites/common-all", "suites/common-tanssi"],
-            "runScripts": ["compile-wasm.ts compile -b ../target/release/tanssi-node -o wasm -c flashbox_dev"],
+            "runScripts": ["compile-wasm.ts compile -b ../target/release/tanssi-node -o wasm -c flashbox_dev -a \"--mock-container-chain=2000 --mock-container-chain=2001\""],
```
tmpolaczyk, line 516: Is this needed? You do `.unwrap_or(vec![2000, 2001])` if the arg is `None`, so it should be the same without this new arg. (I didn't test it)

tmpolaczyk: Still pretty convinced that you don't need this, because this command is used to compile the runtime wasm, and that does not depend on the genesis state, only on the genesis runtime wasm.

fgamundi: I added it so it replicates the previous behavior of having 2000 and 2001 as mock containers. I now do `.unwrap_or(vec![])` in the `load_spec` fn, to allow generating specs without any mocks.
`compile-wasm.ts` does indeed compile the runtime wasm, but it's also where the [test runtime spec is generated](https://github.com/moondance-labs/tanssi/blob/fg/fix-flashbox-spec/test/scripts/compile-wasm.ts#L97) and saved as `tmp/dancebox_dev-raw.json`, which is the one that's later [used by the TS tests](https://github.com/moondance-labs/tanssi/blob/fg/fix-flashbox-spec/test/moonwall.config.json#L29)

node/src/command.rs
```diff
@@ -125,7 +131,7 @@ impl SubstrateCli for Cli {
     }
 
     fn load_spec(&self, id: &str) -> std::result::Result<Box<dyn sc_service::ChainSpec>, String> {
-        load_spec(id, self.para_id.unwrap_or(1000).into())
+        load_spec(id, self.para_id, None, Some(vec![2000, 2001]), None)
```
tmpolaczyk, line 134: When is this used?

fgamundi: On all other commands that are not build-spec. So launching a network `tanssi-node --dev` will mock the containers

tmpolaczyk: Ok and we were already doing that because the old `load_spec` always added those mock chains, except when the chain_id was a `path`, so same as we do now.

tmpolaczyk (approved):

node/src/command.rs
```diff
@@ -38,55 +38,61 @@ use {
...
-fn load_spec(id: &str, para_id: ParaId) -> std::result::Result<Box<dyn ChainSpec>, String> {
+fn load_spec(
+    id: &str,
+    para_id: Option<u32>,
+    container_chains: Option<Vec<String>>,
+    mock_container_chains: Option<Vec<u32>>,
+    invulnerables: Option<Vec<String>>,
+) -> std::result::Result<Box<dyn ChainSpec>, String> {
+    let para_id: ParaId = para_id.unwrap_or(1000).into();
+    let container_chains = container_chains.unwrap_or(vec![]);
```
tmpolaczyk, line 49: If the default value is going to be an empty vec, you can remove the `Option` from the argument, and have the caller pass `vec![]` instead of `None`.

node/src/command.rs
fgamundi: That's right

girazoki (approved):

PR #534, opened 2024-04-29
Fix smoke test when flashbox chain has no credits

tmpolaczyk:


test/util/payment.ts
```diff
@@ -17,6 +17,7 @@ export async function hasEnoughCredits(
     paraApi: ApiPromise,
     paraId: ParaId,
     blocksPerSession: bigint,
+    // TODO: minSessionRequirement should be 2 if the chain had collators in the previous session, and 1 otherwise
```
tmpolaczyk, line 20: Not sure if there is an easy way to check this, we would need to get the assignment from the previous session

girazoki: I think it's one in either case no?¿

tmpolaczyk: I was thinking about this if condition: https://github.com/moondance-labs/tanssi/blob/443f13c0cc2e4ee44c89273c6bdfbfd61181a62c/runtime/dancebox/src/lib.rs#L795

I guess it depends on when we check this, if we look at the pending assignment we may need 2, if we check the current assignment we need 1

girazoki: This is done in the collator-assignation, and we always (AFAIK) do it with pending. Why do we need 2? We are charging immediatly the collator-assignment for the **next session**. I remember discussing this with @fgamundi

girazoki (approved):

girazoki:
I would not overcomplicate the test. Let's leave it as is, and let's rethink about this in the future

test/util/payment.ts
fgamundi: Yes, it's done with pending. So it goes that if you want to be assigned for the next session, and you are already set to collate for session about to start, you need enough to produces the blocks for this + next session. But the session credits checked to be assigned is always only 1

PR #544, opened 2024-05-02
deprecate at least in container-chain nodes, native executor

girazoki:
We have seen a lot of people customizing the template runtime but not changing the spec name. This makes the full-nodes use the native compiled runtime and fail importing blocks.

I am not a big fan of doing it, but since it's the trend that comes from polkadot-sdk (see https://github.com/paritytech/polkadot-sdk/pull/4329) I guess we should start using only the wasm-executor

### DIFF
--- a/container-chains/nodes/frontier/src/service.rs
+++ b/container-chains/nodes/frontier/src/service.rs
@@ -18,7 +18,6 @@
 
 #[allow(deprecated)]
 use {
-    crate::client::TemplateRuntimeExecutor,
     container_chain_template_frontier_runtime::{opaque::Block, RuntimeApi},
     cumulus_client_cli::CollatorOptions,
     cumulus_client_consensus_common::ParachainBlockImport as TParachainBlockImport,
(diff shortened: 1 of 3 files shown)


bkchr:
> I am not a big fan of doing

Why? :D

girazoki:
> > I am not a big fan of doing
> 
> Why? :D

Just because we lose a bit of efficiency in our tests and those chains we have implementing the template untouched. But I guess it's more a bit of paranoia than anything else

container-chains/nodes/frontier/src/service.rs
```diff
@@ -18,7 +18,6 @@
 
 #[allow(deprecated)]
 use {
-    crate::client::TemplateRuntimeExecutor,
```
ParthDesai, line 21: We need to remove this native executor as well right?

ParthDesai (changes requested):

bkchr:
Okay :)

container-chains/nodes/frontier/src/service.rs
girazoki: yep

girazoki: Updated

ParthDesai (approved):

PR #545, opened 2024-05-07
Services payment RPC

fgamundi:
Adds a runtime API and node RPC methods for getting the block production and collator assignment cost for `ServicesPayment`

### DIFF
--- a/client/services-payment/src/lib.rs
+++ b/client/services-payment/src/lib.rs
@@ -1,0 +1,94 @@
+// Copyright (C) Moondance Labs Ltd.
+// This file is part of Tanssi.
+
+// Tanssi is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+//! RPC client for Services Payment pallet
+
+pub use pallet_services_payment_runtime_api::ServicesPaymentApi as ServicesPaymentRuntimeApi;
+use {
+    core::marker::PhantomData,
+    jsonrpsee::{
+        core::{async_trait, RpcResult},
+        proc_macros::rpc,
+    },
+    sc_client_api::UsageProvider,
+    sp_api::ProvideRuntimeApi,
+    sp_runtime::traits::Block as BlockT,
+    std::sync::Arc,
+};
+
+#[rpc(server)]
+pub trait ServicesPaymentApi<Balance, ParaId> {
+    #[method(name = "tanssi_servicesPaymentBlockCost")]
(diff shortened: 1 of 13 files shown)


test/suites/common-tanssi/services-payment/test_services_payment_rpc.ts
```diff
@@ -0,0 +1,25 @@
...
+                } catch (e: any) {
+                    expect(e.message.toString()).to.eq("No more params");
+                }
+
+                expect(await customDevRpcRequest("tanssi_servicesPaymentBlockCost", [1000])).eq(1000000);
+                expect(await customDevRpcRequest("tanssi_servicesPaymentCollatorAssignmentCost", [1000])).eq(100000000);
+            },
+        });
+    },
+});
```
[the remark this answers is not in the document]
tmpolaczyk: Nevermind, I didn't see it is already here as the last line

client/services-payment/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    std::sync::Arc,
+};
+
+#[rpc(server)]
+pub trait ServicesPaymentApi<Balance, ParaId> {
+    #[method(name = "tanssi_servicesPaymentBlockCost")]
+    async fn block_cost(&self, para_id: ParaId) -> RpcResult<Balance>;
+
+    #[method(name = "tanssi_servicesPaymentCollatorAssignmentCost")]
+    async fn collator_assignment_cost(&self, para_id: ParaId) -> RpcResult<Balance>;
```
tmpolaczyk, line 38: I'm not sure if we want those methods to have a para id as input, the current implementation of `ProvideBlockProductionCost` and `ProvideCollatorAssignmentCost` ignores the para id, because the cost is the same for all para ids.

But on the other hand, if we later change the cost to depend on the para id, we will need to change the rpc method as well...

fgamundi: I had the same thoughts, added it as I saw that the the Providers do receive the ParaId but ignores it. Is it planned for the cost to change depending on the parachain?

tmpolaczyk: I don't know, it would be a good idea to have a different cost if the para id is a parathread or a parachain. But in that case it would be enough to have a bool as input, not a para id.

And the reason we have the para id param in the provider trait is because it was added a long time ago, not because we are planning to do something with it.

I'm thinking that let's just leave it with a para id param, and before the release we revisit this, to avoid breaking the rpc in the future.

fgamundi: I see, sounds good. I've created a ticket to revisit this.

tmpolaczyk (approved):

pallets/services-payment/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) Moondance Labs Ltd.
```
ParthDesai, line 1: I would suggest to move this in `pallets/services-payment/runtime-api/src/` as this runtime api can also be used directly, without rpc.

ParthDesai (approved):
One minor comment. Otherwise looks good.

pallets/services-payment/rpc/runtime-api/src/lib.rs
fgamundi: I agree with you, but I followed how it is defined in other pallets (e.g. `stream-paymet/rpc/runtime-api/src/`). We can create a separate PR changing both paths if we agree on it

ParthDesai: Okay. No problem. Let's do that.

PR #546, opened 2024-05-09
Allow additional account as container chain manager

fgamundi:
Allows the container chain registrar creator to assign a separate account as the only valid manager for that chain in Tanssi.
On chain registration the manager will be set to the creator, and then only the creator can set a new manager.


pallets/registrar/src/lib.rs
```diff
@@ -595,14 +627,17 @@ pub mod pallet {
...
-            RegistrarDeposit::<T>::get(para_id)
+            // * The signed_account is either the deposit creator or the para manager
+            let is_creator = RegistrarDeposit::<T>::get(para_id)
                 .map(|deposit_info| deposit_info.creator)
                 .as_ref()
-                == Some(account)
+                == Some(&account);
+
+            // Short circuit to avoid a DB read if is_creator
+            is_creator || ParaManager::<T>::get(para_id) == Some(account)
```
tmpolaczyk, line 640: This works, but not sure if we want to have more than 1 manager account. I imagine `set_para_manager` will transfer ownership, but in reality it will add an additional manager.

So I would do something like:
```suggestion
            ParaManager::<T>::get(para_id).unwrap_or(deposit_info.creator) == Some(account)
```

pallets/registrar/src/lib.rs
```diff
@@ -229,6 +229,10 @@ pub mod pallet {
     #[pallet::getter(fn registrar_deposit)]
     pub type RegistrarDeposit<T: Config> = StorageMap<_, Blake2_128Concat, ParaId, DepositInfo<T>>;
 
+    #[pallet::storage]
+    pub type ParaManager<T: Config> =
+        StorageMap<_, Blake2_128Concat, ParaId, T::AccountId, OptionQuery>;
```
tmpolaczyk, line 234: `ParaManager` should be removed on deregister, in `cleanup_deregistered_para_id`

pallets/registrar/src/lib.rs
```diff
@@ -1126,7 +1160,7 @@ where
         let signed_account =
             <frame_system::EnsureSigned<_> as EnsureOrigin<_>>::try_origin(o.clone())?;
 
-        if !Pallet::<T>::is_para_manager(para_id, &signed_account) {
+        if !Pallet::<T>::is_para_manager(para_id, signed_account.clone()) {
```
tmpolaczyk, line 1163: You can avoid this clone by using `.as_ref()` inside that function to turn `Option<T>` into `Option<&T>`

pallets/registrar/src/lib.rs
fgamundi: As you mentioned, I approached this as being able to add an additional manager, thinking that most of the time the creator account will be the only manager. I can change it so there's only one manager, and on registration `ParaManager` is set to the creator and `is_para_manager` will then only check `ParaManager`. What do you think?
(c.c @girazoki @ParthDesai)

girazoki: I think what we want to achieve is to have an additional manager that is what we call "`administrator`". This is the go-to approach in pallet-assets.

fgamundi: I like `ParaAdministrator` more than `ParaManager`, to not confuse with the already established manager term in the pallet. Will change it

fgamundi: It was decided that the account stored in `ParaManager` will be the only manager. The PR has been updated so the parachain creator (and only the creator) can at any time set a new manager

pallets/registrar/src/lib.rs
```diff
@@ -585,6 +596,30 @@ pub mod pallet {
 
             Ok(())
         }
+
+        #[pallet::call_index(8)]
+        #[pallet::weight(T::WeightInfo::set_para_manager())]
+        pub fn set_para_manager(
```
girazoki, line 602: Can we by default populate ParaManager with creator at registration? we probably need a migration for that too :(

girazoki: Either that or in `is_para_manager` we compare against the the creator if the paraManager is None

pallets/registrar/src/lib.rs
```diff
@@ -598,11 +633,8 @@ pub mod pallet {
         pub fn is_para_manager(para_id: &ParaId, account: &T::AccountId) -> bool {
             // This check will only pass if both are true:
             // * The para_id has a deposit in pallet_registrar
-            // * The deposit creator is the signed_account
-            RegistrarDeposit::<T>::get(para_id)
-                .map(|deposit_info| deposit_info.creator)
-                .as_ref()
-                == Some(account)
+            // * The signed_account is the para manager
```
tmpolaczyk, line 636: Since we are no longer checking the deposit, you can remove this comment, as the new logic is simple enough to follow without comments.

fgamundi: I've updated the logic again to also check for the creator is the manager is not set (for whatever reason)

tmpolaczyk (approved):

girazoki:
merge master back in @fgamundi please!

girazoki (approved):

PR #550, opened 2024-05-13
Fix: do not shuffle collators before applying priority

tmpolaczyk:
The implementation of collator priority in #471 has a bug that renders it useless: collators are being shuffled twice. This resulted in the priority being applied to a shuffled list of collators, so it was equivalent to having no priority.

This PR fixes that by removing one line: `collators.shuffle(&mut rng);`, and fixing all the tests that break because we changed the way we use the randomness. Also adds a new test `assign_collators_truncates_before_shuffling` to check that collator priority works as expected.

### DIFF
--- a/pallets/collator-assignment/src/tests.rs
+++ b/pallets/collator-assignment/src/tests.rs
@@ -751,21 +751,21 @@ fn assign_collators_rotation() {
 
         // Random assignment depends on the seed, shouldn't change unless the algorithm changes
         let shuffled_assignment = BTreeMap::from_iter(vec![
-            (1, 1004),
+            (1, 1003),
             (2, 1000),
-            (3, 1000),
+            (3, 1001),
             (4, 1003),
-            (5, 1001),
+            (5, 1000),
             (6, 1000),
-            (7, 1002),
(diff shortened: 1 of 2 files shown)


girazoki (approved):

pallets/collator-assignment/src/tests.rs
```diff
@@ -794,15 +794,15 @@ fn assign_collators_rotation_container_chains_are_shuffled() {
 
         MockData::mutate(|m| {
             // Seed chosen manually to see the case where container 1002 is given priority
-            m.random_seed = [2; 32];
```
ParthDesai, line 797: Why did you change the random seed? Even if you don't change it, wouldn't it result in different order in fixed code now?

ParthDesai (approved):
Looks good to me. Just one minor question.

pallets/collator-assignment/src/tests.rs
tmpolaczyk: If I don't change it collators get assigned to container 1001, and this test is supposed to check the case where they are assigned to 1002.

PR #551, opened 2024-05-13
Data Preservers Profiles

nanocryk:
Allow users to register profile for their data provider nodes. Profiles can be forcefully managed by origins configured in the runtime.

Later PRs will replace the current `BootNodes` storage with assignements from chain managers, both in a forced/free manner and a paid manner (which will add in profiles the payment plan required by the provider).


girazoki:
hey @nanocryk please make the CI pass while we review this

nanocryk:
> Can we include a migration creating a profile for all existing bootnodes? I feel this would reduce the complexity in the follow-up PRs as you already have the correct data in storage

We'll add it in the PR that'll add assignements and will effectively replace `Bootnodes`. Since assignements will deal with payment options (free, stream, other?) it'll add a new field to profiles, which would then require a second migration.

pallets/data-preservers/src/lib.rs
```diff
@@ -130,6 +172,84 @@ pub mod pallet {
...
+
+    /// Data preserver profile.
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    #[derive(
+        RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, Encode, Decode, CloneNoBound, TypeInfo,
+    )]
+    #[scale_info(skip_type_params(T))]
+    pub struct Profile<T: Config> {
+        pub url: BoundedVec<u8, T::MaxBootNodeUrlLen>,
+        pub limited_to_para_ids: Option<BoundedVec<ParaId, T::MaxParaIdsVecLen>>,
```
tmpolaczyk, line 196: It would also be interesting to allow blacklisting some para ids, I imagine the case of a chain that is very active so this data preserver doesn't want to handle all the load.

pallets/data-preservers/src/lib.rs
```diff
@@ -151,6 +271,246 @@ pub mod pallet {
...
+                Err(Error::<T>::UnknownProfileId)?
+            };
+
+            ensure!(
+                existing_profile.account == account,
+                sp_runtime::DispatchError::BadOrigin,
+            );
+
+            // Update deposit
+            let new_deposit = T::ProfileDeposit::profile_deposit(&profile)?;
```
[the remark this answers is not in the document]
nanocryk: Both can work. You can also impl this trait ignoring the profile and returning a const worst case cost :) This approach in the pallet is more versatile.

pallets/data-preservers/src/lib.rs
nanocryk: so you want to have the choice between whitelisting and blacklisting?

pallets/data-preservers/src/lib.rs
```diff
@@ -130,6 +172,84 @@ pub mod pallet {
...
+        StorageMap<_, Blake2_128Concat, ProfileId, RegisteredProfile<T>, OptionQuery>;
+
+    #[pallet::storage]
+    pub type NextProfileId<T: Config> = StorageValue<_, ProfileId, ValueQuery>;
+
+    /// Balance used by this pallet
+    pub type BalanceOf<T> =
+        <<T as Config>::Currency as Inspect<<T as frame_system::Config>::AccountId>>::Balance;
+
+    pub type ProfileId = u64;
```
girazoki, line 186: maybe this can be made a config item too?

pallets/data-preservers/src/lib.rs
girazoki: I guess it's not a bad idea and it does not cost much to add it

girazoki: Dont forget to update the deposit as we have new bytes now

pallets/data-preservers/src/lib.rs
```diff
@@ -130,6 +172,84 @@ pub mod pallet {
...
+        pub account: T::AccountId,
+        pub deposit: BalanceOf<T>,
+        pub profile: Profile<T>,
+    }
+
+    pub trait ProfileDeposit<Profile, Balance> {
+        fn profile_deposit(profile: &Profile) -> Result<Balance, DispatchErrorWithPostInfo>;
+    }
+
+    pub struct BytesProfileDeposit<BaseCost, ByteCost>(PhantomData<(BaseCost, ByteCost)>);
```
girazoki, line 225: I think this is not a bad idea to be honest, but it's the first time that I see it in the pallet haha. I like it but since it's a configuration option maybe we can move it to a different file?

girazoki: Also let's add some comments to it

pallets/data-preservers/src/lib.rs
```diff
@@ -151,6 +271,246 @@ pub mod pallet {
...
+
+            let deposit = T::ProfileDeposit::profile_deposit(&profile)?;
+            T::Currency::hold(&HoldReason::ProfileDeposit.into(), &account, deposit)?;
+
+            let id = NextProfileId::<T>::get();
+            NextProfileId::<T>::set(id.checked_add(1).ok_or(ArithmeticError::Overflow)?);
+
+            ensure!(
+                !Profiles::<T>::contains_key(id),
+                Error::<T>::NextProfileIdShouldBeAvailable
```
girazoki, line 294: This would be a very bad thing right? It would mean that something goes really wrong. In that case can we add a extrinsic to set the next profile id?

pallets/data-preservers/src/lib.rs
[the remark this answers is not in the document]
girazoki: I like that we adjust it to whatever profile is being created tbh. It's definitely not the same to create a profile where you inject 100 blacklisted para-ids than one where you insert none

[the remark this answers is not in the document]
tmpolaczyk: Yeah but what is the max difference going to be? 1000 bytes? How expensive is a deposit for that? My point is that I don't like to be handling dust on every profile update.

pallets/data-preservers/src/lib.rs
nanocryk: If it were to actually overflow it would be very bad, thus why I protect for it. However it shouldn't overflow with the proper Id type. As you suggested above I can make ProfileId generic and configurable, however one must be carefull it will not realistically overflow.

tmpolaczyk: `ParaId` is a u32 in the relay chain and it has exactly the same implementation, so I wouldn't worry about overflows.

pallets/data-preservers/src/lib.rs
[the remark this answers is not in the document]
girazoki: how are you handling dust? you are just either releasing some amount or holding more tokens. And it's the user paying for it..

pallets/data-preservers/src/lib.rs
girazoki: let's imagine we code something wrong (e.g., the way we assign profile ids, that at some point is not incremented for whatever reason). In this case no-one else would be able to set a profile id anymore. Hence my suggestion for adding a `force_next_profile_id` that can set it to whatever you want

pallets/data-preservers/src/lib.rs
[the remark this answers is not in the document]
tmpolaczyk: I mean that the difference in deposit is so small that it doesn't matter, but I'm just guessing, maybe deposits are more expensive than I think.

pallets/data-preservers/src/lib.rs
tmpolaczyk: I don't think that's useful, if there's a bug we will release a new runtime with a migration and never use that extrinsic again, and we can just sudo.setStorage instead.

pallets/data-preservers/src/lib.rs
nanocryk: Deposit is computed from size so it should already take into account the change.

pallets/data-preservers/src/benchmarks.rs
```diff
@@ -55,5 +76,209 @@ mod benchmarks {
         assert_eq!(Pallet::<T>::boot_nodes(para_id), boot_nodes);
     }
 
-    impl_benchmark_test_suite!(Pallet, crate::mock::new_test_ext(), crate::mock::Test);
+    #[benchmark]
+    fn create_profile(x: Linear<1, 200>, y: Linear<1, 10>) {
```
tmpolaczyk, line 80: According to the weights, none of the benchmarks depend on x or y, that's great!

nanocryk: In the runtime weigts it seems to use x and y but with very small multipliers. Should we simplify and remove those parameters?

tmpolaczyk: I would leave it, or maybe use one param which is "profile size", but it's fine

pallets/data-preservers/src/lib.rs
girazoki: fair enough

girazoki (approved):

PR #553, opened 2024-05-17
Allow to register para using relay proof

tmpolaczyk:
WIP

### DIFF
--- a/pallets/registrar/src/lib.rs
+++ b/pallets/registrar/src/lib.rs
@@ -40,11 +40,16 @@ pub use weights::WeightInfo;
 
 pub use pallet::*;
 
+use dp_chain_state_snapshot::GenericStateProof;
+use dp_chain_state_snapshot::ReadEntryErr;
+use sp_core::H256;
+use sp_runtime::traits::Convert;
+use sp_runtime::traits::Verify;
 use {
     frame_support::{
         pallet_prelude::*,
         traits::{Currency, ReservableCurrency},
-        DefaultNoBound, LOG_TARGET,
+        DefaultNoBound, Hashable, LOG_TARGET,
     },
     frame_system::pallet_prelude::*,
     parity_scale_codec::{Decode, Encode},
@@ -53,13 +58,13 @@ use {
     tp_container_chain_genesis_data::ContainerChainGenesisData,
     tp_traits::{
         GetCurrentContainerChains, GetSessionContainerChains, GetSessionIndex, ParaId,
-        ParathreadParams as ParathreadParamsTy, SlotFrequency,
+        ParathreadParams as ParathreadParamsTy, SessionContainerChains, SlotFrequency,
     },
 };
 
 #[frame_support::pallet]
 pub mod pallet {
-    use {super::*, tp_traits::SessionContainerChains};
+    use super::*;
 
     #[pallet::pallet]
     #[pallet::without_storage_info]
@@ -129,6 +134,12 @@ pub mod pallet {
 
         #[pallet::constant]
         type MaxLengthTokenSymbol: Get<u32>;
+
+        #[pallet::constant]
+        type AllowRegisterWithRelayProof: Get<bool>;
+
+        // TODO: proper trait
+        type RelayStorageRootProvider: Convert<u32, Option<H256>>;
 
         type SessionIndex: parity_scale_codec::FullCodec + TypeInfo + Copy + AtLeast32BitUnsigned;
 
@@ -268,6 +279,10 @@ pub mod pallet {
         NotSufficientDeposit,
         /// Tried to change parathread params for a para id that is not a registered parathread
         NotAParathread,
+        /// register_with_relay_proof disabled in config
+        RegisterWithRelayProofDisabledInConfig,
+        /// The relay storage root for the corresponding block number could not be retrieved
+        RelayStorageRootNotFound,
     }
 
     #[pallet::hooks]
@@ -407,40 +422,7 @@ pub mod pallet {
         pub fn deregister(origin: OriginFor<T>, para_id: ParaId) -> DispatchResult {
             T::RegistrarOrigin::ensure_origin(origin)?;
 
-            // Check if the para id is in "PendingVerification".
-            // This is a special case because then we can remove it immediately, instead of waiting 2 sessions.
-            let is_pending_verification = PendingVerification::<T>::take(para_id).is_some();
-            if is_pending_verification {
-                Self::deposit_event(Event::ParaIdDeregistered { para_id });
-                // Cleanup immediately
-                Self::cleanup_deregistered_para_id(para_id);
-            } else {
-                Self::schedule_paused_parachain_change(|para_ids, paused| {
-                    // We have to find out where, in the sorted vec the para id is, if anywhere.
-
-                    match para_ids.binary_search(&para_id) {
-                        Ok(index) => {
-                            para_ids.remove(index);
-                        }
-                        Err(_) => {
-                            // If the para id is not registered, it may be paused. In that case, remove it from there
-                            match paused.binary_search(&para_id) {
-                                Ok(index) => {
-                                    paused.remove(index);
-                                }
-                                Err(_) => {
-                                    return Err(Error::<T>::ParaIdNotRegistered.into());
-                                }
-                            }
-                        }
-                    }
-
-                    Ok(())
-                })?;
-                // Mark this para id for cleanup later
-                Self::schedule_parachain_cleanup(para_id)?;
-                Self::deposit_event(Event::ParaIdDeregistered { para_id });
-            }
+            Self::do_deregister(para_id)?;
 
             Ok(())
         }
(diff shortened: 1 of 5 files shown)


pallets/registrar/src/lib.rs
```diff
@@ -585,6 +540,152 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            para_id: ParaId,
+            parathread_params: Option<ParathreadParamsTy>,
+            relay_proof_block_number: u32,
+            relay_storage_proof: sp_trie::StorageProof,
+            manager_signature: cumulus_primitives_core::relay_chain::Signature,
+            genesis_data: ContainerChainGenesisData<T::MaxLengthTokenSymbol>,
+        ) -> DispatchResult {
+            let account = ensure_signed(origin)?;
+            if !T::AllowRegisterWithRelayProof::get() {
```
girazoki, line 558: Maybe we can prevent this at the origin level? (EnsureNever)

pallets/registrar/src/lib.rs
```diff
@@ -585,6 +540,152 @@ pub mod pallet {
...
+            // Should include:
+            // * para_id, in case the manager has more than 1 para in the relay
+            // * accountid in tanssi, to ensure that the creator role is assigned to the desired account
+            // * relay_proof_block_number or something time based, to make the signature expiring
+            //   ^ this can make UX very bad, but maybe we can check a conservative range and not the
+            //   exact block number? eg by using (relay_proof_block_number / 128) as the signature msg
+            // * parathread_params, to avoid registering a parathread as parachain and viceversa
+            //   ^ maybe not needed, the relay manager should trust the tanssi creator
+            // * also don't need genesis_data as part of the signature
+            let signature_msg: Vec<u8> = (para_id, &account, relay_proof_block_number).encode();
```
girazoki, line 599: We should indeed speak about this yes, as it is important, A few ideas:

- the signature includes something network-specific (so that a signature cannot be re-used across networks). Maybe the relay block hash against which we are veryfing could be a good sign
- By default by signing the block hash this is expiring, as this is one of our stored last N block hashes

test/scripts/build-spec.sh
```diff
@@ -7,7 +7,7 @@ set -e
 cd $(dirname $0)/..
 
 mkdir -p specs
-../target/release/container-chain-simple-node build-spec --disable-default-bootnode --add-bootnode "/ip4/127.0.0.1/tcp/33049/ws/p2p/12D3KooWHVMhQDHBpj9vQmssgyfspYecgV6e3hH1dQVDUkUbCYC9" --parachain-id 2000 --raw > specs/template-container-2000.json
-../target/release/container-chain-frontier-node build-spec --disable-default-bootnode --add-bootnode "/ip4/127.0.0.1/tcp/33050/ws/p2p/12D3KooWFGaw1rxB6MSuN3ucuBm7hMq5pBFJbEoqTyth4cG483Cc" --parachain-id 2001 --raw > specs/template-container-2001.json
+../target/release/container-chain-simple-node build-spec --disable-default-bootnode --parachain-id 2000 --raw > specs/template-container-2000.json
+../target/release/container-chain-frontier-node build-spec --disable-default-bootnode --parachain-id 2001 --raw > specs/template-container-2001.json
```
tmpolaczyk, line 11: Undo, it turns out those bootnodes are not needed, but this shouldn't be in this PR

ParthDesai (approved):

pallets/registrar/src/benchmark_blob.rs
```diff
@@ -0,0 +1,162 @@
...
+            // Account generated by
+            // create_funded_user::<T>("caller", 0, T::DepositAmount::get());
+            let account32 = AccountId32::decode(
+                &mut [
+                    92u8, 119, 3, 63, 206, 138, 144, 69, 130, 74, 102, 144, 187, 249, 156, 109,
+                    178, 105, 80, 47, 10, 141, 29, 42, 0, 133, 66, 213, 105, 10, 7, 73,
+                ]
+                .as_slice(),
+            )
+            .unwrap();
```
girazoki, line 144: An idea to make these run faster:remove the y parameteres in both register and deregister. In the `register` benchmarks we can completely ignore the y parameter as we have a mapping now for the chains pending verification.

In the `deregister` extrinsic, let's assume x and y is always max, since it's what we are doing anyways in the pallet. That way we iterate only over z

pallets/registrar/src/lib.rs
```diff
@@ -1182,3 +1343,22 @@ where
...
 }
+
+// TODO: import this from dancekit
+pub const REGISTRAR_PARAS_INDEX: &[u8] =
+    &hex_literal::hex!["3fba98689ebed1138735e0e7a5a790abcd710b30bd2eab0352ddcc26417aa194"];
+
+// Need to copy ParaInfo from
+// polkadot-sdk/polkadot/runtime/common/src/paras_registrar/mod.rs
+// Because its fields are not public...
+// TODO: import this from dancekit
```
girazoki, line 1354: Let's make a PR to parity to make them public, it does not make any sense that they are not

pallets/registrar/src/lib.rs
```diff
@@ -743,6 +819,91 @@ pub mod pallet {
...
+
+            T::RegistrarHooks::para_marked_valid_for_collating(para_id);
+
+            Ok(())
+        }
+
+        /// Relay parachain manager signature message. Includes:
+        /// * para_id, in case the manager has more than 1 para in the relay
+        /// * accountid in tanssi, to ensure that the creator role is assigned to the desired account
+        /// * relay_storage_root, to make the signature network-specific, and also make it expire
```
girazoki, line 897: Nice!

pallets/registrar/src/lib.rs
```diff
@@ -620,6 +586,116 @@ pub mod pallet {
...
+                    cumulus_primitives_core::relay_chain::AccountId,
+                    cumulus_primitives_core::relay_chain::Balance,
+                >>(key.as_slice())
+                .map_err(|_| Error::<T>::InvalidRelayStorageProof)?;
+            if relay_para_info.is_some() {
+                return Err(Error::<T>::ParaStillExistsInRelay.into());
+            }
+
+            // Take the deposit immediately and give it to origin account
+            if let Some(asset_info) = RegistrarDeposit::<T>::take(para_id) {
```
girazoki, line 682: This needs to come after the deregister. Because:

- if the deregister has already been made by the manager, then it is very likely that the deregistration in the relay has not finished yet. Therefore this slashing condition will apply. And the deposit will still have not been unreserved because we do it when the deregistration completes.

- In contrast if we do it after the deregister, the deregister should error if you are trying to deregister something that is already in the deregistering process.

What do you think? If agreed, can we create a test for this?

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
```diff
@@ -0,0 +1,173 @@
...
+                const credits2001 = (await polkadotJs.query.servicesPayment.blockProductionCredits(2001)).toJSON();
+                expect(credits2001).to.be.null;
+            },
+        });
+    },
+});
+
+// Creating a relay storage proof in dev tests is not possible, because there is no relay chain, but we can re-use the
+// proof from setValidationData, which includes the registrar->paras entries because we added them.
+async function fetchStorageProofFromValidationData(polkadotJs) {
```
girazoki, line 148: love this

pallets/registrar/src/lib.rs
girazoki: Nevermind, I am still with the past mentality in which extrinsics where not transactional. Forget about it :)

tmpolaczyk: The deposit needs to be taken before, because there is an edge case in deregister that allows the deposit to be removed immediately, so it will not exist if we try to take the deposit after the call to deregister. But I should add tests.

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
tmpolaczyk: I hate it

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
```diff
@@ -0,0 +1,173 @@
...
+                // TODO: fix once we have types
+                expect(parasScheduled.toJSON()).to.deep.equal([2000]);
+
+                // Checking that in session 2 paras are registered
+                await jumpSessions(context, 2);
+
+                // Expect now paraIds to be registered
+                const parasRegistered = await polkadotJs.query.registrar.registeredParaIds();
+                // TODO: fix once we have types
+                expect(parasRegistered.toJSON()).to.deep.equal([2000]);
```
girazoki, line 82: Can we check we get the deposit?

girazoki: that we slash the deposit I mean

test/suites/dev-tanssi/register-with-relay-proof/test_register_with_relay_proof.ts
```diff
@@ -0,0 +1,195 @@
...
+            polkadotJs = context.polkadotJs();
+
+            // We must generate the same account as in service.rs, using get_ed25519_pairs
+            // Instead of having to implement that function in javascript, we can just hardcode the private key here
+            // Create a new keyring because we need to use ed25519
+            const relayKeyring = new Keyring({
+                type: "ed25519",
+            });
+            relayManager = relayKeyring.addFromUri(
+                "0x3132333435363738393031323334353637383930313233343536373839303132"
```
girazoki, line 27: let's put this into a constnat

test/suites/dev-tanssi/register-with-relay-proof/test_register_with_relay_proof.ts
```diff
@@ -0,0 +1,195 @@
...
+                const givenFreeCredits2 = await polkadotJs.query.servicesPayment.givenFreeCredits(paraId);
+                expect(givenFreeCredits2.isNone).to.be.true;
+            },
+        });
+    },
+});
+
+// Creating a relay storage proof in dev tests is not possible, because there is no relay chain, but we can re-use the
+// proof from setValidationData, which includes the registrar->paras entries because we added them.
+async function fetchStorageProofFromValidationData(polkadotJs) {
```
girazoki, line 170: Move to utils

girazoki (approved):
Besides small comments (the benchmark can be addressed in a follow up PR), looks good @tmpolaczyk, good job!

test/suites/dev-tanssi/register-with-relay-proof/test_register_with_relay_proof.ts
tmpolaczyk: In utils or here?

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
tmpolaczyk: In this test I cannot do it because that parachain is registered in genesis, so it doesn't have a deposit.

girazoki: mmm it's a pity

girazoki: can we register another chain insted?

tmpolaczyk: I registered chain 2003 instead

PR #555, opened 2024-05-20
Wait for lock file in container chain restart

### LINKED ISSUE #554: Collators sometimes fail to spawn a container-chain collation process because DB is in use
Collators occasionally fail to spawn a container-chain collation process because apparently the DB they are trying to use (which comes from the previously spawned sync process) is still in use. This probably indicates that the amount of time we wait between closing the syncing process and spawning the collator one is not enough (currently, 10 seconds). Logs:
```

2024-05-18 16:38:33.022 INFO tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Stopping container chain 3098
2024-05-18 16:38:33.022 INFO tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Stopping container chain 3041
2024-05-18 16:38:34.660 WARN tokio-runtime-worker parity-db: Shutdown error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
2024-05-18 16:38:43.023 INFO tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Detected assignment for container chain 3098
2024-05-18 16:38:43.023 INFO tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Reading container chain genesis data from orchestrator chain at block #2090404 0x992d…1f3e
2024-05-18 16:38:43.027 INFO tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Loaded chain spec for container chain 3098
2024-05-18 16:38:43.027 ERROR tokio-runtime-worker tanssi_node::container_chain_spawner: [Orchestrator] Failed to start container chain 3098: Backend error: Database file is in use. (Resource temporarily unavailable (os error 11))
```
Restarting the collator fixes this, or even a future rotation event. However, we should make something more robust and check the existence of the db-lock before spawning the collator, re-trying for a given specified timeout period until the lock is released

tmpolaczyk:
Improves the hack we use to wait for the database to close when restarting a container chain. Now instead of a `sleep(10)`, we wait until the database lock file is released (with a 60 second timeout).

Fix #554

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15415,6 +15415,7 @@ dependencies = [
  "flume 0.10.14",
  "frame-benchmarking",
  "frame-benchmarking-cli",
+ "fs2",
  "futures 0.3.30",
  "jsonrpsee",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -311,6 +311,7 @@ clap = { version = "4.1.6", default-features = false, features = [ "derive" ] }
 core_extensions = "1.5.3"
 exit-future = { version = "0.2.0" }
 flume = "0.10.9"
+fs2 = "0.4.3"
 futures = { version = "0.3.1" }
 futures-timer = "3.0.1"
 hex = { version = "0.4.3", default-features = false }
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -16,6 +16,7 @@ async-trait = { workspace = true }
 clap = { workspace = true, features = [ "derive" ] }
 exit-future = { workspace = true }
 flume = { workspace = true }
+fs2 = { workspace = true }
 futures = { workspace = true }
 jsonrpsee = { workspace = true, features = [ "server" ] }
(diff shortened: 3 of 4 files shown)


girazoki (approved):
nice work @tmpolaczyk

ParthDesai:
@tmpolaczyk Is it not possible to drop db when we are shutting down the node? That way we don't have to do the timeout stuff?

tmpolaczyk:
>@tmpolaczyk Is it not possible to drop db when we are shutting down the node? That way we don't have to do the timeout stuff?

I tried that in the past, even after dropping the db and waiting for the reference count of the `Arc<Backend>` to go to 0, the db may still be open. My guess is that paritydb has some background thread that keeps the db open, and I don't know if there is a way to wait for that thread to finish.

ParthDesai (approved):

PR #556, opened 2024-05-21
Remove empty folders after removing container db folder

tmpolaczyk:
We still leave one folder, "network", because it has a p2p key inside: "secret_ed25519".

### DIFF
--- a/node/src/container_chain_spawner.rs
+++ b/node/src/container_chain_spawner.rs
@@ -742,14 +742,48 @@ fn open_and_maybe_delete_db(
     Ok(())
 }
 
-// TODO: this leaves some empty folders behind, because it is called with db_path:
-//     Collator2002-01/data/containers/chains/simple_container_2002/paritydb/full-container-2002
-// but we want to delete everything under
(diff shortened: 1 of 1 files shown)


node/src/container_chain_spawner.rs
```diff
@@ -742,14 +742,48 @@ fn open_and_maybe_delete_db(
...
+    // Remove folder `full-container-2002`
+    let _ = std::fs::remove_dir_all(db_path);
+    if let Some(parent) = db_path.parent() {
+        // Remove folder `paritydb` if empty
+        let _ = delete_empty_folders_recursive(parent);
+        if let Some(parent2) = parent.parent() {
+            // Remove folder `simple_container_2002` if empty
+            let _ = delete_empty_folders_recursive(parent2);
+        }
+    }
```
fgamundi, line 761: Wouldn't just running `delete_empty_folders_recursive(parent2)` suffice?

tmpolaczyk: Yes you are right

girazoki (approved):

PR #557, opened 2024-05-22
Script benchmarking all runtimes

fgamundi:
Script that runs runtime benchmarks for all four current runtimes to simplify the process in the benchmarking server


tools/benchmark_all_runtimes.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
```
tmpolaczyk, line 2: I like to start my scripts like this, especially the "cd dirname", because it allows you to run the script from any folder and get the same result

```sh
# Exit on any error
set -e

# Always run the commands from the project dir
cd "$(dirname "$0")/.."
```

fgamundi: Uuhh I like that! Committed

tmpolaczyk: Also the file is not executable, run `chmod +x script.sh` and commit

fgamundi: Done

girazoki (approved):
LGMT!

PR #559, opened 2024-05-24
Add signature verification + additional validation in xcm-core-buyer

ParthDesai:
## Description
This PR adds signature verification in the xcm-core-buyer. So that only collator can purchase the core. Apart from that the pallet now only allows collator to buy core if it is inline with the slot frequency.

Changes are as follows:
1. xcm-core-buyer now takes into account slot frequency.
2. couple of hooks added in xcm-core-buyer to get current slot and latest author info from other pallets.
3. xcm-core-buyer now do signature verification in validate_unsigned for buy_core
4. two runtime api exposed: 1. Get slot frequency 2. Check whether a client is allowed to buy core or not which can be used by client to decide whether it should buy core or not.

Remaining:
- [x] Benchmarks
- [x] Tests


pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -640,44 +733,31 @@ pub mod pallet {
...
-                    return InvalidTransaction::Stale.into()
+            if let Call::buy_core {
+                para_id,
+                collator_account_id,
+                proof,
+            } = call
+            {
+                let block_number = <frame_system::Pallet<T>>::block_number();
+
+                let current_nonce = CollatorSignatureNonce::<T>::get();
```
tmpolaczyk, line 744: Why do we need a nonce? Can't we use the block number as the nonce for example?

ParthDesai: Hmm. Yeah. We can use block number as nonce. But problem is that it is not guaranteed at which block this tx will be included and any tx signed with earlier block number can become invalid. 

So, the idea is to have nonce just like a substrate account but shared between all calls.

tmpolaczyk: I don't like that it's an extra storage map. But block number won't work, because the tx may not be included in the next block, you're right. Container chain block number (from author-noting pallet) also won't work, because in case of failure we won't be able to retry... There must be something that we can use though.

ParthDesai: It is not extra storage map. It's just one storage value at the moment. 

I think it is nice trade off between amount of space and the uncertainity.

tmpolaczyk: Oh, really? Sorry, I haven't look into the code yet, but what will happen if two collators try to buy a core in the same block? Will it work or will it fail?

ParthDesai: One of them would succeed and other one would fail. That is sort of drawback. This is the reason why I wanted it originally as storage map.

On second thought, let me modify it as storage map first and then we can arrive at equivalent solution.

tmpolaczyk: Yeah I agree, let's work with a storage map now, write some tests to make sure everything works, and then we can think about how to remove this storage map.

pallets/registrar/rpc/runtime-api/src/lib.rs
```diff
@@ -42,13 +43,17 @@ sp_api::decl_runtime_apis! {
...
+        ///
+        /// Simply put, parathread must produce a block after `min`  but before `(min+max)` slots.
         ///
         /// # Returns
         ///
-        /// * `Some(min)`, where the condition for the slot to be valid is `(slot - parent_slot) >= min`.
+        /// * `Some(slot_frequency)`.
         /// * `None` if the `para_id` is not a parathread.
-        fn min_slot_freq(para_id: ParaId) -> Option<Slot>;
+        fn parathread_slot_frequency(para_id: ParaId) -> Option<SlotFrequency>;
```
girazoki, line 57: We need to make sure we were not using this before, although I think we were not

pallets/xcm-core-buyer/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+//! Runtime API for XCM core buyer pallet
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use pallet_xcm_core_buyer::BuyingError;
+
+sp_api::decl_runtime_apis! {
```
girazoki, line 23: FMT

pallets/xcm-core-buyer/src/benchmarks.rs
```diff
@@ -45,6 +46,62 @@ mod benchmarks {
...
+                buy_execution_cost: BUY_EXECUTION_COST,
+                weight_at_most: PLACE_ORDER_WEIGHT_AT_MOST,
+                _phantom: PhantomData,
+            }),
+        ));
+
+        let x = 1000u32;
+
+        let para_id = ParaId::from(x + 1);
+        for i in 0..=x {
```
girazoki, line 64: I assume you dont want your benchmark to depend on X here, hence why 1000? was this chosen in some arbitrary form?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -218,15 +256,73 @@ pub mod pallet {
...
+
+        #[cfg(test)]
+        pub fn set_public_key(&mut self, public_key: PublicKey) {
+            self.public_key = public_key;
+        }
+
+        #[cfg(test)]
+        pub fn set_signature(&mut self, signature: PublicKey::Signature) {
+            self.signature = signature;
+        }
```
tmpolaczyk, line 325: I would make the fields public instead of having setters

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -218,15 +256,73 @@ pub mod pallet {
...
+    where
+        PublicKey: RuntimeAppPublic + Clone + core::fmt::Debug,
+    {
+        fn verify_signature(&self, para_id: ParaId) -> bool {
+            let payload = (self.nonce, para_id).encode();
+            self.public_key.verify(&payload, &self.signature)
+        }
+
+        #[cfg(any(test, feature = "runtime-benchmarks"))]
+        pub fn new(nonce: u64, para_id: ParaId, public_key: PublicKey) -> Option<Self> {
```
tmpolaczyk, line 301:
```suggestion
        pub fn new(nonce: u64, para_id: ParaId, public_key: PublicKey) -> Option<Self> {
```

Collators will probably need this to be public to create the proof, right?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -218,15 +256,73 @@ pub mod pallet {
...
     }
 
     /// Proof that I am a collator, assigned to a para_id, and I can buy a core for that para_id
     #[derive(Encode, Decode, CloneNoBound, PartialEq, Eq, DebugNoBound, TypeInfo)]
-    #[scale_info(skip_type_params(T))]
-    pub struct BuyCoreCollatorProof<T: Config> {
-        account: T::AccountId,
-        // TODO
-        _signature: (),
+    pub struct BuyCoreCollatorProof<PublicKey>
```
girazoki, line 282: Maybe move this to a separate file called proof.rs or something similar?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -506,17 +597,45 @@ pub mod pallet {
...
+                    });
                 }
             }
 
             // Check that the para id is a parathread
-            let parathread_params = T::GetParathreadParams::get_parathread_params(para_id);
-            if parathread_params.is_none() {
-                return Err(Error::<T>::NotAParathread.into());
+            let maybe_parathread_params = T::GetParathreadParams::get_parathread_params(para_id);
+            let parathread_params = if let Some(parathread_params) = maybe_parathread_params {
```
girazoki, line 609:
```suggestion
            let parathread_params = T::GetParathreadParams::get_parathread_params(para_id).ok_or(BuyingError::NotAParathread)?;
```

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -506,17 +597,45 @@ pub mod pallet {
...
+                return Err(BuyingError::NotAParathread);
+            };
+
+            let maybe_latest_author_info =
+                T::LatestAuthorInfoFetcher::get_latest_author_info(para_id);
+            if let Some(latest_author_info) = maybe_latest_author_info {
+                let current_slot = T::SlotBeacon::slot();
+                if !parathread_params.slot_frequency.should_parathread_buy_core(
+                    Slot::from(current_slot as u64),
+                    Slot::from(2u64),
```
girazoki, line 621:
```suggestion
                    Slot::from(2u64),
```

this should be a config parameter probably no?

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -276,33 +376,28 @@ pub mod pallet {
...
-            // want collators to pay fees.
             ensure_none(origin)?;
 
-            let assigned_collators = T::GetAssignedCollators::get_parathread_collators(para_id);
-            if assigned_collators.is_empty() {
-                return Err(Error::<T>::NoAssignedCollators.into());
-            }
-
-            if !assigned_collators.contains(&proof.account) {
+            let is_valid_collator = T::CheckCollatorValidity::is_valid_collator(
```
girazoki, line 388: arent we validating this in `validate_unsigned`?

girazoki:
I general I like the approach taking but I am missing a bunch of integration tests (with the xcm-emulator, if possible). Any other tests we can run @tmpolaczyk? I feel like any dev-test wont work so at most we can do the zombienet one you have already

pallets/xcm-core-buyer/src/lib.rs
girazoki: this should use the private key to sign, not the public one

tmpolaczyk:
> I general I like the approach taken but I am missing a bunch of integration tests (with the xcm-emulator, if possible). Any other tests we can run @tmpolaczyk? I feel like any dev-test wont work so at most we can do the zombienet one you have already.

We don't have any zombienet tests? I see the one in `test/suites/dev-tanssi/xcm-core-buyer/test_xcm_core_buyer.ts`, which is a dev test, and we can only use it to test that the xcm message has been sent, (because dev test = no relay chain), unless we mock the response somehow.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: @girazoki It uses private key from the keystore.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: yes. I have added to-do for that.

pallets/xcm-core-buyer/runtime-api/src/lib.rs
ParthDesai: What do you mean? I did ran `cargo fmt`

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: I only want to make the fields freely accessible for testing. Can I do that? 

Reason being is that if you mutate this struct, it will become invalid as either public key or signature changes.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: yes. Missed this. Will remove.

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: For types that can be deserialized from bytes, such as this one that implements `Decode`, you must assume they can be in an invalid state anyway, so I don't see much benefit from having private fields.

pallets/xcm-core-buyer/src/benchmarks.rs
ParthDesai: @girazoki Yes. The idea is to have some entries in storage. There is no tangible benefits except maybe slightly more accurate storage read/write timings. 

We can remove it if you want.

pallets/xcm-core-buyer/src/lib.rs
tmpolaczyk: Then the generic name is wrong? It should be `private_key: PrivateKey` I guess?

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: Oh. I did not meant it for validation. Verify function will return `false` in that case anyway.

My idea was to indicate anybody who is using this structure in non-test environment to be able to just call `new` and get valid instance which they cannot modify as any change there can render the struct invalid and call to fail. So, basically reduce the error surface for anybody working with the struct.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: No. This is the public key. The `KeyStoreExt` underneath the `verify` call, looks up the private key from keystone and signs with it.

pallets/xcm-core-buyer/src/lib.rs
ParthDesai: @tmpolaczyk Now, the fields are public. :)

tmpolaczyk: Cool, thank you

girazoki (approved):
I think it looks great now, great job! I have a few questions though:

- We have in our xcm-emulator tests tests that use `force_buy_core`. How difficult would it be to implement one that uses `buy_core`? I am not sure which collators do we set there but if they are collators that are backed by a private key (e.g., Alice) we could do this right?

- The zombienet parathread test version  can be changed once the node changes are implemented

ParthDesai:
> I think it looks great now, great job! I have a few questions though:
> 
> * We have in our xcm-emulator tests tests that use `force_buy_core`. How difficult would it be to implement one that uses `buy_core`? I am not sure which collators do we set there but if they are collators that are backed by a private key (e.g., Alice) we could do this right?
It is not much of effort to replace `force_buy_core` with `buy_core`. But there is no point as `buy_core` is equivalent to `force_buy_core` with extra validation for sender but has no effect on xcm execution.

> * The zombienet parathread test version  can be changed once the node changes are implemented
Sure.

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -640,44 +700,31 @@ pub mod pallet {
...
-                if <Pallet<T>>::is_online(heartbeat.authority_index) {
-                    // we already received a heartbeat for this authority
-                    return InvalidTransaction::Stale.into()
+            if let Call::buy_core {
+                para_id,
+                collator_account_id,
+                proof,
+            } = call
+            {
+                let block_number = <frame_system::Pallet<T>>::block_number();
```
tmpolaczyk, line 709: I added the block number to be able to do

```
.and_provides((block_number, para_id))
```

because without that, only a single unsigned transaction would ever be included by the block builder. Now that we have the nonce we probably don't need it. But I don't know a good alternative, `(para_id, nonce)` won't work because the nonce can be the same for different collators... So probably it's better to just leave it like it is. And unfortunately this is going to be very hard to test...

tmpolaczyk (approved):

PR #560, opened 2024-05-24
Add zombie upgrade tests containers

girazoki:
Adds new tests for zombienet upgrades in container-chains. This should extend our toolset and that of the container-chains to test the runtime upgrades. These are triggered by the CI but also can be triggered as:

- from the tests folder, running `pnpm moonwall test zombie_frontier_template_upgrade`
- or running `pnpm moonwall test zombie_simple_template_upgrade`

What the test does:

The test fetches the latest node associated with the latest runtime release and spawns a zombienet network composed of one relay, tanssi and a container. Once these are up, the container-chain is upgraded with the compiled runtime.


test/scripts/download-latest-rt-binaries.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# Exit on any error
+set -e
+
+# Always run the commands from the "test" dir
+cd $(dirname $0)/..
+
+LATEST_RUNTIME_RELEASE=$(curl -s https://api.github.com/repos/moondance-labs/tanssi/releases | jq -r '.[] | select(.name | test("runtime";"i")) | .tag_name' | head -n 1 | tr -d '[:blank:]') && [[ ! -z "${LATEST_RUNTIME_RELEASE}" ]]
```
tmpolaczyk, line 9: I bet this won't work because jq is not installed in the runner (but maybe I'm wrong)

girazoki: it should, it's the same we are doing in the CI through a script

girazoki: https://github.com/moondance-labs/tanssi/blob/a8336a40e4fe65e8d6f907002f7ef69e1885e9d4/.github/workflows/release.yml#L55

fgamundi (approved):

tmpolaczyk (approved):

PR #561, opened 2024-05-27
UPdate 1.11.0

girazoki:
TODOs:

- [X] By changing the `DealWithFees` struct (we were forced to) we now are losing the treasury.Deposit event (as we are using ResolveTo instead, something that also parity does;https://github.com/paritytech/polkadot-sdk/blob/4ab078d6754147ce731523292dd1882f8a7b5775/polkadot/runtime/common/src/impls.rs#L64. We should investigate if there is a way of maintaining still that event, although it is redundant (and always has be) because the balances.Deposit event as well

- [x] Write migrations for pallets `ForeignAssetCreator`, where we should migrate keys from V3 to V4.

- [X] Include migrations to execute on xcm-executor-utils & pallet-xcm

- [x] Fix all ts tests


runtime/dancebox/build.txt
```diff
@@ -0,0 +1,93 @@
+⚡ Found 3 strongly connected components which includes at least one cycle each
```
tmpolaczyk, line 1: remove

node/src/service.rs
```diff
@@ -89,7 +90,7 @@ type FullBackend = TFullBackend<Block>;
 pub struct ParachainNativeExecutor;
 
 impl sc_executor::NativeExecutionDispatch for ParachainNativeExecutor {
-    type ExtendHostFunctions = frame_benchmarking::benchmarking::HostFunctions;
+    type ExtendHostFunctions = ParachainHostFunctions;
```
tmpolaczyk, line 93: @girazoki do benchmarks work? Maybe this breaks it by removing the benchmarking host functions

girazoki: I have not tried, maybe we have to put a tuple here?

girazoki: Fixed, benchmarks working again

container-chains/nodes/frontier/src/rpc/mod.rs
```diff
@@ -355,7 +356,7 @@ where
 {
     // Frontier offchain DB task. Essential.
     // Maps emulated ethereum data to substrate native data.
-    match params.frontier_backend {
+    match &*params.frontier_backend {
```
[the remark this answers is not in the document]
tmpolaczyk: Probably to allow pattern matching on an `Arc`, see: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9e10ecf1d77ee0c47b30012e24f77982

(that example works with just `*` because `u8` implements `Copy`, otherwise you need `&*`)

container-chains/nodes/frontier/src/command.rs
```diff
@@ -246,10 +247,14 @@ pub fn run() -> Result<()> {
             match cmd {
                 BenchmarkCmd::Pallet(cmd) => {
                     if cfg!(feature = "runtime-benchmarks") {
-                        runner.sync_run(|config| cmd.run::<Block, ()>(config))
+                        runner.sync_run(|config| {
```
ParthDesai, line 250: Why did we change this?

container-chains/nodes/simple/src/command.rs
```diff
@@ -225,14 +226,19 @@ pub fn run() -> Result<()> {
...
         Some(Subcommand::Benchmark(cmd)) => {
             let runner = cli.create_runner(cmd)?;
+
             // Switch on the concrete benchmark sub-command-
             match cmd {
                 BenchmarkCmd::Pallet(cmd) => {
                     if cfg!(feature = "runtime-benchmarks") {
-                        runner.sync_run(|config| cmd.run::<Block, ()>(config))
+                        runner.sync_run(|config| {
+                            cmd.run_with_spec::<sp_runtime::traits::HashingFor<Block>, ReclaimHostFunctions>(Some(
```
ParthDesai, line 235: Just curious, Why did this change?

pallets/invulnerables/src/benchmarking.rs
```diff
@@ -23,7 +23,7 @@ use super::*;
 #[allow(unused)]
 use crate::Pallet as InvulnerablesPallet;
 use {
-    frame_benchmarking::{account, impl_benchmark_test_suite, v2::*, BenchmarkError},
```
ParthDesai, line 26: Why was this import removed? I do not see any other change in this file.

pallets/pooled-staking/src/benchmarking.rs
```diff
@@ -25,7 +25,7 @@ use {
         HoldReason,
         PendingOperationKey::{JoiningAutoCompounding, JoiningManualRewards},
     },
-    frame_benchmarking::{account, impl_benchmark_test_suite, v2::*, BenchmarkError},
```
ParthDesai, line 28: Same as above.

tmpolaczyk (approved):

runtime/dancebox/src/lib.rs
```diff
@@ -427,45 +438,45 @@ impl pallet_balances::Config for Runtime {
...
-impl<R> OnUnbalanced<NegativeImbalance<R>> for DealWithFees<R>
+impl<R> OnUnbalanced<Credit<R::AccountId, pallet_balances::Pallet<R>>> for DealWithFees<R>
 where
-    R: pallet_balances::Config + pallet_treasury::Config,
-    pallet_treasury::Pallet<R>: OnUnbalanced<NegativeImbalance<R>>,
+    R: pallet_balances::Config + pallet_treasury::Config + frame_system::Config,
+    pallet_treasury::NegativeImbalanceOf<R>: From<NegativeImbalance<R>>,
 {
     // this seems to be called for substrate-based transactions
-    fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance<R>>) {
```
ParthDesai, line 441: This seems to be typo.

runtime/dancebox/tests/common/xcm/core_buyer.rs
```diff
@@ -474,12 +476,16 @@ fn xcm_core_buyer_enough_balance_except_for_existential_deposit() {
...
                         spot_price,
+                        ordered_by,
                     }
                 ) => {
                     para_id: *para_id == ParaId::from(PARATHREAD_ID),
                     spot_price: *spot_price == spot_price2,
+                    ordered_by: *ordered_by == parathread_tank_in_relay,
                 },
-                RuntimeEvent::MessageQueue(pallet_message_queue::Event::Processed { success: true, .. }) => {},
+                // TODO: this now emits "success: false" even though the on demand order was placed, will
```
ParthDesai, line 486: I will need to investigate this.

pallets/invulnerables/src/benchmarking.rs
girazoki: not sure but I guess it was not needed (and the CI was telling me so). Maybe the benchmarking macros changed

container-chains/nodes/frontier/src/rpc/mod.rs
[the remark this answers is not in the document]
fgamundi: I see, need to learn more about Arcs. Thanks a lot!

runtime/dancebox/src/lib.rs
girazoki: which  part? I think it just made the line bigger and therefore fmt has decided to put it this way?

test/configs/zombieTemplateUpgradeSimple.json
```diff
@@ -35,10 +37,6 @@
             {
                 "name": "charlie",
                 "validator": true
```
ParthDesai, line 39: Why did we remove 4th node?

test/suites/common-container-chains/test-pallet-xcm-executor-utils/test-custom-policy.ts
```diff
@@ -58,13 +58,8 @@ describeSuite({
                         {
                             allowedAssets: [
                                 {
-                                    concrete: {
-                                        parents: 2,
-                                        interior: { Here: null },
-                                    },
```
ParthDesai, line 64: Why fungible assets were removed from allowed assets?

test/configs/zombieTemplateUpgradeSimple.json
girazoki: This is my bad, I will fix it

typescript-api/scripts/generate-types.ts
```diff
@@ -36,6 +36,7 @@ async function main() {
             "--dev-service",
             "--wasm-execution=interpreted-i-know-what-i-do",
             "--rpc-port=9933",
+            "--unsafe-force-node-key-generation"
```
ParthDesai, line 39: Is this new CLI option added? What does it do?

test/configs/zombieTemplateUpgradeSimple.json
girazoki: We have only dancebox + one container, so I think we can reduce it to 2. Do you want me to try this?

typescript-api/scripts/generate-types.ts
girazoki: it allows to generate automatically a node-key. Before this was done by default if you did not pass a key, now you need to explicitely say it

ParthDesai (approved):
Few questions. But apart from that looks good.

test/suites/common-container-chains/test-pallet-xcm-executor-utils/test-custom-policy.ts
girazoki: I think it is because now they are only identified by their Location

girazoki: Actually I think this test was wrong before, I am not sure why it was working. According to my knowledge the right type is a multilocation. @fgamundi do you know why we put a fungible there¿?

test/configs/zombieTemplateUpgradeSimple.json
ParthDesai: Generally, I am in favor of reducing complexity of test setup. So, yeah we can reduce it to 2. But, I would suggest to do it in another PR as this PR is already too big. :)

runtime/dancebox/src/lib.rs
ParthDesai: `on_unbalanceds`

girazoki: it's a trait so if there is a typo, its in parity

test/suites/common-container-chains/test-pallet-xcm-executor-utils/test-custom-policy.ts
fgamundi: Not sure why fungible was there. The right type is indeed a `MultiLocation` (`AssetId`, really). Must have gotten confused with the actual `XcmFragment`s in the test.
I just did some tests and seems like Polkadot.js just ignores any additional data that doesn't belong to the correct type, ignoring `fun` in this case

PR #563, opened 2024-05-29
Data Preservers Assignments

nanocryk:
Modifies the pallet to support assignments.

TODO in this PR:
- [x] Update runtime configs to support only free assignment for now
- [x] Benchmark the new extrinsics
- [x] Write migration from old Bootnodes storage to new Profiles + Assignements


pallets/data-preservers/src/lib.rs
```diff
@@ -62,6 +63,144 @@ use {
 pub mod pallet {
     use super::*;
 
+    #[cfg(feature = "std")]
+    pub trait SerdeIfStd: serde::Serialize + serde::de::DeserializeOwned {}
```
tmpolaczyk, line 67: I've seen this in polkadot-sdk, can we use that one instead?

https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/traits/trait.MaybeSerialize.html

pallets/data-preservers/src/lib.rs
```diff
@@ -62,6 +63,144 @@ use {
...
+        pub mode: ProfileMode,
+        pub assignment_request: ProviderRequestOf<T>,
+    }
+
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    #[derive(
+        RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, Encode, Decode, CloneNoBound, TypeInfo,
+    )]
+    #[scale_info(skip_type_params(T))]
+    pub enum ParaIdsFilter<T: Config> {
```
girazoki, line 110: Nice

pallets/data-preservers/src/lib.rs
```diff
@@ -550,6 +639,93 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            profile_id: T::ProfileId,
+            para_id: ParaId,
+            assigner_param: AssignerParameterOf<T>,
+        ) -> DispatchResultWithPostInfo {
+            let assigner = T::AssignmentOrigin::ensure_origin(origin, &para_id)?;
+
+            let mut profile = Profiles::<T>::get(profile_id).ok_or(Error::<T>::UnknownProfileId)?;
+
+            if profile.assignment.is_some() {
```
girazoki, line 655: This means one assignment per profile id right? If so let's make it clear by documenting it

girazoki (commented):
Everything looks good to me so far. Let's start configuring runtimes and migrations for this

pallets/data-preservers/src/lib.rs
```diff
@@ -550,6 +639,93 @@ pub mod pallet {
...
+                Err(Error::<T>::ProfileAlreadyAssigned)?
+            }
+
+            match &profile.profile.para_ids {
```
tmpolaczyk, line 659: Probably worth it to abstract this match into a function like `profile.profile.para_ids.can_assign(para_id)`

pallets/data-preservers/src/lib.rs
nanocryk: Tried to replace it with `MaybeSerializeDeserialize` but I can't get it to build :/ When creating the types with `cfg_attr(feature = "std", ...)` it tells me Serialize/Deserialize is not implemented (even without std enabled). I'd keep the current code for now and we can do a cleanup PR later (with other code cleanups, such as using `macro_rules_attribute` to remove all those annoyingly long derives)

tmpolaczyk: Ok no worries

pallets/data-preservers/src/lib.rs
```diff
@@ -47,55 +48,185 @@ use {
...
-            });
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    #[derive(
+        RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, Encode, Decode, CloneNoBound, TypeInfo,
+    )]
+    #[scale_info(skip_type_params(T))]
+    pub enum ParaIdsFilter<T: Config> {
+        AnyParaId,
+        Whitelist(BoundedVec<ParaId, T::MaxParaIdsVecLen>),
+        Blacklist(BoundedVec<ParaId, T::MaxParaIdsVecLen>),
```
tmpolaczyk, line 111: Just realized this could be a `BoundedBTreeSet` to speed up `.contains()`.

pallets/data-preservers/src/lib.rs
```diff
@@ -168,143 +318,53 @@ pub mod pallet {
...
         _,
         Blake2_128Concat,
         ParaId,
-        BoundedVec<BoundedVec<u8, T::MaxBootNodeUrlLen>, T::MaxBootNodes>,
+        BoundedVec<BoundedVec<u8, T::MaxNodeUrlLen>, T::MaxAssignmentsPerParaId>,
         ValueQuery,
     >;
 
     #[pallet::storage]
+    #[pallet::getter(fn profiles)]
```
tmpolaczyk, line 349:
```suggestion
```

Parity is deprecating getters, so we shouldn't add new ones

pallets/data-preservers/src/lib.rs
```diff
@@ -550,22 +613,132 @@ pub mod pallet {
...
+        }
+
+        #[pallet::call_index(8)]
+        #[pallet::weight(0)]
+        pub fn stop_assignment(
+            origin: OriginFor<T>,
+            profile_id: T::ProfileId,
+            para_id: ParaId,
+        ) -> DispatchResultWithPostInfo {
+            let assigner = T::AssignmentOrigin::ensure_origin(origin, &para_id)?;
```
tmpolaczyk, line 678: So the container chain manager is the one who can start and stop an assignment, how can the bootnode provider stop an assignment? I guess deleting their profile? Or closing the payment stream?

pallets/data-preservers/src/lib.rs
```diff
@@ -168,143 +318,53 @@ pub mod pallet {
...
-            Ok(deposit)
-        }
-    }
+    #[pallet::storage]
+    #[pallet::getter(fn assignments)]
+    pub type Assignments<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        ParaId,
+        BoundedVec<T::ProfileId, T::MaxAssignmentsPerParaId>,
```
tmpolaczyk, line 362: Also here
```suggestion
        BoundedBTreeSet<T::ProfileId, T::MaxAssignmentsPerParaId>,
```

pallets/data-preservers/src/lib.rs
```diff
@@ -550,22 +613,132 @@ pub mod pallet {
...
         pub fn check_valid_for_collating(para_id: ParaId) -> DispatchResult {
-            // To be able to call mark_valid_for_collating, a container chain must have bootnodes
-            if Pallet::<T>::boot_nodes(para_id).len() > 0 {
-                Ok(())
-            } else {
-                Err(Error::<T>::NoBootNodes.into())
+            if Self::assignments_profiles(para_id)
+                .filter(|profile| profile.mode == ProfileMode::Bootnode)
+                .count()
+                == 0
```
tmpolaczyk, line 736: To stop iterating after first match:

```suggestion
                .any(|profile| profile.mode == ProfileMode::Bootnode)
```

pallets/data-preservers/src/lib.rs
```diff
@@ -168,143 +318,53 @@ pub mod pallet {
...
+        WrongParaId,
+        MaxAssignmentsPerParaIdReached,
     }
 
     #[pallet::composite_enum]
     pub enum HoldReason {
         ProfileDeposit,
     }
 
+    #[deprecated]
```
tmpolaczyk, line 338: This makes CI error because of a deprecated warning... Can the storage item be removed now?

node/src/chain_spec/dancebox.rs
```diff
@@ -250,10 +246,6 @@ fn testnet_genesis(
         },
         parachain_system: Default::default(),
         configuration,
-        data_preservers: DataPreserversConfig {
-            para_id_boot_nodes,
-            ..Default::default()
-        },
```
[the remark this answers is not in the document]
nanocryk: Well there is no Bootnodes storage to assimilate now, and the replacement involves multiple coupled storages.

pallets/data-preservers/src/lib.rs
nanocryk: It is used in the migration.

pallets/data-preservers/src/lib.rs
girazoki: why this?

pallets/data-preservers/src/lib.rs
girazoki: We can make it not use it by quering directly by key, like here: https://github.com/moondance-labs/tanssi/blob/2fb1c0813d52c5c603a86ae46fec01545840e058/runtime/common/src/migrations.rs#L392

pallets/data-preservers/src/lib.rs
girazoki: let's make sure whenever we do this that the storages are public please

pallets/data-preservers/src/lib.rs
girazoki: Yes the profile owner should be able to do this. Also:

- on deassignation, we dont pay the profile-id for now (we will see if this is done in the future, but let's keep things simple). If you want to be paid prior to assignation, then call the extrinsic

- Also, yes let's disallow deleting any assigned profiles

pallets/data-preservers/src/lib.rs
nanocryk: In the migration we iterate over the keys, so it is more complex than accessing a single key.

pallets/data-preservers/src/lib.rs
tmpolaczyk: To speed up contains, also to make it easier to insert and delete items. Performance-wise shouldn't matter it we keep the number of para ids small, so I would do the one that has better ux. Which I believe is the set, but if I'm wrong then it's fine.

pallets/data-preservers/src/lib.rs
tmpolaczyk: I believe we did this in some migration in the past

pallets/data-preservers/src/lib.rs
girazoki: ALright I am fine with whatever, but I expect maybe oin ther order of 50-ish assignments per para max

pallets/data-preservers/src/lib.rs
girazoki: yes here an example @nanocryk https://github.com/moonbeam-foundation/moonbeam/blob/44e47fd9e9fa462329743c07db7fda96840eb32c/pallets/asset-manager/src/migrations.rs#L54

pallets/data-preservers/src/lib.rs
```diff
@@ -550,22 +612,133 @@ pub mod pallet {
...
+                .into_iter()
+                .filter_map(Profiles::<T>::get)
+                .map(|profile| profile.profile)
+        }
+
         /// Function that will be called when a container chain is deregistered. Cleans up all the storage related to this para_id.
         /// Cannot fail.
         pub fn para_deregistered(para_id: ParaId) {
-            BootNodes::<T>::remove(para_id);
+            Assignments::<T>::remove(para_id);
```
nanocryk, line 731: This however doesn't remove assignment in each profile.

girazoki: this is a good question in fact. `para_deregistered` can only happen in session changes right? here we use max_weight but depending on the amount of deregistration this can get quite messy

Because otherwise what is the alternative? list in the dapp anything that is either free or assigned to a de-registerd chain?

girazoki: I guess we could also make a non-paid extrinsic to clean it up..

tmpolaczyk: >para_deregistered can only happen in session changes right?

Almost always, there is an edge case when deregistering a pending chain, which is immediate, not sure if we allow pending chains to start an assignment... I guess yes because they should make sure to have bootnodes before calling `mark_valid_for_collating`.

> Because otherwise what is the alternative?

Having some cleanup extrinsic, or doing it using on_idle. I would probably delete the first 1 or 2 assigned profiles automatically, which should handle most use cases, and then use the lazy unassignment for the rest.

girazoki: the on_idle case is actually not a bad idea...but only if we have a onIdleUnassigment storage item where we push profile-ids

test/suites/para/test_tanssi_containers.ts
```diff
@@ -243,15 +243,22 @@ describeSuite({
...
+
+                const profileId = await paraApi.query.dataPreservers.nextProfileId();
+                const profileTx = paraApi.tx.dataPreservers.createProfile({
+                    url: "/ip4/127.0.0.1/tcp/33051/ws/p2p/12D3KooWSDsmAa7iFbHdQW4X8B2KbeRYPDLarK6EbevUSYfGkeQw",
+                    paraIds: "AnyParaId",
+                    mode: "Bootnode",
+                    assignmentRequest: "Free",
+                });
+
+                const tx3 = polkadotJs.tx.dataPreservers.startAssignment(profileId, 2002, "Free");
```
tmpolaczyk, line 255:
```suggestion
                const tx3 = paraApi.tx.dataPreservers.startAssignment(profileId, 2002, "Free");
```

In case you haven't figured it out already

.rustfmt.toml
```diff
@@ -1,6 +1,2 @@
-# Must be run in nightly:
-# cargo +nightly fmt
-
 reorder_imports = true
-max_width = 100
-imports_granularity = "One"
\ No newline at end of file
+max_width = 100
```
tmpolaczyk, line 2: Undo?

nanocryk: I removed  `imports_granularity = "One"` which requires nightly and is apparently only used by me haha

tmpolaczyk: Yeah sorry for not using it, but I still think it has some value, similar to clippy warnings which we fix once in a while. And if you want to remove it it's better to open a separate PR

pallets/data-preservers/src/lib.rs
nanocryk: For the `GenesisConfig` types need to implement serde traits even in no-std. However it seems it is no longer needed to feature-gate serde in our crate. We should then in a later PR cleanup serde usages to remove the feature-gates.

test/suites/one-node/test_tanssi_one_node.ts
```diff
@@ -190,11 +190,18 @@ async function registerEmptyParathread(api, manager, paraId) {
...
     );
-    const bootNodes = ["/ip4/127.0.0.1/tcp/33051/ws/p2p/12D3KooWSDsmAa7iFbHdQW4X8B2KbeRYPDLarK6EbevUSYfGkeQw"];
-    const tx2 = api.tx.dataPreservers.setBootNodes(paraId, bootNodes);
-    txs.push(tx2);
-    const tx3 = api.tx.registrar.markValidForCollating(paraId);
-    txs.push(tx3);
+
+    const profileId = await api.query.dataPreservers.nextProfileId();
+    txs.push(
+        api.tx.dataPreservers.createProfile({
```
tmpolaczyk, line 196: In all zombienet tests we should use `forceCreateProfile`, copy it from test/suites/parathreads/test_tanssi_parathreads.ts

This currently fails if you run

pnpm moonwall test zombie_tanssi_one_node

runtime/common/src/migrations.rs
tmpolaczyk, line 1: This file doesn't compile with --features=try-runtime but I believe it's also broken in master, add a

use sp_runtime::DispatchError

pallets/data-preservers/src/lib.rs
```diff
@@ -62,36 +61,243 @@ use {
...
+        pub assignment: Option<(ParaId, AssignmentWitnessOf<T>)>,
+    }
+
+    /// Computes the deposit cost of a profile.
+    pub trait ProfileDeposit<Profile, Balance> {
+        fn profile_deposit(profile: &Profile) -> Result<Balance, DispatchErrorWithPostInfo>;
+    }
+
+    /// Implementation of `ProfileDeposit` based on the size of the SCALE-encoding.
+    pub struct BytesProfileDeposit<BaseCost, ByteCost>(PhantomData<(BaseCost, ByteCost)>);
```
girazoki, line 173: This is fairly generic ( you only have Profile that requires an encode) I am surprised substrate does not provide anyhting like this alreay. But in any case perhaps this does not belong here but rather somewhere more generic, so that it can be re-used by other pallets. What do you think?

pallets/data-preservers/src/lib.rs
```diff
@@ -62,36 +61,243 @@ use {
...
+                .ok_or(ArithmeticError::Overflow)?
+                .checked_add(&base)
+                .ok_or(ArithmeticError::Overflow)?;
+
+            Ok(deposit)
+        }
+    }
+
+    /// Allows to process various kinds of payment options for assignments.
+    pub trait AssignmentPayment<AccountId> {
```
girazoki, line 202: Can we move it to a assignment.rs file within this crate? so that the pallet becomes a bit cleaner

pallets/data-preservers/src/lib.rs
girazoki: I found that substrate has something similar: https://github.com/paritytech/polkadot-sdk/blob/b65313e81465dd730e48d4ce00deb76922618375/substrate/frame/support/src/traits/storage.rs#L183

Essentially you can define a FootPrint structure with the `count` and `len` of profile

girazoki: We could even consider using `HoldConsideration` https://github.com/paritytech/polkadot-sdk/blob/b65313e81465dd730e48d4ce00deb76922618375/substrate/frame/support/src/traits/tokens/fungible/mod.rs#L245

runtime/common/src/migrations.rs
girazoki: yes please! thank you tomasz

test/suites/one-node/test_tanssi_one_node.ts
girazoki: it depends how we do it though, but in general yes, any batch call executed with sudo should do that

test/suites/common-tanssi/pallet-data-preservers/test_pallet_data_preservers.ts
```diff
@@ -37,12 +38,14 @@ describeSuite({
                 const storedProfile = await polkadotJs.query.dataPreservers.profiles(profileId);
                 expect(storedProfile.toJSON()).to.be.deep.equal({
                     account: general_user_bob.address,
-                    deposit: 10_190_000_000_000,
+                    deposit: 10_200_000_000_000,
```
girazoki, line 41: Can we add start_assignemnt and force_start_assignment as well as stop_assignment tets here?

pallets/data-preservers/src/mock.rs
```diff
@@ -219,17 +221,204 @@ where
...
+}
+
 impl pallet_data_preservers::Config for Test {
     type RuntimeEvent = RuntimeEvent;
     type RuntimeHoldReason = RuntimeHoldReason;
     type Currency = Balances;
     type ProfileId = u64;
+    type AssignmentPayment = AssignmentPayment;
+    type AssignmentOrigin = MockContainerChainManager<Test>;
     type SetBootNodesOrigin = MockContainerChainManagerOrRootOrigin<Test, EnsureRoot<AccountId>>;
```
ParthDesai, line 416: Is this still used?

pallets/data-preservers/src/lib.rs
```diff
@@ -120,22 +146,33 @@ pub mod pallet {
             + Debug
             + Eq
             + CheckedAdd
-            + One;
+            + One
+            + Ord;
 
         // Who can call set_boot_nodes?
         type SetBootNodesOrigin: EnsureOriginWithArg<Self::RuntimeOrigin, ParaId>;
```
ParthDesai, line 153: Is this still used?

test/suites/one-node/test_tanssi_one_node.ts
tmpolaczyk: It was only this test that was failing, I fixed it and pushed a commit

pallets/data-preservers/src/lib.rs
nanocryk: Nice catch, I'll remove it :)

girazoki (approved):

tmpolaczyk (approved):

PR #565, opened 2024-05-30
add weight to migrations for registrar

girazoki:
To better track the weight

### DIFF
--- a/runtime/common/src/migrations.rs
+++ b/runtime/common/src/migrations.rs
@@ -393,11 +393,14 @@ where
         )
         .unwrap_or_default();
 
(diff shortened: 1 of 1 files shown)


runtime/common/src/migrations.rs
```diff
@@ -393,11 +393,14 @@ where
         )
         .unwrap_or_default();
 
+        let total_weight = T::DbWeight::get().reads(1);
```
tmpolaczyk, line 396: `unhashed::take` does 1 read and 1 write

tmpolaczyk (approved):

PR #566, opened 2024-06-03
Simplify registrar benchmarks

tmpolaczyk:


pallets/registrar/src/benchmark_blob.rs
```diff
@@ -33,54 +33,53 @@ pub struct BenchmarkBlob {
...
-    128, 21, 69, 87, 39, 80, 189, 63, 244, 210, 244, 71, 86, 224, 172, 98, 5, 171, 30, 40, 194,
-    211, 222, 117, 47, 81, 150, 51, 191, 236, 129, 237, 184, 128, 223, 91, 79, 3, 86, 79, 50, 130,
-    149, 40, 193, 33, 8, 158, 35, 82, 137, 134, 195, 10, 60, 248, 193, 39, 236, 182, 225, 143, 110,
-    206, 62, 80, 169, 1, 159, 12, 182, 243, 110, 2, 122, 187, 32, 145, 207, 181, 17, 10, 181, 8,
-    127, 137, 0, 104, 95, 6, 21, 91, 60, 217, 168, 201, 229, 233, 162, 63, 213, 220, 19, 165, 237,
-    32, 0, 0, 0, 0, 0, 0, 0, 0, 104, 95, 8, 49, 108, 191, 143, 160, 218, 130, 42, 32, 172, 28, 85,
-    191, 27, 227, 32, 0, 0, 0, 0, 0, 0, 0, 0, 128, 254, 108, 203, 37, 75, 132, 240, 210, 32, 46,
-    181, 5, 189, 223, 159, 84, 203, 158, 189, 15, 178, 113, 144, 114, 233, 46, 229, 124, 29, 161,
-    216, 9,
+    0, 55, 133, 53, 79, 12, 10, 168, 38, 54, 66, 50, 252, 54, 227, 253, 113, 215, 29, 209, 8, 221,
```
tmpolaczyk, line 36: This has changed because previously I used `let account = ALICE;` below, but that doesn't work because of an invalid signature error when running:

```
cargo test -p pallet-registrar --features=runtime-benchmarks
```

I didn't notice before because this command is not being run in CI

pallets/registrar/src/benchmarks.rs
```diff
@@ -106,14 +147,15 @@ mod benchmarks {
...
-        }
-
-        let storage = new_genesis_data(data);
+    fn register_with_relay_proof(
+        x: Linear<100, 3_000_000>,
+        z: Linear<1, 10>,
+    ) -> Result<(), BenchmarkError> {
+        // This extrinsic is disabled in flashbox runtime, return 0 weight there
+        let _origin = T::RegisterWithRelayProofOrigin::try_successful_origin()
+            .map_err(|_| BenchmarkError::Weightless)?;
```
tmpolaczyk, line 156: I can't use this origin below because the signature is only valid for the account created by `create_funded_user` below, but if this returns an error it means that the extrinsic cannot be called, so we return a weight of zero.

pallets/registrar/src/benchmarks.rs
```diff
@@ -375,7 +447,7 @@ mod benchmarks {
         }
 
         // Second loop to fill RegisteredParaIds to its maximum
-        for k in 1000..(1000 + y) {
+        for k in 1000..(1000 + y - 1) {
```
tmpolaczyk, line 450: This `-1` is because if we will `RegisteredParaIds` to its maximum, the call below will fail because the list is full, so we need to leave space for 1 chain.

pallets/registrar/src/benchmarks.rs
girazoki: nice! then the benchmarks dont fail right?

pallets/registrar/src/benchmarks.rs
girazoki, line 450: is 1000 some sort of limit imposed by the runtime? if so can we fetch ir from there?

pallets/registrar/src/benchmarks.rs
tmpolaczyk: Correct, I have run them and there is a weight of 0 in flashbox, and non-zero in dancebox

girazoki (approved):

pallets/registrar/src/benchmarks.rs
```diff
@@ -197,10 +242,15 @@ mod benchmarks {
...
     #[benchmark]
-    fn deregister_scheduled(x: Linear<5, 3_000_000>, y: Linear<1, 50>) {
-        let storage = vec![(b"code".to_vec(), vec![1; x as usize]).into()];
-        let storage = new_genesis_data(storage);
-        let genesis_para_id_len = Pallet::<T>::registered_para_ids().len();
+    fn deregister_scheduled() {
+        let x = T::MaxGenesisDataSize::get();
+        let y = T::MaxLengthParaIds::get();
+        let storage = max_size_genesis_data::<T>(1, x);
+        let genesis_para_id_len = 0;
```
fgamundi, line 249: Is this needed?

pallets/registrar/src/benchmarks.rs
fgamundi: Maybe it's good to add this as a comment in the code?

tools/dev-benchmark-pallet.sh
fgamundi, line 1: Great script!

pallets/registrar/src/benchmarks.rs
tmpolaczyk: Forgot to remove it, this will always be 0

fgamundi (approved):

PR #568, opened 2024-06-04
Improve zombienet tests once more

tmpolaczyk:


tmpolaczyk:
* Sending the transaction again and again doesn't fix the issue, it always returns an error "invalid"
* "invalid" transactions may get included and finalized, but the RPC will return "invalid", even after the transaction has been finalized, and therefore must be valid. This happens because the tanssi RPC is not a full-node, but a collator. Changing zombienet to start a new full node for tanssi seems to fix this issue.

PR #570, opened 2024-06-04
Proxy precompile in Frontier container chain runtime

fgamundi:
Add Proxy precompiled smart contract from Moonkit to Frontier container chain at address `0x0000000000000000000000000000000000000805`

### DIFF
--- a/container-chains/runtime-templates/frontier/src/migrations.rs
+++ b/container-chains/runtime-templates/frontier/src/migrations.rs
@@ -33,6 +33,53 @@ use {
     sp_std::{marker::PhantomData, prelude::*},
 };
 
+pub struct MigratePrecompileProxyDummyCode<T>(pub PhantomData<T>);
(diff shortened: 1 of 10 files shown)


girazoki (approved):
LGTM

girazoki:
@fgamundi merge master back in please

container-chains/runtime-templates/frontier/src/migrations.rs
```diff
@@ -77,6 +77,53 @@ where
     }
 }
 
+pub struct MigratePrecompileProxyDummyCode<T>(pub PhantomData<T>);
```
tmpolaczyk, line 80: Maybe in the future we can have a generic migration in moonkit that we pass it the precompile address and it automatically writes the bytecode?

fgamundi: That's a neat idea. For sure let's do that for the next one

Agusrodri (approved):

PR #581, opened 2024-06-06
Add migration for ForeignAssetCreator

ParthDesai:


runtime/common/src/migrations.rs
```diff
@@ -564,6 +573,105 @@ where
...
+                &ForeignAssetToAssetId::<Runtime>::final_prefix(),
+                None,
+                removal_result.maybe_cursor.as_deref(),
+            );
+        }
+
+        // Write to the new storage with removed and added fields
+        for (asset_id, old_location) in asset_id_to_foreign_asset_data {
+            if let Ok(new_location) = Runtime::ForeignAsset::try_from(old_location) {
+                AssetIdToForeignAsset::<Runtime>::insert(asset_id, new_location);
```
girazoki, line 647: YOu can probably do both migrations here and not iterate twice

runtime/common/src/migrations.rs
```diff
@@ -564,6 +573,105 @@ where
...
+            } else {
+                log::warn!("Location could not be converted safely to xcmV4")
+            }
+        }
+
+        // One db read and one db write per element, plus the on-chain storage
+        Runtime::DbWeight::get()
+            .reads(migrated_count as u64)
+            .saturating_add(Runtime::DbWeight::get().writes(migrated_count as u64 + 3u64))
+            .saturating_add(Runtime::DbWeight::get().writes(foreign_asset_to_asset_id_count as u64))
```
girazoki, line 665: Dont you need to multiply by 2 here? (as you are writing twice per `foreign_asset_to_asset_id_count`

ParthDesai: The other write is calculated above in `migrated_count`

runtime/common/src/migrations.rs
ParthDesai: Hmm. But, that would mean I am assuming the semantics of these two storage items in runtime.

runtime/common/src/migrations.rs
```diff
@@ -564,6 +573,105 @@ where
...
+            None,
+            None,
+        );
+        while removal_result.maybe_cursor.is_some() {
+            removal_result = storage::unhashed::clear_prefix(
+                &ForeignAssetToAssetId::<Runtime>::final_prefix(),
+                None,
+                removal_result.maybe_cursor.as_deref(),
+            );
+        }
```
tmpolaczyk, line 642: This may not be needed because `storage_key_iter().drain()` above already removes from storage

ParthDesai: Oh. I did not know that drain also removes the item from storage.

tmpolaczyk: Me neither, I had to look at docs: https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/struct.PrefixIterator.html#method.drain

ParthDesai: Removed this call in next commit.

tmpolaczyk (commented):
Looks good so far!

runtime/common/src/migrations.rs
```diff
@@ -34,17 +34,22 @@
 //! This module acts as a registry where each migration is defined. Each migration should implement
 //! the "Migration" trait declared in the pallet-migrations crate.
 
+use frame_support::__private::log;
```
tmpolaczyk, line 37:
```suggestion
```

runtime/common/src/migrations.rs
```diff
@@ -564,6 +574,86 @@ where
...
+        for (old_location, asset_id) in foreign_asset_to_asset_id_data {
+            if let Ok(new_location) = Runtime::ForeignAsset::try_from(old_location) {
+                ForeignAssetToAssetId::<Runtime>::insert(new_location, asset_id);
+            } else {
+                log::warn!("Location could not be converted safely to xcmV4")
+            }
+        }
+
+        // One db read and one db write per element, plus the on-chain storage
+        Runtime::DbWeight::get().reads_writes(migrated_count as u64, 2 * migrated_count as u64)
```
tmpolaczyk, line 647: I see, 1 write when using `drain`, and another write when using `insert`...

girazoki (approved):

PR #585, opened 2024-06-10
Rename runtime-common to tanssi-runtime-common

akramhussein:
The `runtime-common` crate name conflicts with `polkadot-sdk:runtime-common`, therefore to make it easier to integrate more `polkadot-sdk` crates it should be renamed to `tanssi-runtime-common`


pnpm-lock.yaml
tmpolaczyk, line 1: Undo this

tmpolaczyk (approved):

PR #586, opened 2024-06-10
fix, hopefully for all, tests regarding consistency payment

girazoki:
Tested against `smoke-test-failure-85 `

### DIFF
--- a/test/suites/smoke-test-common/test-configuration-consistency.ts
+++ b/test/suites/smoke-test-common/test-configuration-consistency.ts
@@ -115,10 +115,12 @@ describeSuite({
                         // if currently assigned, then 2
                         let sessionRequirements: bigint;
 
-                        if(currentAuthorityAssignment["containerChains"][container.toString()].length == 0) {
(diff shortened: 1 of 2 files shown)


tmpolaczyk (approved):

test/suites/smoke-test-common/test-configuration-consistency.ts
```diff
@@ -105,16 +111,25 @@ describeSuite({
...
                     for (const container of containersToCompareAgainst) {
-                        // we should only check those who have enough credits
-                        // we compare against the credits they had just before new session obviously
-                        // that is when they were charged for tokens
+                        // if not currently assigned, then one session
+                        // if currently assigned, then 2
+                        let sessionRequirements: bigint;
+
+                        if (
+                            currentAuthorityAssignment["containerChains"][container.toString()] == null ||
```
tmpolaczyk, line 119: Technically this should be undefined instead of null, but this also works because undefined == null returns true

PR #588, opened 2024-06-12
Enable PoV reclaim in all runtimes

tmpolaczyk:
Allows to reclaim unused proof size. So if the benchmark says that an extrinsic will consume 1000 bytes because that's the worst case, but in reality it only consumes 10 bytes, the 990 bytes of difference will not count for the weight calculation and will be available for other extrinsics.

I followed this guide: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_pov_reclaim/index.html

To test locally:

```
export RUST_LOG=runtime::storage_reclaim=trace
pnpm moonwall run zombie_tanssi_one_node
```

And in polkadot js, send any extrinsic, for example dataPreservers.createProfile. You will see this in node logs:

```
2024-06-12 13:08:54.022 TRACE tokio-runtime-worker runtime::storage_reclaim: [Orchestrator] Reclaiming storage weight. benchmarked: 3694, consumed: 38 unspent: 0    
```

Also updated polkadot-sdk to fix problem with litep2p dependency not compiling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -441,7 +441,7 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "asset-test-utils"
 version = "7.0.0"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-v1.11.0#bd3dd974ac6bbb66ca339575edbe590ad9da0afb"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-v1.11.0#1bbb680f1a54d4568e4c00e344bccaa61e926434"
 dependencies = [
  "cumulus-pallet-parachain-system",
  "cumulus-pallet-xcmp-queue",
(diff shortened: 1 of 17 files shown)


tmpolaczyk:
I got this warning when generating the typescript-api:

```
2024-06-12 13:58:10        REGISTRY: Unknown signed extensions StorageWeightReclaim found, treating them as no-effect
```

I guess we need to update some packages?

tmpolaczyk:
I tried to add typescript tests but it turns out PoV reclaim is not enabled in dev mode. And I don't know how to enable it.

But I have verified that it works in zombienet. This is the block weight after sending a simple balance transfer, in dev tests (no PoV reclaim) and in zombienet (with PoV reclaim):

```
Dev tests:
{
  normal: { refTime: 297476000, proofSize: 165 },
  operational: { refTime: 0, proofSize: 0 },
  mandatory: { refTime: 7542446554, proofSize: 161004 }
}

Zombienet tests:
{
  normal: { refTime: 297476000, proofSize: 3593 },
  operational: { refTime: 0, proofSize: 0 },
  mandatory: { refTime: 8453106766, proofSize: 174362 }
}
```

Also I confirmed that proof size does not affect transaction fee, this is because our `WeightToFee` implementation ignores proof size.

girazoki (approved):

PR #590, opened 2024-06-13
Fix benches and new weights

girazoki:
Fix benchmarks and add new weights with respect to 1.11.0


pallets/registrar/src/benchmarks.rs
```diff
@@ -116,32 +116,16 @@ mod benchmarks {
 
     #[benchmark]
     fn register(x: Linear<100, 3_000_000>, z: Linear<1, 10>) {
-        let y = T::MaxLengthParaIds::get();
         let storage = max_size_genesis_data::<T>(z, x);
 
-        for i in 1..y {
```
ParthDesai, line 122: This was removed because we removed `y` right?

girazoki: it's mainly because it made sense before because the `pendingVerification` storage was a vector and not a mapping. And therefore, the cost of inserting a new one depended on the number of chains already existing.

@tmpolaczyk changed this storage item from a vector to a mapping. And therefore now it does not matter how many chains you have `pendingVerification`, so this parameter can be removed. And even better, speeds up benchmarks a lot

runtime/dancebox/src/lib.rs
```diff
@@ -2158,7 +2158,7 @@ impl_runtime_apis! {
                     // We only care for native asset until we support others
                     // TODO: refactor this case once other assets are supported
                     vec![Asset{
-                        id: AssetId(Location::here()),
```
ParthDesai, line 2161: Why was this changed?

test/suites/dev-frontier-template/test-precompiles/test-precompile-xcm-utils.ts
```diff
@@ -5,7 +5,7 @@ import { XcmVersionedXcm } from "@polkadot/types/lookup";
 import { u8aToHex } from "@polkadot/util";
 import { expectEVMResult, descendOriginFromAddress20 } from "../../../helpers";
 
-export const CLEAR_ORIGIN_WEIGHT = 2731000n;
```
ParthDesai, line 8: Changes due to weight re-generation right?

tmpolaczyk (approved):

runtime/dancebox/src/lib.rs
girazoki: For some reason it seems xcm benchmarks are now more properly tested and our configuration was wrong. We were putting in holding a token referenced by `Here` but we dont have any currency matching such multilocation. And then we were marking as `fee_asset` the native token. Again for some reason (I guess the benchmark upstream was not covering all cases) this was working before, but now it is not. Now the configuration should be correct

PR #592, opened 2024-06-14
Add XcmPaymentApi to all runtimes

tmpolaczyk:
https://github.com/paritytech/polkadot-sdk/pull/3607


container-chains/runtime-templates/simple/src/lib.rs
```diff
@@ -1125,6 +1129,37 @@ impl_runtime_apis! {
             SLOT_DURATION
         }
     }
+
+    impl xcm_fee_payment_runtime_api::XcmPaymentApi<Block> for Runtime {
+        fn query_acceptable_payment_assets(xcm_version: staging_xcm::Version) -> Result<Vec<VersionedAssetId>, XcmPaymentApiError> {
+            if !matches!(xcm_version, 3 | 4) {
```
girazoki, line 1135: here we need to return (if we can) whatever we have in the mapping of assetRate. assetRate defines asset_id to native asset-rates, so we first need to iterate what Ids we have there, then query ForeignAssetCreator to retrieve the ML with respet to an asset_id

container-chains/runtime-templates/simple/src/lib.rs
```diff
@@ -1125,6 +1129,37 @@ impl_runtime_apis! {
...
+                .collect())
+        }
+
+        fn query_weight_to_asset_fee(weight: Weight, asset: VersionedAssetId) -> Result<u128, XcmPaymentApiError> {
+            let local_asset = VersionedAssetId::V4(xcm_config::SelfReserve::get().into());
+            let asset = asset
+                .into_version(4)
+                .map_err(|_| XcmPaymentApiError::VersionedConversionFailed)?;
+
+            if asset != local_asset { return Err(XcmPaymentApiError::AssetNotFound); }
```
girazoki, line 1150: here we need to check if an asset-id exists for such an assetId in ForeignAsset. If it exists, we need to retrieve the asset-rate with such an assetId (if there is one).

container-chains/runtime-templates/simple/src/lib.rs
tmpolaczyk: Reading the keys of `pallet_asset_rate::ConversionRateToNative` I get an `AssetId` which is defined as `u16`, can I use that `u16` to create a `VersionedAssetId` directly? Or do I need the `ForeignAssetCreator`? Because this method doesn't need the multilocation

tmpolaczyk: Nevermind, I do need the location

runtime/dancebox/src/lib.rs
```diff
@@ -2607,6 +2611,62 @@ impl_runtime_apis! {
...
+    impl xcm_fee_payment_runtime_api::XcmPaymentApi<Block> for Runtime {
+        fn query_acceptable_payment_assets(xcm_version: staging_xcm::Version) -> Result<Vec<VersionedAssetId>, XcmPaymentApiError> {
+            if !matches!(xcm_version, 3 | 4) {
+                return Err(XcmPaymentApiError::UnhandledXcmVersion);
+            }
+
+            Ok([VersionedAssetId::V4(xcm_config::SelfReserve::get().into())]
+                .into_iter()
+                .chain(
+                    pallet_asset_rate::ConversionRateToNative::<Runtime>::iter_keys().filter_map(|asset_id_u16| {
```
girazoki, line 2624: why is this not implemented in the container-chain templates?

girazoki (approved):
Minor comment, but looks good.

runtime/dancebox/src/lib.rs
tmpolaczyk: Fixed

PR #597, opened 2024-06-18
Cleanup code

nanocryk:
- No longer require "std" feature to implement serde.
- Use a macro to reduce boilerplate with long derives (in a few places, we can remove this if you really don't like macros)
- Manually implement storage getters


primitives/traits/src/alias.rs
```diff
@@ -0,0 +1,108 @@
...
+            $crate::alias::__reexports::Serialize,
+            $crate::alias::__reexports::Deserialize,
+        )]
+        $($tt)*
+    }
+}
+
+// This currently doesn't work due to a quirk in RuntimeDebugNoBound, PartialEqNoBound
+// and CloneNoBound, as there seem to be something breaking macro hygiene. This is not an
+// issue when using the derive directly, but doesn't compile when adding it through our macro.
```
tmpolaczyk, line 64: What problems? I tried to use this macro in `pallets/data-preservers/src/types.rs` and it compiles fine:

```diff
-#[derive(RuntimeDebugNoBound, PartialEqNoBound, EqNoBound, CloneNoBound)]
-#[apply(derive_scale_and_serde)]
+#[apply(derive_storage_traits_no_bounds)]
 #[scale_info(skip_type_params(T))]
 pub enum ParaIdsFilter<T: Config> {
```

primitives/traits/src/alias.rs
```diff
@@ -0,0 +1,108 @@
...
+        #[derive(
+            $crate::alias::__reexports::RuntimeDebug,
+            ::core::cmp::PartialEq,
+            ::core::cmp::Eq,
+            ::core::clone::Clone,
+            $crate::alias::__reexports::Encode,
+            $crate::alias::__reexports::Decode,
+            $crate::alias::__reexports::TypeInfo,
+            $crate::alias::__reexports::Serialize,
+            $crate::alias::__reexports::Deserialize,
```
tmpolaczyk, line 41: Generally we don't need serde traits for storage, it's only needed if using a type in genesis, or in rpc. So I would rather see 2 derives above the type definition, one for storage traits and one for serde. And the serde alias is not that useful as its only 2 traits, it will probably look better if it's explicit.

container-chains/runtime-templates/frontier/src/lib.rs
```diff
@@ -572,9 +573,20 @@ impl pallet_utility::Config for Runtime {
...
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Encode,
+    Decode,
+    Debug,
+    MaxEncodedLen,
+    TypeInfo,
+    Serialize,
+    Deserialize,
```
tmpolaczyk, line 589: Actually we don't need serialize here, not sure why did we add that. For the other derives I don't care if they are unused, especially the ones from the standard library, but serde is an external crate so it would be nice to only use it when needed
```suggestion
```

primitives/traits/src/alias.rs
nanocryk: Really? I spent quite some time pulling my hair on this one, and it didn't worked even after doing a `cargo clean`

girazoki: were we able to solve this @nanocryk ?

tmpolaczyk: It turns out the macro is implemented wrong and it won't work in an alias, except in some cases such as the `ParaIdsFilter` because it is an enum, but for structs the `*NoBound` macros cannot be aliased

fgamundi (approved):

primitives/traits/src/alias.rs
girazoki: how important it is? do we use it?

girazoki: @nanocryk maybe we can create an issue to PR polkadot about the macro not being correctly implemented?

girazoki (approved):

tmpolaczyk (approved):

PR #598, opened 2024-06-18
Stream payment: Prevent making a request with a past deadline

nanocryk:
Prevents making a request with a mandatory change having a deadline in the past.
Also fix the `stream_payment_status` function to return `stalled = true` when deadline is reached.


girazoki:
Thank you @nanocryk , let's fix dev tests and merge

girazoki (approved):

tmpolaczyk (approved):

PR #599, opened 2024-06-19
add core buyer xcm emulator tests (non-forced)

girazoki:
Added to a different file in which now the core-buyer is a collator. Tests are identical

Includes a refactor to decide whether we want to test forced or non-forced manner


ParthDesai:
@girazoki If possible, can we modify `do_test` to either use `force_buy_core` or `buy_core` based on flag? Similarly we can have common test body with the flag and two wrapper around it basically one asking the test body to use `buy_core` and another one to use `force_buy_core`.

girazoki:
I made everything common but rather I  kept tests separate for now

runtime/dancebox/tests/common/xcm/core_buyer_common.rs
```diff
@@ -0,0 +1,367 @@
...
+    },
+    sp_core::Pair,
+    sp_runtime::{traits::ValidateUnsigned, AccountId32},
+    staging_xcm::v3::QueryId,
+    staging_xcm_executor::traits::ConvertLocation,
+    tp_traits::{ParaId, SlotFrequency},
+    xcm_emulator::{Chain, RelayChain},
+};
+
+pub(crate) const PARATHREAD_ID: u32 = 3333;
```
tmpolaczyk, line 49: I tend to avoid `pub(crate)`, `pub` also compiles, no?

girazoki: yes, let me do that

ParthDesai (approved):

ParthDesai (approved):

PR #606, opened 2024-06-26
Change Registrar balances reserves to holds and migration

fgamundi:
Change the balance deposit when registering a container chain from `Currency`'s reserve to `fungible`'s hold.
Migration to existing deposit also included


runtime/common/src/migrations.rs
```diff
@@ -454,6 +458,73 @@ where
...
+    T: pallet_balances::Config,
+    <T as pallet_balances::Config>::RuntimeHoldReason: From<pallet_registrar::HoldReason>,
+    <T as pallet_balances::Config>::Balance: From<
+        <<T as pallet_registrar::Config>::Currency as frame_support::traits::fungible::Inspect<
+            T::AccountId,
+        >>::Balance,
+    >,
+    <T as pallet_balances::Config>::Balance: From<u128>,
+{
+    fn friendly_name(&self) -> &str {
```
girazoki, line 474: do we know how many of these we have? did we try with try-runtime what is the weight used in, e.g., dancebox¿?

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
```diff
@@ -160,19 +159,11 @@ describeSuite({
...
+                expect(balanceAfterBob.free.toBigInt()).toEqual(
+                    balanceBeforeBob.free.toBigInt() + expectedDepositValue - fee
                 );
-                const bobDepositEvent = filtered.find(
-                    (event) => event[0] === bob.address && BigInt(event[1]) === expectedDepositValue
-                );
-                if (!bobDepositEvent) {
-                    console.log("deposit events: ", filtered);
-                }
-                expect(bobDepositEvent).to.not.be.undefined;
```
girazoki, line 174: why remove this? isnt there an event?

runtime/common/src/migrations.rs
```diff
@@ -454,6 +458,73 @@ where
     }
 }
 
+pub struct RegistrarReserveToHoldMigration<T>(pub PhantomData<T>);
```
girazoki, line 461: I think the migration looks good, can you add a rust test for this?

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
fgamundi: There isn't an event anymore, unfortunately (at least not that I could find)

runtime/common/src/migrations.rs
fgamundi:
```
Flashbox consumed weight: Weight(ref_time: 19450000000, proof_size: 0)
Dancebox consumed weight: Weight(ref_time: 11200000000, proof_size: 0)
```

girazoki: that's quite little, cool

test/suites/dev-tanssi/register-with-relay-proof/test_deregister_with_relay_proof.ts
girazoki: can we at least check that Balances::Holds contains the amount held?

pallets/registrar/src/lib.rs
```diff
@@ -673,15 +685,16 @@ pub mod pallet {
...
-                let (slashed_deposit, _not_slashed_deposit) =
-                    T::Currency::slash_reserved(&asset_info.creator, asset_info.deposit);
-                // _not_slashed_deposit can be greater than 0 if the account does not have enough reserved balance
-                // Should never happen but if it does, we just drop it
-                // TODO: is that the same as burning? Or as doing nothing? Either seems fine.
-
-                // Give deposit to origin account
                 // TODO: error handling
-                let _ = T::Currency::resolve_into_existing(&account, slashed_deposit);
+                let _ = T::Currency::transfer_on_hold(
```
girazoki, line 679: This is interesting, do we have a usecase for this? or a test? @tmpolaczyk do we have a test testing that if someone proves that you dont have ownership, the hold gets transferred to the prover?

pallets/registrar/src/lib.rs
```diff
@@ -731,8 +744,7 @@ pub mod pallet {
                 ) -> (T::AccountId, DepositBalanceOf<T>) {
                     const SEED: u32 = 0;
                     let user = account(string, n, SEED);
-                    T::Currency::make_free_balance_be(&user, total);
-                    let _ = T::Currency::issue(total);
+                    assert_ok!(T::Currency::mint_into(&user, total));
```
girazoki, line 737: Is this effectively the same?

runtime/common/src/migrations.rs
```diff
@@ -518,6 +593,7 @@ where
             Box::new(migrate_registrar_pending_verification),
```
girazoki, line 597: can you comment those migrations that have been applied?

girazoki:
The PR looks good to me but please let's add an integration test with the migration to be 100% sure.

@nanocryk can you review it since you have touched holds?

girazoki (approved):

pallets/registrar/src/lib.rs
nanocryk: Since we're creating a new user it's previous balance is zero, so both seems equivalent.
However calling 2 time `create_funded_user` with the same args will have a different behavior (the old one actually increasing the supply while the balance don't increase).

pallets/registrar/src/lib.rs
```diff
@@ -749,11 +760,10 @@ pub mod pallet {
             genesis_data: ContainerChainGenesisData<T::MaxLengthTokenSymbol>,
         ) -> DispatchResult {
             let deposit = T::DepositAmount::get();
-
-            // Verify we can reserve
-            T::Currency::can_reserve(&account, deposit)
-                .then_some(true)
-                .ok_or(Error::<T>::NotSufficientDeposit)?;
+            // Verify we can hold
```
nanocryk, line 763: The previous code also checked then reserve/hold, but why not hold directly?

nanocryk (approved):

pallets/registrar/src/lib.rs
tmpolaczyk: Yes, `can_deregister_with_empty_relay_state` and `can_deregister_pending_para` check that Bob is given Alice deposit

pallets/registrar/src/lib.rs
fgamundi: You're right, I guess we could hold directly and let it error if it can't succeed

PR #608, opened 2024-06-26
Refactor container_chain_spawner

### LINKED ISSUE #486: Error when removing db directory non-grafefuly
When we rotate the collator we somtimes see the following error:

`[Orchestrator] Shutdown error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })`

This is not a critical error, but it's a consequence of a non-graceful shutdown of the parity-db (which we do by removing the container-chain db folder). However it has no impact AFAICT.

I tried removing the logs before removing the db folder but no success. this is something we should investigate in the future

tmpolaczyk:
Several improvements to the spawn function based on what we see in testnet, and some refactors to make the code cleaner.
Enabling "hide whitespace" in files view makes this easier to review.

* Fix "shutdown error" log by removing database after closing it, instead of doing it while it is still open.
* Improve check for block diff to avoid unneeded restart of the container chain. Now instead of always restarting, we only restart if we need to remove the db. Since in most cases the db will be good, this reduces the time it takes to start a container chain by a few seconds (guess: 5s). Also change block diff number from 100 blocks to 30_000, to ensure we use full sync when it is more efficient than warp sync.
* Do not remove db if the block number is 0. This avoids a restart of the warp sync process for chains with big state. (untested)
* ~~Change select_sync_mode to return warp sync for an existing database. Even if we return warp sync it will still use full sync.~~ **undid this change**, see [1]
* Move `try_spawn` logic to a function instead of using a closure.
* Move params of spawn function to a new type `ContainerChainSpawnParams` to make cloning and passing to `try_state` easier.
* When checking container chain genesis hash, only check v0 if v1 does not match, this saves calculating an additional state root when the hash matches.
* Change `spawn` to be a regular `async fn` instead of an fn that returns a boxed future

Fix #486

Edit:

[1]: because warp sync has some bugs and can get stuck, we decided to default to full sync if a database exists. Combined with the change of "Do not remove db if the block number is 0", this means that now by default if warp sync fails to sync in 1 session, or if the node is stopped manually while the warp sync is in progress, when the chain restarts it will use full sync. By "warp sync" I mean the state sync part of warp sync, if the node is stopped while the block history download is in progress, nothing breaks.

Since using full sync can be very bad for chains with a big state, I added an error level log so that collators are aware that they may not be able to sync in time. They can retry a warp sync by stopping the node and manually removing the database, but if warp sync got stuck the first time it will probably also get stuck the second time. It's not clear why warp sync gets stuck but it seems related to bootnodes banning collators with error "Same state request multiple times". More investigation is needed.


client/node-common/src/command.rs
```diff
@@ -42,19 +42,5 @@ pub fn generate_genesis_block<Block: BlockT>(
...
-        <<Block as BlockT>::Header as HeaderT>::new(
-            Zero::zero(),
-            extrinsics_root,
-            state_root,
-            Default::default(),
-            Default::default(),
-        ),
-        Default::default(),
-    ))
+    Ok(construct_genesis_block(state_root, genesis_state_version))
```
girazoki, line 45: is this doing effectively the same thing?

node/src/container_chain_spawner.rs
```diff
@@ -58,11 +57,42 @@ use {
...
+/// Timeout to wait for the database to close before starting it again, used in `wait_for_paritydb_lock`.
+/// This is the max timeout, if the db is closed in 1 second then that function will only wait 1 second.
+const MAX_DB_RESTART_TIMEOUT: Duration = Duration::from_secs(60);
+
+/// Block diff threshold above which we decide it will be faster to delete the database and
+/// use warp sync, rather than using full sync to download a large number of blocks.
+/// This is only needed because warp sync does not support syncing from a state that is not
+/// genesis, it falls back to full sync in that case.
+/// 30_000 blocks = 50 hours at 6s/block.
+/// Assuming a syncing speed of 100 blocks per second, this will take 5 minutes to sync.
```
girazoki, line 69: is this assumption based of anything? I am glad to set this, we can tweak it through experimentation

tmpolaczyk: My main goal was to set it to so that it takes 5 minutes to sync, which is the session time in flashbox. More than that, and we get some downtime in flashbox, and less than that and we delete a potentially long chain only to download it again (because warp sync also downloads block history for now). But the 100 blocks per second is just a guess.

Also in the future we can make this value depend on the network (dancebox/flashbox) and on the size of the container chain if we find a way to estimate it.

client/node-common/src/command.rs
tmpolaczyk: Yes, I tried to refactor it further to avoid trying both v0 and v1 state, but I gave up because the way we do it now is probably faster. But I left this because it seemed duplicated code from polkadot-sdk.

node/src/container_chain_spawner.rs
```diff
@@ -674,9 +783,14 @@ fn handle_update_assignment_state_change(
     }
 }
 
-/// Select `SyncMode` to use for a container chain.
-/// We want to use warp sync unless the db still exists, or the block number is 0 (because of a warp sync bug in that case).
-/// The reason is that warp sync doesn't work if a database already exists, it falls back to full sync instead.
+/// Select [SyncMode] to use for a container chain.
+/// We want to use warp sync unless the container chain is still at genesis block (because of a warp sync bug in that case).
```
tmpolaczyk, line 787:
```suggestion
/// We want to use warp sync unless the db still exists, or the container chain is
/// still at genesis block (because of a warp sync bug in that case).
```

node/src/container_chain_spawner.rs
```diff
@@ -124,208 +144,177 @@ pub enum CcSpawnMsg {
...
-                    container_chain_para_id.into(),
-                    genesis_data,
-                    chain_type.clone(),
-                    relay_chain.clone(),
-                    boot_nodes,
-                )
-                .map_err(|e| format!("failed to create container chain chain spec from on chain genesis data: {}", e))?;
+// Separate function to allow using `?` to return a result, and also to avoid using `self` in an
+// async function. Mutable state should be written by locking `state`.
+// TODO: `state` should be an async mutex
```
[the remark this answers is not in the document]
tmpolaczyk: I know but this refactor is already big enough, we were using the sync mutex already and it works good enough. To ensure no deadlocks happen I make sure to lock the mutex for the shortest amount of time possible, and never hold it across await points (the latter case doesn't even compile because MutexGuard is not Send).

PR #612, opened 2024-07-02
Add support for parathread - part 3

ParthDesai:
# Description
This PR enables support for parathread on client side. When we detect that there is no core assigned for this para id, we check if it is parathread and if yes, we attempt to buy the core. 

To enable this I had to do some minor refactoring of node service:
- Propogation of orchestrator client and tx pool to send unsigned extrinsic
- Adding runtime api to query nonce for a para id 
- Refactoring  `CheckCollatorValidity` used to validate unsigned tx to make it simpler
- Minor refactoring

What is remaining:
- [x] Local testing
- [x]  Writing automated tests

Note that this PR's core is subject to change based on my findings in tests.


client/consensus/src/collators.rs
```diff
@@ -299,40 +299,42 @@ impl<Pub: Clone> SlotClaim<Pub> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum ClaimMode {
```
girazoki, line 303: nice

client/consensus/src/collators/lookahead.rs
```diff
@@ -74,8 +85,255 @@ use {
...
+        ClaimMode::ParathreadCoreBuying {
+            drift_permitted: buy_core_container_slot_drift.into(),
+        },
+        keystore,
+    )
+    .ok_or(BuyCoreError::UnableToClaimSlot)?;
+
+    let pubkey = slot_claim.author_pub;
+
+    orchestrator_runtime_api.is_core_buying_allowed(orchestrator_best_hash, para_id)??;
```
girazoki, line 299: I would pass the public key to `is_core_buying_allowed`. Just in case in the future we want to filter based of key

client/consensus/src/collators/lookahead.rs
```diff
@@ -342,6 +613,34 @@ where
...
+                                        tracing::trace!(target: crate::LOG_TARGET, ?block_hash, "Sent unsigned extrinsic to buy the core");
+                                    },
+                                    Err(buy_core_error) => {
+                                        buy_core_error.log_error(slot, params.para_id, relay_parent);
+                                    }
+                                };
+                                break; // No point in continuing as we need to wait for few relay blocks in order for our core to be available.
+                            },
+                            (Some(_slot_frequency), Some(core_index)) => {
+                                tracing::error!(target: crate::LOG_TARGET, ?core_index, "Not a parachain but we have a core (most likely from previous buy order)");
```
girazoki, line 639: I would not call this an error, as it sound like something that can happen

girazoki: I guess here we propose right?

client/consensus/src/collators/lookahead.rs
```diff
@@ -74,8 +85,255 @@ use {
...
+
+    let buy_core_slot_drift = orchestrator_client
+        .runtime_api()
+        .get_buy_core_slot_drift(orchestrator_best_hash)?;
+
+    // Convert drift in terms of container chain slots for parity between client side calculation and
+    // orchestrator runtime calculation
+    // We tried to made this calculation as generic as possible so that it can handle
+    // arbitrary slot timings as well and won't assume anything.
+    // Formula is: (Slot_drift_in_orchestrator_slot * orchestrator_slot_duration) / container_slot_duration
```
girazoki, line 276:
```suggestion
    // Formula is: (Slot_drift_in_orchestrator_slot * orchestrator_slot_duration) / container_slot_freq
```

runtime/dancebox/src/lib.rs
```diff
@@ -2576,10 +2584,29 @@ impl_runtime_apis! {
         }
     }
 
-    impl pallet_xcm_core_buyer_runtime_api::XCMCoreBuyerApi<Block, BlockNumber, ParaId> for Runtime {
+    impl pallet_xcm_core_buyer_runtime_api::XCMCoreBuyerApi<Block, BlockNumber, ParaId, NimbusId> for Runtime {
         fn is_core_buying_allowed(para_id: ParaId) -> Result<(), BuyingError<BlockNumber>> {
```
girazoki, line 2588: Here we should pass:
- slot
- NimbusId

client/consensus/src/collators/lookahead.rs
```diff
@@ -74,8 +85,255 @@ use {
...
+        .await
+        .map_err(BuyCoreError::TxSubmissionError)?
+        .fuse();
+
+    let current_time = tokio::time::Instant::now();
+    // Wait at max for 3 orchestrator slots
+    let wait_till = current_time + (orchestrator_slot_duration.as_duration() * 3);
+    loop {
+        select! {
+            _ = tokio::time::sleep_until(wait_till) => return Err(BuyCoreError::TxInclusionError),
```
tmpolaczyk, line 325: Why no just `tokio::time::sleep(timeout)`? With `timeout = orchestrator_slot_duration.as_duration() * 3` defined above.

client/consensus/src/collators/lookahead.rs
```diff
@@ -98,8 +356,26 @@ pub struct Params<BI, CIDP, Client, Backend, RClient, CHP, SO, Proposer, CS, GOH
...
+    SO,
+    Proposer,
+    CS,
+    GOH,
+    OClient,
+    OBlock,
+>(
+    mut params: Params<GSD, BI, CIDP, Client, Backend, RClient, CHP, SO, Proposer, CS, GOH>,
+    orchestrator_tx_pool: Arc<FullPool<OBlock, OClient>>,
+    orchestrator_client: Arc<OClient>,
```
tmpolaczyk, line 378: This function should only have 1 param, `params`, so move these two new params inside that struct

client/consensus/src/collators/lookahead.rs
```diff
@@ -342,6 +613,34 @@ where
                             Ok(h) => h,
                         };
 
+                        // TODO: Currently we use just the first core here, but for elastic scaling
+                        // we iterate and build on all of the cores returned.
+                        // More info: https://github.com/paritytech/polkadot-sdk/issues/1829
+                        let (is_parachain, core_index) = match (&aux_data.slot_freq, core_indices.first()) {
+                            (None, None) => break, // We are parachain and we do not have core allocated, nothing to do,
+                            (None, Some(core_index)) => (true, core_index), // We are parachain and we have core allocated, let's continue
```
tmpolaczyk, line 621: Love these comments, maybe we can turn them into `log::trace!` logs? So when we are trying to debug why a collator is not producing blocks, we can enable that log target.

client/consensus/src/collators/lookahead.rs
ParthDesai: I put those two fields outside `params` as it is too large with too many generics. 

Is there any specific reason why you would prefer to put those two new argument inside `params` struct?

tmpolaczyk: Yes, so that we can call using the syntax `run(Params { orchestrator_tx_pool, orchestrator_client })`, in a way that the order of the arguments doesn't matter. The reason `Params` structs exists is because the `run` function had too many arguments in the first place.

client/consensus/src/collators/lookahead.rs
ParthDesai: Problem is I want to only wait for 3 slot, no matter how many times the status of tx changes before it is either in block or finalized. 

If I do `tokio::time::sleep(timeout)` it will wait everytime the status changes for 3 slots and in a scenario where tx stays in a pool and does not make it to block, it will wait far too long.

client/consensus/src/collators/lookahead.rs
ParthDesai: Sure. Will do that.

client/consensus/src/collators/lookahead.rs
tmpolaczyk: Oh, I see, because of the outer `loop`. Can you add a comment about this?

client/consensus/src/collators/lookahead.rs
ParthDesai: yes. Will convert it to trace.

runtime/dancebox/src/lib.rs
ParthDesai: We get `slot` from `Slotbeacon` in runtime. So, only passing `NimbusId` is sufficient.

client/consensus/src/collators/lookahead.rs
ParthDesai: sure

client/consensus/src/collators/lookahead.rs
ParthDesai: Make sense. Will make changes.

tmpolaczyk:
I left the zombienet suite running overnight and 2 collators crashed with this backtrace:

```
Version: 0.8.0-53f3340954c

   0: sp_panic_handler::set::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic::{{closure}}
   3: std::sys_common::backtrace::__rust_end_short_backtrace
   4: std::panicking::begin_panic
   5: <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
   6: tc_consensus::collators::lookahead::run::{{closure}}
   7: <sc_service::task_manager::prometheus_future::PrometheusFuture<T> as core::future::future::Future>::poll
   8: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
   9: <tracing_futures::Instrumented<T> as core::future::future::Future>::poll
  10: tokio::runtime::task::core::Core<T,S>::poll
  11: tokio::runtime::task::harness::Harness<T,S>::poll
  12: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  13: tokio::runtime::scheduler::multi_thread::worker::Context::run
  14: tokio::runtime::context::scoped::Scoped<T>::set
  15: tokio::runtime::context::runtime::enter_runtime
  16: tokio::runtime::scheduler::multi_thread::worker::run
  17: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
  18: tokio::runtime::task::core::Core<T,S>::poll
  19: tokio::runtime::task::harness::Harness<T,S>::poll
  20: std::sys_common::backtrace::__rust_begin_short_backtrace
  21: core::ops::function::FnOnce::call_once{{vtable.shim}}
  22: std::sys::pal::unix::thread::Thread::new::thread_start
  23: start_thread
             at ./nptl/pthread_create.c:442:8
  24: __GI___clone3
             at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81


Thread 'tokio-runtime-worker' panicked at 'SelectNextSome polled after terminated', /home/tomasz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/stream/stream/select_next_some.rs:32
```

A few lines above I see this log, the collator seems stuck because of a low tx priority, and also it has 0 peers:

```
2024-07-09 19:30:12 [Container-2001] Unable to send buy core unsigned extrinsic through orchestrator tx pool relay_parent=0x7284781837492b809b351e77a75cf3c6d4bef1e7cd8b93e4ff04a6c8d6ccea32 para_id=Id(2001) slot=Slot(286757702) pool_error=Pool(TooLowPriority { old: 18446744073709551615, new: 18446744073709551615 })
```

I was testing on this commit:
53f3340954cb87860473397e49eec76200b65d69

I have more logs if needed

ParthDesai:
> I left the zombienet suite running overnight and 2 collators crashed with this backtrace:
> 
> ```
> Version: 0.8.0-53f3340954c
> 
>    0: sp_panic_handler::set::{{closure}}
>    1: std::panicking::rust_panic_with_hook
>    2: std::panicking::begin_panic::{{closure}}
>    3: std::sys_common::backtrace::__rust_end_short_backtrace
>    4: std::panicking::begin_panic
>    5: <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
>    6: tc_consensus::collators::lookahead::run::{{closure}}
>    7: <sc_service::task_manager::prometheus_future::PrometheusFuture<T> as core::future::future::Future>::poll
>    8: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
>    9: <tracing_futures::Instrumented<T> as core::future::future::Future>::poll
>   10: tokio::runtime::task::core::Core<T,S>::poll
>   11: tokio::runtime::task::harness::Harness<T,S>::poll
>   12: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
>   13: tokio::runtime::scheduler::multi_thread::worker::Context::run
>   14: tokio::runtime::context::scoped::Scoped<T>::set
>   15: tokio::runtime::context::runtime::enter_runtime
>   16: tokio::runtime::scheduler::multi_thread::worker::run
>   17: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
>   18: tokio::runtime::task::core::Core<T,S>::poll
>   19: tokio::runtime::task::harness::Harness<T,S>::poll
>   20: std::sys_common::backtrace::__rust_begin_short_backtrace
>   21: core::ops::function::FnOnce::call_once{{vtable.shim}}
>   22: std::sys::pal::unix::thread::Thread::new::thread_start
>   23: start_thread
>              at ./nptl/pthread_create.c:442:8
>   24: __GI___clone3
>              at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
> 
> 
> Thread 'tokio-runtime-worker' panicked at 'SelectNextSome polled after terminated', /home/tomasz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/futures-util-0.3.30/src/stream/stream/select_next_some.rs:32
> ```
> 
> A few lines above I see this log, the collator seems stuck because of a low tx priority, and also it has 0 peers:
> 
> ```
> 2024-07-09 19:30:12 [Container-2001] Unable to send buy core unsigned extrinsic through orchestrator tx pool relay_parent=0x7284781837492b809b351e77a75cf3c6d4bef1e7cd8b93e4ff04a6c8d6ccea32 para_id=Id(2001) slot=Slot(286757702) pool_error=Pool(TooLowPriority { old: 18446744073709551615, new: 18446744073709551615 })
> ```
> 
> I was testing on this commit: [53f3340](https://github.com/moondance-labs/tanssi/commit/53f3340954cb87860473397e49eec76200b65d69)
> 
> I have more logs if needed

I have identified and fixed the problem. 

The problem is that while reading from tx progress stream we are using `select_next_some` function which panics if the stream is closed. This is generally handled by the future library's `select!`  as it checks for stream termination.  But, tokio's `select!` does not handle this. Now, I am using `next` which will return `None` on fused stream which we will detect and return an error.

client/consensus/src/collators/lookahead.rs
```diff
@@ -74,8 +84,287 @@ use {
...
+    )?;
+
+    let mut tx_stream = orchestrator_tx_pool
+        .submit_and_watch(orchestrator_best_hash, TransactionSource::Local, *extrinsic)
+        .await
+        .map_err(BuyCoreError::TxSubmissionError)?
+        .fuse();
+
+    // We wait at max till 3 slots
+    let sleep = tokio::time::sleep(orchestrator_slot_duration.as_duration() * 3);
```
tmpolaczyk, line 335: Now that I think about it, this timeout here looks wrong. The loop in the `run` function gets triggered every time there is a new relay block. But this `try_to_buy_core` function will wait for up to 3 tanssi slots. So it can get stuck waiting for transaction inclusion, while another collator buys a core.

I guess my main concern is that this will cause us to miss relay block notifications... well actually can't it cause us to try to build blocks on top of an older relay block? Since `maybe_relay_parent_header` comes from a stream, and we don't have any logic to skip stale blocks, if this function sleeps then the new relay blocks will start to pile up.

Anyway, my proposal for `try_to_buy_core`:

* Have state about latest sent transaction, initialized to None
* If latest sent transaction is None, or it was sent more than 3 slots ago:
  * Log warning about why it was not included, send another buy core transaction, set state to Some(current_slot), and return immediately
* Else if transaction was sent less than 3 slots ago:
  * Return immediately, we will check if it's included the next time this function is called
* In the outer loop in the `run` function: if the parathread has a core available, set state of transaction to None

ParthDesai: Regarding your proposal, I think keeping a local state is not something I would prefer (unless there is no other way of course), as this could complicate things when there more than one collators per parathread and each collator's lookahead loop maintains a state. 

I had similar first iteration of this algorithm was where we are not waiting for the unsigned tx to be included in the block but I wanted bit of certainty that the tx was indeed included in the block before we do anything else. Maybe I can just revert to that and see if it makes any difference. 

Another solution that I am experimenting with is where we have another async loop that will try to buy core whenever it is eligible to do so and since it will run in separate loop, there is less chance of missing relay import notification or any other delay that is currently arising from waiting for tx. 

Will write results of these two approaches here and then we can decide.

tmpolaczyk: Keeping a local state is ugly but I think it's the simplest approach, I'm a bit scared of the idea of a separate loop but it could also work so yeah let's try.

test/suites/parathreads/test_tanssi_parathreads.ts
```diff
@@ -154,8 +280,9 @@ describeSuite({
             id: "T05",
             title: "Blocks are being produced on container 2001",
             test: async function () {
-                // Produces 1 block every 2 slots, which is every 24 seconds
-                await sleep(24000);
+                // Produces 1 block every 5 slots, which is every 60 seconds
+                // Give it a bit more time just in case
```
tmpolaczyk, line 284: This comment was outdated, slot time is defined in orchestrator chain slots, so 5 slots = 30 seconds.

test/suites/parathreads/test_tanssi_parathreads.ts
```diff
@@ -259,23 +386,26 @@ describeSuite({
                 expect(Number(await context.web3().eth.getBalance(randomAccount.address))).to.be.greaterThan(0);
             },
         });
+        /* Commented out as block generation frequency for parathread is
```
tmpolaczyk, line 389: From my tests it seems the block frequency is about 2x slower than expected because of slot time, so maybe we can add this test back but with a high margin of error.

test/suites/parathreads/test_tanssi_parathreads.ts
```diff
@@ -76,65 +74,193 @@ describeSuite({
...
+                        buyExecutionCost: 50_000_000_000,
+                        weightAtMost: {
+                            refTime: 1_000_000_000,
+                            proofSize: 100_000,
+                        },
+                    })
+                );
+                const setRelayChainTx = paraApi.tx.sudo.sudo(paraApi.tx.xcmCoreBuyer.setRelayChain("Rococo"));
+                await signAndSendAndInclude(
+                    paraApi.tx.sudo.sudo(paraApi.tx.utility.batch([tx, setRelayChainTx])),
```
tmpolaczyk, line 170: `signAndSendAndInclude` is a bit slow to call because it waits for the tx to be finalized, which can take 30-60 seconds, so if possible try to batch as many extrinsics as possible into one call. So instead of 2 separate calls to `signAndSendAndInclude`, move `tx4` here.

```suggestion
                    paraApi.tx.sudo.sudo(paraApi.tx.utility.batch([tx4, tx, setRelayChainTx])),
```

It doesn't matter much so if it breaks just keep it

tmpolaczyk (approved):

tmpolaczyk (approved):

girazoki (approved):
please lets check the tests failing :)

test/suites/parathreads/test_tanssi_parathreads.ts
```diff
@@ -322,7 +450,7 @@ async function assertSlotFrequency(blockData, expectedSlotDiff) {
     expect(
         Math.abs(avgSlotDiff - expectedSlotDiff),
         `Average slot time is different from expected: average ${avgSlotDiff}, expected ${expectedSlotDiff}`
-    ).to.be.lessThan(0.5);
+    ).to.be.lessThan(2.5);
```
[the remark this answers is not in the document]
ParthDesai: Already pushed the change for 3.  Let's see if that works out. If not will increase it to `3.5`

PR #613, opened 2024-07-02
Add pallet_registrar to Starlight as ContainerRegistrar

tmpolaczyk:


solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1703,6 +1707,140 @@ impl pallet_state_trie_migration::Config for Runtime {
     type MaxKeyLen = MigrationMaxKeyLen;
 }
 
+pub struct NoRelayStorageRoots;
```
girazoki, line 1710: I would implement this in tp_traits itself for (), what do you think?

tmpolaczyk: I would rather not, because that implementation will be wrong most of the time, it doesn't seem a good default value

fgamundi (approved):

PR #614, opened 2024-07-03
Starlight Collator & Authority Assignments

girazoki:
Adds the following pallets to Starlight runtime:

- TanssiInitializer
- TanssiInvulnerables
- TanssiCollatorAssignment
- TanssiAuthorityAssignment
- TanssiAuthorityMapping

Some integration tests related to the overall functioning of these pallets were added inside `collator_assignment_tests.rs` file.

This PR also adds a bool const to the config of both `pallet-configuration` and `pallet-collator-assignment` called `ForceEmptyOrchestrator`, which when enabled to `true`, indicates that no collators will be collating for any orchestrator chain while using Starlight runtime.


pallets/collator-assignment/src/lib.rs
```diff
@@ -103,6 +103,7 @@ pub mod pallet {
         type CollatorAssignmentHook: CollatorAssignmentHook<BalanceOf<Self>>;
         type Currency: Currency<Self::AccountId>;
         type CollatorAssignmentTip: CollatorAssignmentTip<BalanceOf<Self>>;
+        type AllowEmptyOrchestrator: Get<bool>;
```
girazoki, line 106: @tmpolaczyk what do you think about this?

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -2523,56 +2544,126 @@ pub struct OwnApplySession;
...
+    }
+}
+
+impl pallet_collator_assignment::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type HostConfiguration = HostConfigurationGetter;
+    type ContainerChains = ContainerChainsGetter;
+    type SessionIndex = u32;
+    type SelfParaId = MockParaId;
+    type ShouldRotateAllCollators = ();
```
girazoki, line 2651: For now these are all mocks until we have the proper pallets for them

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Ugly but works.

Perhaps I would change the variable to mean `DoNotAssignToOrchestrator` or `ForceEmptyOrchestrator`, with the default value of false being the same as before, but when set to true it ignores `min_collators_for_orchestrator` and `max_collators_for_orchestrator` and sets them to 0 instead, and disables the sanity check. To better indicate that not only we allow orchestrator to be empty, we force it to be empty.

girazoki: sounds good

solo-chains/runtime/starlight/src/genesis_config_presets.rs
```diff
@@ -212,15 +217,16 @@ fn starlight_testnet_genesis(
...
-                            x.5.clone(),
-                            x.6.clone(),
-                            x.7.clone(),
+                            x.babe.clone(),
+                            x.grandpa.clone(),
+                            x.para_validator.clone(),
+                            x.para_assignment.clone(),
+                            x.authority_discovery.clone(),
+                            x.beefy.clone(),
+                            x.nimbus.clone(),
```
tmpolaczyk, line 229: Nice! Haha

pallets/collator-assignment/src/lib.rs
Agusrodri: @tmpolaczyk done in a1663bdacd529b5057b29aa163b4732dfdd4da76 :)

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -2580,57 +2596,105 @@ sp_api::impl_runtime_apis! {
...
         CollatorConfiguration::initializer_on_new_session(&session_index);
-        // Next: Registrar
+        // 2. Second, registrar
         ContainerRegistrar::initializer_on_new_session(&session_index);
-        // // Next: AuthorityMapping
-        // AuthorityMapping::initializer_on_new_session(&session_index, &all_validators);
 
-        // let next_collators = queued.iter().map(|(k, _)| k.clone()).collect();
+        // 3. AuthorityMapping
+        TanssiAuthorityMapping::initializer_on_new_session(&session_index, &all_validators);
```
girazoki, line 2612: This is wrong, the authority mapping should only be done for collators, i.e. invulnerables. So let's fetch the invulnerable info first

Agusrodri: Fixed in 0e9953bc54d2eb955fda6b0031cffa6582fddc7e

solo-chains/runtime/starlight/tests/collator_assignment_tests.rs
```diff
@@ -0,0 +1,615 @@
...
+                vec![alice_keys.babe.clone(), bob_keys.babe.clone()]
+            );
+
+            run_to_session(2u32);
+            assert!(TanssiAuthorityMapping::authority_id_mapping(0).is_none());
+
+            let key_mapping_session_1 = TanssiAuthorityMapping::authority_id_mapping(1).unwrap();
+            assert_eq!(key_mapping_session_1.len(), 2);
+            assert_eq!(
+                key_mapping_session_1.get(&alice_keys_2.nimbus),
```
Agusrodri, line 502: Is this behavior expected? The test reflects that the new keys are placed inside the mapping for session 1, but shouldn't them be applied after session 2?

girazoki: solved

primitives/traits/src/lib.rs
```diff
@@ -219,6 +225,15 @@ pub trait RemoveInvulnerables<AccountId> {
     ) -> Vec<AccountId>;
 }
 
+impl<AccountId: Clone> RemoveInvulnerables<AccountId> for () {
+    fn remove_invulnerables(
+        collators: &mut Vec<AccountId>,
+        _num_invulnerables: usize,
+    ) -> Vec<AccountId> {
+        collators.clone()
```
tmpolaczyk, line 233: The `RemoveInvulnerables` is a bit awkward to use, but you are expected to remove items from the `collators` vec. So a better implementation would be to return `vec![]`, which would indicate that there are no invulnerables. That's not correct in starlight (because all collators are invulnerables) but since this trait is used to implement the invulnerables priority feature, it shouldn't matter.
```suggestion
        // Default impl: no collators are invulnerables
        vec![]
```

girazoki: So to me `RemoveInvulnerables = ()` should in fact indicate that I dont want any changes, you think it's not descriptive enough?

tmpolaczyk (approved):

PR #616, opened 2024-07-05
Remove unneeded migration pallets from starlight

tmpolaczyk:


solo-chains/runtime/starlight/Cargo.toml
```diff
@@ -84,7 +84,6 @@ pallet-scheduler = { workspace = true }
 pallet-session = { workspace = true }
 pallet-society = { workspace = true }
 pallet-staking = { workspace = true }
-pallet-state-trie-migration = { workspace = true }
```
girazoki, line 87: I feel like a few more should dissappear from this right?

tmpolaczyk: The identity migration is inside `runtime-common`, so no extra dependency for that one

girazoki (approved):

PR #618, opened 2024-07-08
Add author-noting to starlight and make it generic over para/relay

tmpolaczyk:
Adds a `RelayOrPara` config type that can be set to `ParaMode` or `RelayMode`. In relay mode we still use an inherent, but instead of passing the storage proof to the inherent, we pass `()` and read the storage directly using the `NativeStorageReader`.


pallets/author-noting/src/lib.rs
```diff
@@ -437,3 +428,91 @@ impl<T: Config> LatestAuthorInfoFetcher<T::AccountId> for Pallet<T> {
         LatestAuthor::<T>::get(para_id)
     }
 }
+
+pub struct RelayStuff;
```
girazoki, line 432: Very nice so far! I like it. A few nitpicks:

`ParaStuff` -> ParaStateReader

`RelayStuff` -> RelayStateReader

decouple `self_para_id` into a different trait, or for now even ignore it if you want

Move to primitives

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1861,6 +1863,46 @@ impl pallet_registrar::RegistrarHooks for StarlightRegistrarHooks {
     }
 }
 
+impl pallet_author_noting::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type ContainerChains = ContainerRegistrar;
+    type SelfParaId = ();
+    type SlotBeacon = dp_consensus::AuraDigestSlotBeacon<Runtime>;
+    type ContainerChainAuthor = TanssiCollatorAssignment;
+    type RelayChainStateProvider = NoRelayChainStateProvider;
```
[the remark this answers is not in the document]
tmpolaczyk: Yes, also `SelfParaId` was not used in this pallet, not sure why did we have it (copy-paste error?)

pallets/author-noting/src/benchmarks.rs
```diff
@@ -57,31 +57,45 @@ benchmarks! {
...
-            let num_each_container_chain = 2;
-            T::ContainerChainAuthor::set_authors_for_para_id(para_id, vec![author; num_each_container_chain]);
-            sproof_builder.num_items += 1;
-        }
-
-        let (root, proof) = sproof_builder.into_state_root_and_proof();
-
-        let data = tp_author_noting_inherent::OwnParachainInherentData {
-            relay_storage_proof: proof,
+        let data = if TypeId::of::<<<T as Config>::RelayOrPara as RelayOrPara>::InherentArg>() == TypeId::of::<tp_author_noting_inherent::OwnParachainInherentData>() {
```
girazoki, line 62: Can we create something like `InherentDataOf<T>`?

pallets/author-noting/src/benchmarks.rs
```diff
@@ -57,31 +57,45 @@ benchmarks! {
...
+            });
+
+            let arg = tp_author_noting_inherent::OwnParachainInherentData {
+                relay_storage_proof: proof,
+            };
+
+            *(Box::new(arg) as Box<dyn Any>).downcast().unwrap()
+        } else if TypeId::of::<<<T as Config>::RelayOrPara as RelayOrPara>::InherentArg>() == TypeId::of::<()>() {
+            // TODO: we need to write to storage the elements from the proof
+            unimplemented!("Benchmark will fail because headers have not been written to storage");
```
girazoki, line 90: Ok this should not be too difficult to do if we are able to configure this for a runtime that has `parachains_common` installed. else we directly put in storage with the HEAD_KEYS

pallets/author-noting/src/lib.rs
```diff
@@ -140,7 +138,7 @@ pub mod pallet {
         #[pallet::weight((T::WeightInfo::set_latest_author_data(<T::ContainerChains as GetCurrentContainerChains>::MaxContainerChains::get()), DispatchClass::Mandatory))]
         pub fn set_latest_author_data(
             origin: OriginFor<T>,
-            data: tp_author_noting_inherent::OwnParachainInherentData,
+            data: <<T as pallet::Config>::RelayOrPara as RelayOrPara>::InherentArg,
```
girazoki, line 141:
```suggestion
            data: InherentDataOf<T>,
```

pallets/author-noting/src/benchmarks.rs
tmpolaczyk: The main problem here is that the benchmark script cannot create proofs and so it needs to use the `mock_proof`, but the `mock_proof` only works when you need a storage proof. I could have added a second mock_proof or duplicated the logic to support relay mode here, but I won't be able to test it until we support benchmarks in starlight.

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1521,6 +1525,57 @@ impl pallet_registrar::RegistrarHooks for StarlightRegistrarHooks {
...
+    type AuthorNotingHook = ();
+    #[cfg(not(feature = "runtime-benchmarks"))]
+    type AuthorNotingHook = ();
+    // TODO: uncomment when pallets exist
+    //type AuthorNotingHook = (InflationRewards, ServicesPayment);
+    type RelayOrPara = pallet_author_noting::RelayMode;
+    type WeightInfo = pallet_author_noting::weights::SubstrateWeight<Runtime>;
+}
+
+pub struct NoRelayChainStateProvider;
```
girazoki, line 1557: We should move this to primitives.

Also there is still a public getter, I would assume if they remove the getter, they would make the storage public.

But in Addition, you can rely on this trait https://github.com/paritytech/polkadot-sdk/blob/02e50adf7ba6cc65a9ef5c332b3e2974c8d23f48/substrate/frame/system/src/lib.rs#L2187C17-L2187C36

tmpolaczyk: This was needed for `type RelayChainStateProvider = ...`, but since I moved that to be inside of `ParaMode`, it is no longer needed here, I can just remove it.

pallets/author-noting/src/benchmarks.rs
girazoki: Alright

girazoki (approved):
good job @tmpolaczyk

girazoki:
@tmpolaczyk maybe we can merge back master after you made it more consistent?

pallets/author-noting/src/benchmarks.rs
ParthDesai: Why are we comparing `TypeId`, would not be better to have a trait restriction for now with a note?

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1521,6 +1524,35 @@ impl pallet_registrar::RegistrarHooks for StarlightRegistrarHooks {
...
+    }
+}
+
+impl pallet_author_noting::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type ContainerChains = ContainerRegistrar;
+    type SlotBeacon = BabeSlotBeacon;
+    type ContainerChainAuthor = TanssiCollatorAssignment;
+    // We benchmark each hook individually, so for runtime-benchmarks this should be empty
+    #[cfg(feature = "runtime-benchmarks")]
```
ParthDesai, line 1546: Since the associated type definition is same whether `runtime-benchmarks` feature is enabled or not, can we combine them into one?

tmpolaczyk: Better not to, because when we add something to the hook we only want to add it under `#[cfg(not(feature = "runtime-benchmarks"))]`.

pallets/author-noting/src/benchmarks.rs
tmpolaczyk: Sorry I missed this, using TypeId is a hack but it was the fastest way I could think of to get this working. It is mainly used to get the code to compile. If I added a where bound forcing those 2 types to be equal this would not compile with the starlight runtime, only with dancebox and flashbox. And I didn't want to add a benchmark-only trait method like we usually do because this one depends on the mocked proof, so I would need to refactor that as well.

Anyway when we start running benchmarks for starlight we will need to think of a better way of handling the mock proof, so we will probably remove this hack as well.

ParthDesai (approved):

PR #621, opened 2024-07-09
Migrations pallets for Starlight

fgamundi:
Migration pallets for Starlight runtime.
Two different pallets have been added:
- Moonkit's migration pallet as `Migrations` to handle single block migrations
- Substrate's migrations pallet as `MultiBlockMigrations` to handle migrations that require multiple blocks to execute


Cargo.toml
```diff
@@ -145,6 +145,7 @@ pallet-assets = { git = "https://github.com/moondance-labs/polkadot-sdk", branch
 pallet-balances = { git = "https://github.com/moondance-labs/polkadot-sdk", branch = "tanssi-polkadot-v1.11.0", default-features = false }
 pallet-identity = { git = "https://github.com/moondance-labs/polkadot-sdk", branch = "tanssi-polkadot-v1.11.0", default-features = false }
 pallet-message-queue = { git = "https://github.com/moondance-labs/polkadot-sdk", branch = "tanssi-polkadot-v1.11.0", default-features = false }
+pallet-multiblock-migrations = { package = "pallet-migrations", git = "https://github.com/moondance-labs/polkadot-sdk", branch = "tanssi-polkadot-v1.11.0", default-features = false }
```
tmpolaczyk, line 148: Good job renaming it. So we now have 2 different "pallet-migrations" in Cargo.lock, one from moonkit and one from polkadot-sdk. It's amazing that this doesn't break everything, Rust is such a great language.

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1158,6 +1159,31 @@ impl pallet_configuration::Config for Runtime {
     type WeightInfo = ();
 }
 
+impl pallet_migrations::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type MigrationsList = ();
```
tmpolaczyk, line 1164: Even though we don't have any migrations yet, we probably want this to be
```suggestion
    type MigrationsList = (tanssi_runtime_common::migrations::StartlightMigrations<Runtime>,);
```

If possible that type should be defined in tanssi_runtime_common crate along DanceboxMigrations and FlashboxMigrations, but if that causes some dependency errors, just copy it to some file in this crate

Cargo.toml
fgamundi: Yeah.. it's far from ideal but indeed amazing that it works

tmpolaczyk (approved):

ParthDesai (approved):
Looks good.

PR #627, opened 2024-07-16
Refactor container chain spawning logic to not need require embeded node outside of collation

nanocryk:
Depends on https://github.com/moondance-labs/dancekit/pull/26

Currently the container chain spawning logic requires an embeded orchestrator client, which prevents using it in data preservers nodes that may interact with the orchestrator through a WebSocket connection. The linked PR modifies the `OrchestratorChainInterface` to add functions to get necessary data without the need for a client. When collating a client is still needed, as a data preserver node will not collate on the orchestrator chain.


pallets/registrar/src/lib.rs
```diff
@@ -77,7 +77,8 @@ pub mod pallet {
     #[derive(DefaultNoBound)]
     pub struct GenesisConfig<T: Config> {
         /// Para ids
-        pub para_ids: Vec<(ParaId, ContainerChainGenesisData<T::MaxLengthTokenSymbol>)>,
+        pub para_ids: Vec<(ParaId, ContainerChainGenesisData)>,
```
tmpolaczyk, line 80: You can remove `T::MaxLengthTokenSymbol` from config now, right?

So basically instead of having a configurable max length, we now fix it to 255 in the struct definition. Not ideal but I guess it will be good enough for our use case. There's an old TODO about adding bounds to the other vecs, if we follow this approach it will be easier to do, the alternative was to add like 5 generics or a custom trait to each struct, which looks quite ugly.

node/src/container_chain_spawner.rs
```diff
@@ -796,7 +810,7 @@ fn handle_update_assignment_state_change(
 /// interrupted before it finished downloading the state, in that case the node will use warp sync.
 /// If it was interrupted during the block history download, the node will use full sync but also
 /// finish the block history download in the background, even if sync mode is set to full sync.
-fn select_sync_mode(
+pub fn select_sync_mode_using_client(
```
tmpolaczyk, line 813: In this function we only need the `orchestrator_client` to check an edge case for when the container chain is still at block 0. We could also use the relay_chain_interface and get the latest finalized block from there, would that be better for the ws node?

girazoki: we could yes, in the end we are fetching that information from the relay itself

Cargo.toml
```diff
@@ -108,9 +108,11 @@ pallet-cc-authorities-noting = { git = "https://github.com/moondance-labs/dancek
 pallet-xcm-executor-utils = { git = "https://github.com/moondance-labs/dancekit", branch = "tanssi-polkadot-v1.11.0", default-features = false }
 test-relay-sproof-builder = { git = "https://github.com/moondance-labs/dancekit", branch = "tanssi-polkadot-v1.11.0", default-features = false }
 
+# dp-container-chain-genesis-data = { git = "https://github.com/moondance-labs/dancekit", branch = "tanssi-polkadot-v1.11.0", default-features = false}
+dp-container-chain-genesis-data = { path = "/home/nanocryk/Documents/Tanssi/dancekit/primitives/container-chain-genesis-data", default-features = false }
```
girazoki, line 112: You can now cherry-pick the change and tchange this

node/src/container_chain_spawner.rs
nanocryk: Which function in the relay interface should be called? Indeed that would allow to get rid of the generic and not need the orchestrator client.

tmpolaczyk: Probably `paras->heads`, but you would need to parse the header to get the block number. There are some utility functions in `pallet_author_noting`, such as `author_from_log`. @girazoki any better ideas? Maybe there is some other storage that only exists after the parachain has included its first block?

test/scripts/downloadChainSpec.ts
```diff
@@ -49,7 +49,7 @@ yargs(hideBin(process.argv))
                 }
                 process.stdout.write(`Done ✅\n`);
                 const onChainGenesisData = await api.createType(
-                    "TpContainerChainGenesisDataContainerChainGenesisData",
+                    "DpContainerChainGenesisDataContainerChainGenesisData",
```
tmpolaczyk, line 52: Remember to mention this in breaking changes, this will probably break all the registration scripts we have

node/src/container_chain_spawner.rs
nanocryk: Can we tackle that in a separate PR?

node/src/container_chain_spawner.rs
```diff
@@ -235,6 +246,8 @@ async fn try_spawn(
     );
 
     if !start_collation {
+        collation_params = None;
```
tmpolaczyk, line 249: I forgot why we used both `validator && start_collation` instead of just one boolean, here is a summary:

The reason `start_collation` exists as a separate arg to `validator` is that in the old days instead of restarting the container chain node we used `collate_on` to sent a message to enable the collator. So we would have to pass `collation_params` set to `Some` and `start_collation` set to false, and then use the `collate_on` closure to start collation later.

Now we don't have that `collate_on`, so just setting to `None` here is ok. Although I'm considering bringing `collate_on` back, for parathreads, but that's an issue for the future.

node/src/container_chain_spawner.rs
```diff
@@ -70,11 +69,21 @@ const MAX_DB_RESTART_TIMEOUT: Duration = Duration::from_secs(60);
 /// Assuming a syncing speed of 100 blocks per second, this will take 5 minutes to sync.
 const MAX_BLOCK_DIFF_FOR_FULL_SYNC: u32 = 30_000;
 
+pub trait TSelectSyncMode:
+    Send + Sync + Clone + 'static + (Fn(bool, ParaId) -> sc_service::error::Result<SyncMode>)
```
tmpolaczyk, line 73: Using `Fn()` as a trait bound is very ugly, promise to remove this trait as soon as possible in a future PR

tmpolaczyk:
> so indeed some registration script probably broke

Actually the registration seems to work fine, it's just the node that fails to read the genesis data for some reason

node/src/container_chain_spawner.rs
nanocryk: Haha, as we discussed we should be able to remove `SelectSyncMode` entirely :)

nanocryk:
It was due to the fact that I was storing an orchestrator block hash in the main structure itself, which is probably the genesis block in that test. Since 2002 is registered on the fly later, the spawner was not able to get the genesis data. I modified (again :skull: ) the chain interface to allow fetching the best or finalized head (the relay interface have that too), which should fix the issue.

Currently I use best head to do like previous code, but it might be better to use the finalized one?

tmpolaczyk:
> Currently I use best head to do like previous code, but it might be better to use the finalized one?

In practice it shouldn't matter because the values we read only change after 1 session, but I guess using finalized is more consistent with the `rx_loop` function, which reacts on finalized blocks (because otherwise it would need to unspawn a container chain in case of block revert).

girazoki:
> > Currently I use best head to do like previous code, but it might be better to use the finalized one?
> 
> In practice it shouldn't matter because the values we read only change after 1 session, but I guess using finalized is more consistent with the `rx_loop` function, which reacts on finalized blocks (because otherwise it would need to unspawn a container chain in case of block revert).

I did create an issue to change just this, but let's do it as a separate PR

node/src/container_chain_spawner.rs
girazoki: yes let's do it as a separate PR. But it would definitely make it more compatible with solo-chains

pallets/registrar/src/lib.rs
```diff
@@ -184,13 +182,8 @@ pub mod pallet {
...
     #[pallet::storage]
-    pub type ParaGenesisData<T: Config> = StorageMap<
-        _,
-        Blake2_128Concat,
-        ParaId,
-        ContainerChainGenesisData<T::MaxLengthTokenSymbol>,
-        OptionQuery,
-    >;
+    pub type ParaGenesisData<T: Config> =
+        StorageMap<_, Blake2_128Concat, ParaId, ContainerChainGenesisData, OptionQuery>;
```
girazoki, line 186: there is no migration here right? the values are the same

test/scripts/downloadChainSpec.ts
girazoki: why so?

pallets/registrar/src/lib.rs
nanocryk: I used the value in the runtime as the new constant limit.
Also IIRC the encoded data doesn't contain the limit.

tmpolaczyk: > the encoded data doesn't contain the limit

Correct, and it would only need a migration if you decrease the limit

test/scripts/downloadChainSpec.ts
girazoki: yes but I dont think we use it

girazoki: maybe the dapp does

node/src/container_chain_spawner.rs
girazoki: mayeb we should add a comment for this then

girazoki (approved):

node/src/container_chain_spawner.rs
tmpolaczyk: I can do that in the next refactor

tmpolaczyk (approved):

PR #629, opened 2024-07-17
Core assignation wired to collatorAssignment

girazoki:
Removes core-time traces and uses instead collator-assignation as a source of truth to assign cores. Currently the way it works is:
- parachains get the first `0..n-1` cores, where `n` is the number of parachains served for this session
- parathreads get the rest of the cores until `configuration.num_cores`

In order to seamlessly test this, I had to add the possibility of registering a parachain and invulnerables (collators) from genesis in the tanssi-registrar. Now you can see this works by running the appropriate zombienet suite.

Important: this only works because
- Parachain block production is not for now in charge of the tanssi-collators (ergo, we only care that the parachain has collators assigned in the runtime, nothing else)
- We are assigning Alice and Bob as invulnerables and these have already registered keys. **pallet-session does not allow to register session-keys that are not part of active first validator set on genesis.**

**Whenever we constrcut the solo-chain aware collator node, this will fail**. We will need an alternative solution for being able to inject the collator keys at genesis.

Todos for this PR:

[x] Rust integration tests


solo-chains/runtime/starlight/src/tests/core_scheduling_tests.rs
```diff
@@ -0,0 +1,438 @@
...
+    sp_std::{collections::btree_map::BTreeMap, vec},
+    starlight_runtime::{ContainerRegistrar, OnDemandAssignmentProvider, Paras, ParasSudoWrapper},
+    std::sync::Arc,
+    tp_traits::SlotFrequency,
+};
+
+mod common;
+
+const UNIT: Balance = 1_000_000_000_000_000_000;
+
```
tmpolaczyk, line 36:
```suggestion
```

Also add `mod core_scheduling_tests` alongside `mod session_keys` in tests/mod.rs

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -948,22 +947,102 @@ impl parachains_paras_inherent::Config for Runtime {
...
+            // Even if we did they would need their own collators to produce blocks, but for now
+            // I prefer to forbid.
+            // In this case the parathread would have bought the core for nothing
+            let assignment =
+                parachains_assigner_on_demand::Pallet::<Runtime>::pop_assignment_for_core(
+                    core_idx,
+                )?;
+            if assigned_collators
+                .container_chains
+                .contains_key(&assignment.para_id())
```
tmpolaczyk, line 988: I don't understand this check. If `assigned_paras.get` returns `None`, shouldn't we just return `Assignment::Pool`?

girazoki: the reason for this check is because people can buy (as of today) a core for a parathread that does not have collators assigned. Just for sanity purposes I would rather have assignments returned for just parathreads that have collators.

It is not going to change anything in the end, but at least is should avoid trying to contact non-existing collators for parathreads

girazoki: Maybe this does not play well with the fact that you can have your own collators. But I doubt you would want to use starlight in that case?

tmpolaczyk: It's weird that we do this after calling `pop_assignment_for_core(core_idx)`. Because if a parathread buys core number 0, it will be ignored forever? (as long as we have parachains, because assigned_paras.get(0) will return Some and never enter this branch)

So it feels to me that this way of handling the assignment is a bit too hacky. Does `parachains_assigner_on_demand` know that the first n cores are used by parachains?

girazoki: parathreads cannot buy core 0. in fact they do not know what core they buy. their assignment simply go to  a queue.

this queue is later handled by this code. this is for every non-parachain occupied core, we pop one of the assignments from that queue and they occupy the selected core

girazoki: Here the code: https://github.com/paritytech/polkadot-sdk/blob/6a5b6e03bfc8d0c6f5f05f3180313902c15aee84/polkadot/runtime/parachains/src/on_demand/mod.rs#L284

solo-chains/runtime/starlight/src/tests/core_scheduling_tests.rs
```diff
@@ -16,13 +16,18 @@
 
 #![cfg(test)]
 
+use crate::{ContainerRegistrar, ParasSudoWrapper};
```
girazoki, line 19: fmt

tmpolaczyk (approved):

solo-chains/runtime/starlight/src/lib.rs
tmpolaczyk: Ok so basically this is not an issue because any core affinity gets reset on session changes, as seen in test `test_on_demand_core_affinity_bound_to_core_gets_expired_at_session_boundaries`

ParthDesai (approved):

PR #630, opened 2024-07-17
XCM `DryRun` & `LocationToAccount` runtime APIs

fgamundi:
Adds `DryRun` and `LocationToAccount` XCM runtime APIs to Dancebox and the container chains runtimes


tmpolaczyk (approved):
Looks good. Is it worth it to add typescript tests?

girazoki:
I think it is yes, the only thing is that you will need to create the runtime-api interface yourself @fgamundi

tmpolaczyk:
Oh right... you have an example in `test/suites/common-xcm/xcm/test-xcm-payment-api.ts`

fgamundi:
Added TS tests

ParthDesai (approved):
Looks good.

girazoki (approved):

PR #631, opened 2024-07-18
Test relay parachain

girazoki:
Adds one parachain to produce blocks with appropriate collators (at least until we can assign then from tanssi)

Detected an error based of the order of the workers

### DIFF
--- a/test/configs/zombieStarlight.json
+++ b/test/configs/zombieStarlight.json
@@ -46,15 +46,45 @@
     },
     "parachains": [
         {
-            "id": 2000,
+            "id": 1000,
             "chain_spec_path": "specs/single-container-tanssi-1000.json",
             "collators": [
                 {
-                    "name": "FullNode-2000",
-                    "command": "../target/release/container-chain-simple-node",
+                    "name": "FullNode-1000",
+                    "command": "../target/release/tanssi-node",
                     "args": ["--no-hardware-benchmarks", "--database=paritydb", "--wasmtime-precompiled=wasm"],
-                    "ws_port": 9949,
+                    "ws_port": 9948,
                     "p2p_port": 33049
+                },
+                {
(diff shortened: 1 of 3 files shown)


test/configs/zombieStarlight.json
```diff
@@ -46,15 +46,45 @@
...
         {
-            "id": 2000,
+            "id": 1000,
             "chain_spec_path": "specs/single-container-tanssi-1000.json",
             "collators": [
                 {
-                    "name": "FullNode-2000",
-                    "command": "../target/release/container-chain-simple-node",
+                    "name": "FullNode-1000",
+                    "command": "../target/release/tanssi-node",
```
tmpolaczyk, line 54: Now I see "dancebox/800" as the chain id in port 9948, this needs to be 
```suggestion
                    "command": "../target/release/container-chain-simple-node",
```

test/configs/zombieStarlight.json
```diff
@@ -46,15 +46,45 @@
     },
     "parachains": [
         {
-            "id": 2000,
+            "id": 1000,
             "chain_spec_path": "specs/single-container-tanssi-1000.json",
```
tmpolaczyk, line 50:
```suggestion
            "chain_spec_path": "specs/single-container-template-container-2000.json",
```

test/configs/zombieStarlight.json
tmpolaczyk: Also this is a full node, those should be defined outside of the `collators` array

girazoki: no, I would not be able to create blocks if that is the case. Remember our node is not yet programmed to support starlight containbers, that is why I am directly putting dancebox as a container with their own collators

tmpolaczyk (approved):

PR #635, opened 2024-07-24
Add zombie_flashbox moonwall suite, not enabled in CI

tmpolaczyk:
Identical to the dancebox suite zombie_tanssi, the only difference is in `test/scripts/build-spec-flashbox.sh`, which has `--chain flashbox-local` instead of dancebox-local.

### DIFF
--- a/test/suites/para/test_tanssi_containers.ts
+++ b/test/suites/para/test_tanssi_containers.ts
@@ -36,10 +36,12 @@ describeSuite({
             const relayNetwork = relayApi.consts.system.version.specName.toString();
             expect(relayNetwork, "Relay API incorrect").to.contain("rococo");
 
-            const paraNetwork = paraApi.consts.system.version.specName.toString();
             const paraId1000 = (await paraApi.query.parachainInfo.parachainId()).toString();
-            expect(paraNetwork, "Para API incorrect").to.contain("dancebox");
             expect(paraId1000, "Para API incorrect").to.be.equal("1000");
+            const paraNetwork = paraApi.consts.system.version.specName.toString();
+            expect(paraNetwork, "Para API incorrect").to.satisfy(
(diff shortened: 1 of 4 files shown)


test/suites/para/test_tanssi_containers.ts
```diff
@@ -36,10 +36,12 @@ describeSuite({
             const relayNetwork = relayApi.consts.system.version.specName.toString();
             expect(relayNetwork, "Relay API incorrect").to.contain("rococo");
 
-            const paraNetwork = paraApi.consts.system.version.specName.toString();
             const paraId1000 = (await paraApi.query.parachainInfo.parachainId()).toString();
-            expect(paraNetwork, "Para API incorrect").to.contain("dancebox");
             expect(paraId1000, "Para API incorrect").to.be.equal("1000");
+            const paraNetwork = paraApi.consts.system.version.specName.toString();
```
girazoki, line 41: I dont think this suite is appropriate for flashbox, as it tests rotation which does not exist in flashbox. Maybe we should simply enable the suite without any test?

girazoki: Or am I wrong thinking that the suite tests rotation?

girazoki (approved):
Small comment to be answered

test/suites/para/test_tanssi_containers.ts
tmpolaczyk: Most importantly it tests registering and deregistering new chain. It also has a rotation happening in the middle of the test but that's mostly a coincidence. Since rotation is random there is a probability that the old assignment and the new assignment will be the same, so with that in mind this test should always pass in flashbox :)

girazoki: sounds good yeah! my main point was whether we test that the event thrown was `true` for rotation or something similar, but if not, then we are good to go

PR #636, opened 2024-07-24
Allow manual running of all zombienet tests in CI

tmpolaczyk:
Adds a new dispatchable workflow "Run zombienet tests" that can be used to run any tests in CI. The input test_name defaults to ".\*", but can be set to any regex like "zombie_tanssi.\*" or "zombie_flashbox". Zombienet upgrade tests are manually excluded because they don't work with the workflow template.

Also refactor zombienet test running logic into a new workflow template "zombienet-tests", this makes "release.yml" a bit cleaner.


.github/workflows/run-zombienet-tests.yml
```diff
@@ -0,0 +1,145 @@
+name: Manually run all zombienet tests
+
+on:
+  pull_request:
```
tmpolaczyk, line 4:
```suggestion
```

Remove this before merging, it's only for testing

girazoki (approved):

PR #637, opened 2024-07-25
Refactor Container chain spawner logic in separate crate

nanocryk:
Move logic related to starting embeded container chain node to a dedicated crate.
Adds a POC spawning of container chain node inside rpc provider command to ensure it can be called properly with available information.


client/service-container-chain/src/rpc.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// The client instance to use.
+    pub client: Arc<C>,
+    /// Transaction pool instance.
+    pub pool: Arc<P>,
+    /// Whether to deny unsafe calls
+    pub deny_unsafe: DenyUnsafe,
+    /// Manual seal command sink
+    pub command_sink: Option<futures::channel::mpsc::Sender<EngineCommand<Hash>>>,
+    /// Channels for manual xcm messages (downward, hrmp)
+    pub xcm_senders: Option<(flume::Sender<Vec<u8>>, flume::Sender<(ParaId, Vec<u8>)>)>,
```
girazoki, line 60: @ParthDesai this might be of interest for you, specially for the starlight manual seal

container-chains/nodes/simple/src/command.rs
```diff
@@ -297,20 +300,89 @@ pub fn run() -> Result<()> {
...
+
+                    let keystore_container = sc_service::KeystoreContainer::new(
+                        &sc_service::config::KeystoreConfig::InMemory,
+                    )?;
+                    let keystore = keystore_container.keystore();
+
+                    let para_id = cli.run.parachain_id.unwrap_or(2000);
+
+                    let (_cc_task_manager, _cc_client, _cc_backend) =
+                        tc_service_container_chain::service::start_node_impl_container(
```
girazoki, line 375: Question about this, we probably need to start this with very different arguments for frontier-like nodes anhd non-frontier-like nodes (specially related to the rpc-nodes). Are we going to be able to do so?

girazoki: Or maybe we can wire the ethereum ones to their own rpc-setup, but we for sure need to make sure not to spawn a non-ethereum registered chain in that case.

So `start_node_impl_container` is likely usable for substrate chains + collators but not for ethereum-nodes

container-chains/nodes/simple/src/command.rs
```diff
@@ -297,20 +300,89 @@ pub fn run() -> Result<()> {
...
+                            config,
+                            relay_chain_interface,
+                            client,
+                            keystore,
+                            ParaId::from(2001),
```
girazoki, line 380: Maybe we can ask the orchestrator chain about its own para-id somehow. Is there a runtime-api that we can query for this instead of hardcoding?

client/service-container-chain/src/service.rs
```diff
@@ -0,0 +1,499 @@
...
+        parachain_config.prometheus_registry(),
+        false,
+    )
+    .expect("function never fails");
+
+    (block_import, import_queue)
+}
+
+#[sc_tracing::logging::prefix_logs_with(container_log_str(para_id))]
+fn start_consensus_container(
```
girazoki, line 304: We are targeting this to be launched by both collators & non-collators right? If that is the case I think the first thing we should do is identify which things are needed by collators and which ones are not. One example is the `orchestrator_tx_pool`, which to my understanding is only needed in collation mode, as well as the `orhcestrator_para_id` or the `collator_key`

girazoki: All those things could be made optional and only pass them in collation-mode

girazoki: Sorry this was more for `start_node_impl_container`. But same logic applies. We are injecting orchestrator_para_id there when we dont need in `non-collation-mode`, so we probably should make it optional

client/service-container-chain/src/rpc.rs
```diff
@@ -0,0 +1,116 @@
...
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+//! A collection of node-specific RPC methods.
+//! Substrate provides the `sc-rpc` crate, which defines the core RPC layer
+//! used by Substrate nodes. This file extends those RPC definitions with
+//! capabilities that are specific to this project's runtime configuration.
+
+#![warn(missing_docs)]
+
+pub use sc_rpc::DenyUnsafe;
```
tmpolaczyk, line 24:
```suggestion
use sc_rpc::DenyUnsafe;
```

tmpolaczyk (approved):
This refactor is great, love it

girazoki (approved):
good job @nanocryk

PR #640, opened 2024-07-31
Increase authoring duration to 2s and block weight limit x4

tmpolaczyk:
Block authoring duration: 0.5s -> 2s
Frontier template gas limit: 15M -> 60M
Substrate block weight limit: 0.5e12 -> 2e12

https://github.com/paritytech/polkadot-sdk/pull/5195

### DIFF
--- a/client/service-container-chain/src/service.rs
+++ b/client/service-container-chain/src/service.rs
@@ -474,8 +474,7 @@ fn start_consensus_container(
         relay_chain_slot_duration,
         proposer,
         collator_service,
-        // Very limited proposal time.
-        authoring_duration: Duration::from_millis(500),
+        authoring_duration: Duration::from_millis(2000),
         para_backend: backend,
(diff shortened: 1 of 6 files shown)


girazoki:
Although before we do it, let's just see what other runtimes are doing

girazoki:
And here I have mixed opinions. I see that by default polkadot has 0.5s:

- https://github.com/paritytech/polkadot-sdk/blob/7d0aa89653d5073081a949eca1de2ca2d42a9e98/cumulus/parachains/common/src/lib.rs#L107

- however in the template, they have 2s: https://github.com/paritytech/polkadot-sdk/blob/7d0aa89653d5073081a949eca1de2ca2d42a9e98/substrate/bin/node/runtime/src/lib.rs#L214

tmpolaczyk:
Let's try and see if anything breaks

PR #642, opened 2024-07-31
Remove without_storage_info from pallets

tmpolaczyk:
Use pallet::unbounded to explicitly mark unbounded storage items. Most of those unbounded storage items are actually bounded but the bound depends on other fields so it is hard to tell to the compiler.

https://docs.substrate.io/reference/frame-macros/

> Note that you should only use the #[pallet::without_storage_info] macro if you need to make all of the storage items in your pallet unbounded. If you only need undefined storage for a specific storage item, you can use the #[pallet::unbounded] attribute macro to override the fixed size constraint.
>
> Because the #[pallet::without_storage_info] macro applies to all storage items in your pallet, you should only use it in a test or development environment. You should never use the #[pallet::without_storage_info] attribute macro in a production environment.

Note that parity is using `without_storage_info` for production pallets so not sure how important is this


girazoki (approved):
a bit scary, but as long as it does not require migrations..

tmpolaczyk:
>a bit scary, but as long as it does not require migrations..

Actually I don't know what exactly does this PR change. I guess metadata but I don't know if it's useful. We should probably check.

tmpolaczyk:
Looking at the macros this simply changes the derived trait from

https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.PartialStorageInfoTrait.html

to

https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.StorageInfoTrait.html

Both traits are identical except that "StorageInfoTrait" has a comment that says "It can be used to calculate PoV worst case size.". So the difference will be in some downstream component that uses this trait to calculate PoV stuff. I haven't found where, but it looks safe so I'll merge.

PR #643, opened 2024-08-01
Add manual seal support for starlight

ParthDesai:
# Description
This PR adds support for manual seal for starlight. It is activated when the development chain spec is used. 
To see what is changed in the code that builds the full node, please see module comment at `solo-chains/node/tanssi-relay-service/src/lib.rs`


solo-chains/node/tanssi-relay-service/src/lib.rs
```diff
@@ -14,6 +14,602 @@
...
+    let chain_name = chain_spec.name().to_string();
+    let genesis_hash = client
+        .hash(0)
+        .ok()
+        .flatten()
+        .expect("Genesis block exists; qed");
+    let properties = chain_spec.properties();
+
+    io.merge(ChainSpec::new(chain_name, genesis_hash, properties).into_rpc())?;
+    io.merge(StateMigration::new(client.clone(), backend.clone(), deny_unsafe).into_rpc())?;
```
tmpolaczyk, line 145: You can probably remove this because we removed the state migration pallet from starlight. And the Mmr from below is probably also not needed? If so you can remove the `mmr-rpc` dependency.

```suggestion
```

solo-chains/node/tanssi-relay-service/src/lib.rs
```diff
@@ -14,6 +14,602 @@
...
+                )
+            }
+            Sealing::Manual => {
+                let (sink, stream) = futures::channel::mpsc::channel(1000);
+                // We wait for slot duration here to make sure, we have strictly increasing slot numbers
+                // otherwise babe block import will reject the block.
+                // There might be more efficient way of doing this with stream throttling, but
+                // for development purposes it is okay.
+                let wrapped_stream = stream.then(move |item| {
+                    futures_time::task::sleep(slot_duration.as_duration().into())
```
tmpolaczyk, line 374: It's curious that this is not needed for aura, I guess because it doesn't enforce increasing slot numbers?

ParthDesai: Yes. Babe block import enforces this.

tmpolaczyk: I just remembered that we mock the timestamp for aura:

https://github.com/moondance-labs/tanssi/blob/dfb7b168dc20aae0690732d7fdd8f1a5b246dff6/node/src/service.rs#L107

So maybe you can do something similar here?

solo-chains/node/tanssi-relay-service/src/lib.rs
```diff
@@ -14,6 +14,602 @@
...
+            transaction_pool: transaction_pool.clone(),
+            spawn_handle: task_manager.spawn_handle(),
+            import_queue,
+            block_announce_validator_builder: None,
+            warp_sync_params: None,
+            block_relay: None,
+            metrics,
+        })?;
+
+    if config.offchain_worker.enabled {
```
girazoki, line 297: I dont think we need `offchain_worker` for dev service right?

solo-chains/node/tanssi-relay-service/src/lib.rs
girazoki: Yes let's do something similar

solo-chains/node/tanssi-relay-service/src/lib.rs
```diff
@@ -14,6 +14,602 @@
...
+pub type FullClient = service::TFullClient<
+    Block,
+    RuntimeApi,
+    WasmExecutor<(
+        sp_io::SubstrateHostFunctions,
+        frame_benchmarking::benchmarking::HostFunctions,
+    )>,
+>;
+
+pub struct NewFull {
```
girazoki, line 76: In order to have clarity over this, I would call all this NewDevFull, `create_dev_rpc_extension`, `build_dev_full`. I would even call this file dev_service.rs, to make it clear that we only modify the dev-service and not the actual service

solo-chains/node/tanssi-relay-service/src/lib.rs
ParthDesai: Offchain workers can be used by some pallets including Babe. So, better if we can enable it just so that we can fully test any pallet that uses off chain workers.

solo-chains/node/tanssi-relay-service/src/lib.rs
ParthDesai: I have moved it into `dev_service.rs` so it is now in a separate module itself. I did not rename it as it is redundant to say `dev_service::build_dev_full`.

solo-chains/node/tanssi-relay-service/src/lib.rs
girazoki: alright, thanks!

solo-chains/node/tanssi-relay-service/src/lib.rs
ParthDesai: Yes. I am now utilizing aux storage to store last timestamp and at minimum incrementing by `slot_duration`. 

Also added a typescript test to create bunch of blocks to test it and it seems to work fine.

girazoki: alright so is essentially the same but instead of using a local thread we use an storage item right? I dont see any caveats  doing that right @tmpolaczyk ?

girazoki (approved):
changes are good for me

solo-chains/node/tanssi-relay-service/src/lib.rs
ParthDesai: yes.

tmpolaczyk: Actually I prefer using storage items because they handle rollbacks automatically, but I guess thread_locals are cheaper to use. Either way is fine for tests.

test/suites/dev-tanssi-relay/test-basics.ts
```diff
@@ -0,0 +1,70 @@
+import "@polkadot/api-augment";
```
tmpolaczyk, line 1: How can I run this test?

tmpolaczyk: Nevermind found it:

```
pnpm moonwall test dev_tanssi_relay
```

tmpolaczyk (approved):

solo-chains/node/tanssi-relay-service/Cargo.toml
```diff
@@ -76,13 +77,16 @@ frame-system-rpc-runtime-api = { workspace = true }
 prometheus-endpoint = { workspace = true }
 
 # External Crates
+async-io = { workspace = true }
 async-trait = { workspace = true }
 bitvec = { workspace = true, optional = true }
 codec = { workspace = true }
 futures = { workspace = true }
+futures-time = { workspace = true }
```
tmpolaczyk, line 85:
```suggestion
```

PR #647, opened 2024-08-05
Fix and apply benchmarks

girazoki:
Updated benchmarks


container-chains/runtime-templates/frontier/src/weights/pallet_xcm_benchmarks_generic.rs
```diff
@@ -0,0 +1,348 @@
+// Copyright (C) Moondance Labs Ltd.
```
ParthDesai, line 1: Why this new file?

girazoki: good question, let me check

ParthDesai (approved):

PR #653, opened 2024-08-16
Starlight: Wire `ContainerRegistrar` and `Registrar` pallets

Agusrodri:


solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1484,6 +1492,60 @@ parameter_types! {
...
+        let kv_code = key_values
+            .into_iter()
+            .find(|kv| kv.0 == StorageWellKnownKeys::CODE.to_vec());
+
+        if let None = kv_code {
+            return Err(DispatchError::Other("Code not found"));
+        }
+
+        // Build ValidationCode
+        let validation_code = ValidationCode(kv_code.unwrap().1);
```
tmpolaczyk, line 1524: You can rewrite this to avoid unwrap

```rust
if let Some((_, code)) = kv_code {
    // Build ValidationCode
    let validation_code = ValidationCode(code);
    RegistrarManager::register(who, id, genesis_head, validation_code)
} else {
    return Err(DispatchError::Other("Code not found"));
}
```

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1484,6 +1492,60 @@ parameter_types! {
...
+{
+    fn register(
+        who: AccountId,
+        id: ParaId,
+        genesis_storage: Vec<ContainerChainGenesisDataItem>,
+    ) -> DispatchResult {
+        // Build HeadData
+        let key_values: Vec<(Vec<u8>, Vec<u8>)> =
+            genesis_storage.into_iter().map(|x| x.into()).collect();
+        let genesis_head = HeadData(key_values.clone().encode());
```
tmpolaczyk, line 1512: This HeadData looks wrong, it is supposed to be a block header, right? Or does the genesis head data include all the genesis storage on chain? `max_head_data_size` is configurable but the max value is 1MB, which is smaller than our limit of 5MB in `MaxGenesisDataSize`, so if genesis_head is the genesis state and not a hash, this can cause problems.

Also careful when cloning the genesis storage as it can be huge.

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: fixed

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: The genesis head is supposed to include all the genesis storage on chain yes, so if the limits are different, we may encounter problems there. I'll be thinking of some solutions.

Agusrodri: About the clone, at the end it's not necessary here. By removing it everything compiles well.

Agusrodri: _fixed_

pallets/registrar/src/lib.rs
```diff
@@ -821,6 +857,14 @@ pub mod pallet {
                 Self::deposit_event(Event::ParaIdDeregistered { para_id });
                 // Cleanup immediately
                 Self::cleanup_deregistered_para_id(para_id);
+                if let Err(id) = BufferedParasToDeregister::<T>::try_mutate(|v| v.try_push(para_id))
+                {
+                    log::error!(
+                        target: LOG_TARGET,
```
Agusrodri, line 863: @tmpolaczyk what do you think about this syntax? Would you suggest to use something else instead?

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1490,7 +1490,7 @@ parameter_types! {
     pub const DepositAmount: Balance = 100 * UNITS;
     #[derive(Clone)]
     pub const MaxLengthParaIds: u32 = 100u32;
-    pub const MaxEncodedGenesisDataSize: u32 = 5_000_000u32; // 5MB
+    pub const MaxEncodedGenesisDataSize: u32 = 1_000_000u32; // 1MB
```
tmpolaczyk, line 1493: Maybe add a comment that this should match `max_head_data_size` from the relay chain

pallets/registrar/src/lib.rs
```diff
@@ -1168,6 +1208,15 @@ pub mod pallet {
                         for para_id in new_paras {
                             Self::cleanup_deregistered_para_id(*para_id);
                             removed_para_ids.insert(*para_id);
+                            if let Err(id) =
+                                BufferedParasToDeregister::<T>::try_mutate(|v| v.try_push(*para_id))
+                            {
+                                log::error!(
+                                    target: LOG_TARGET,
+                                    "Failed to add paraId {:?} to deregistration list",
```
tmpolaczyk, line 1216: This one is ok, we cannot do anything if this fails.

solo-chains/runtime/starlight/src/tests/relay_registrar.rs
```diff
@@ -351,6 +350,13 @@ fn deregister_calls_schedule_para_cleanup() {
                 Configuration::set_max_head_data_size(root_origin(), 20500),
                 ()
             );
+
+            // Call end_block() to ensure that parachains_shared::CurrentSessionIndex
+            // storage gets updated properly inside on_finalize() and
+            // matches the one inside pallet_session::CurrentIndex.
+            end_block();
+            start_block();
```
tmpolaczyk, line 358: `run_to_session()` is doing exactly this internally so you can remove all the calls you added as they won't affect the test. If you need to produce an extra block somewhere, use `run_block()`, but you probably don't need it.

Agusrodri: Yes only need to use `end_block()` after the last `run_to_session()`. The previous behavior was needed when we had the call to `InnerRegistrar::deregister` inside `on_initialize()`, but now that we have everything inside `on_finalize()` we don't need it anymore.

Agusrodri: _fixed_

pallets/registrar/src/lib.rs
Agusrodri: _disregard_

pallets/registrar/src/lib.rs
Agusrodri: I changed it in 2f0a142be0c68c02c15fef5aa01c3f7de99f174a and now I'm returning a custom DispatchError (I needed to convert the internal error).

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: _fixed_

pallets/registrar/src/lib.rs
```diff
@@ -315,6 +336,11 @@ pub mod pallet {
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+            // Account for on_finalize weight
+            Weight::zero().saturating_add(T::DbWeight::get().reads(1))
```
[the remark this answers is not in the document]
Agusrodri: I added an approach in 427d0483a806ab6bca0bcaec4e6d510f33bb4c72 to account for all the possible N calls to `T::InnerRegistrar::deregister`. Let me know if it works :)

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: At the end the `HeadData` is supposed to be only the genesis header, not the whole state. Thus, we decided to introduce a new param `head_data` to associated `register` functions of `ContainerRegistrar` pallet. This is because otherwise, without this param it would be really difficult to calculate the needed `HeadData` object (and do it properly).

pallets/registrar/src/lib.rs
```diff
@@ -175,6 +176,8 @@ pub mod pallet {
 
         type RegistrarHooks: RegistrarHooks;
 
+        type InnerRegistrar: RegistrarHandler<Self::AccountId>;
```
girazoki, line 179: comment

pallets/registrar/src/lib.rs
```diff
@@ -320,6 +344,17 @@ pub mod pallet {
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+            // Account for on_finalize weight
+            let mut weight = Weight::zero().saturating_add(T::DbWeight::get().reads(1));
+
+            let buffered_paras = BufferedParasToDeregister::<T>::get();
+            for _para_id in buffered_paras {
+                weight += T::InnerRegistrar::deregister_weight();
```
girazoki, line 353: use buffered_para.len()

pallets/registrar/src/lib.rs
```diff
@@ -320,6 +344,17 @@ pub mod pallet {
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+            // Account for on_finalize weight
+            let mut weight = Weight::zero().saturating_add(T::DbWeight::get().reads(1));
```
girazoki, line 349: isnt take a write?

pallets/registrar/src/lib.rs
```diff
@@ -853,6 +918,7 @@ pub mod pallet {
                 })?;
                 // Mark this para id for cleanup later
                 Self::schedule_parachain_cleanup(para_id)?;
+                T::InnerRegistrar::schedule_para_downgrade(para_id)?;
```
girazoki, line 921: let's add comments anywhere here, for all the actions, the PR is complex enough and I would prefer that everything is documented

pallets/registrar/src/mock.rs
```diff
@@ -148,6 +148,7 @@ impl pallet_registrar::Config for Test {
     type DepositAmount = DepositAmount;
     type RuntimeHoldReason = RuntimeHoldReason;
     type RegistrarHooks = Mock;
+    type InnerRegistrar = ();
```
girazoki, line 151: maybe it would be worth setting this to a mock that we can test @Agusrodri , what do you think? something that updates a variable that we can verify through the test for instance

primitives/traits/src/lib.rs
```diff
@@ -384,3 +385,42 @@ impl GenericStorageReader for NativeStorageReader {
         }
     }
 }
+
+pub trait RegistrarHandler<AccountId> {
```
girazoki, line 389: Comment

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1604,6 +1609,73 @@ parameter_types! {
...
+            return RegistrarManager::make_parachain(id);
+        }
+        Ok(())
+    }
+
+    fn schedule_para_downgrade(id: ParaId) -> DispatchResult {
+        if !RegistrarManager::is_parathread(id) {
+            return RegistrarManager::make_parathread(id);
+        }
+        Ok(())
```
girazoki, line 1661: this means we are not erroring if we schedule a para downgrade for somehitng that is already a parathread right? better document it

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1619,6 +1691,9 @@ impl pallet_registrar::Config for Runtime {
     type DepositAmount = DepositAmount;
     type RegistrarHooks = StarlightRegistrarHooks;
     type RuntimeHoldReason = RuntimeHoldReason;
+    // TODO: replace TestWeightInfo when we use proper weights on paras_registrar
+    type InnerRegistrar =
+        InnerStarlightRegistrar<AccountId, Registrar, paras_registrar::TestWeightInfo>;
     type WeightInfo = pallet_registrar::weights::SubstrateWeight<Runtime>;
 }
```
girazoki, line 1699: Can we create an issue to prevent using the relay registrar with dispatch filters? does not have to be done in this PR

solo-chains/runtime/starlight/src/tests/collator_assignment_tests.rs
```diff
@@ -714,8 +735,22 @@ fn test_paras_registered_but_zero_credits() {
...
+            assert!(assignment.container_chains.is_empty());
+
+            // We need to accept the validation code, so that the para is onboarded after 2 sessions.
+            assert_ok!(Paras::add_trusted_validation_code(
+                root_origin(),
+                get_genesis_data_with_validation_code().1.into()
             ));
+            run_to_session(4);
+            let assignment = TanssiCollatorAssignment::collator_container_chain();
+            assert!(assignment.container_chains.is_empty());
```
girazoki, line 753: I am curious to see, what happens if we register a container-chain that has an non-valid wasm? does it get collators assigned after 2 sessions?

I guess it's no problem, as it will pay for those so there is a clear disincentive for that, but can we add a test? meaning we register but we never add the `trusted_validation_code`

solo-chains/runtime/starlight/src/tests/core_scheduling_tests.rs
```diff
@@ -508,6 +509,13 @@ fn test_parathread_uses_0_and_then_1_after_parachain_onboarded() {
             // this is, 1000 will go first, then 1001. Since we want 1000 to use core 0,
             // the only way to achieve this is by assigning the parathread a higher para-id
             run_to_block(2);
+
+            // Coordinate both ParasShared and Session CurrentIndex storages.
+            // If we don't do this, the one inside ParasShared is configured to
+            // one session before, and we want the to be the same for later checking in
+            // session 6.
+            ParasShared::set_session_index(Session::current_index());
```
girazoki, line 517: wait this should happen somewhere on_finalize or on_initialize no?

pallets/registrar/src/lib.rs
Agusrodri: fixed in 9a76e31422088f5a1231e8671063cb0ed3da1581.

pallets/registrar/src/lib.rs
Agusrodri: fix included in 9a76e31422088f5a1231e8671063cb0ed3da1581 as well.

pallets/registrar/src/lib.rs
Agusrodri: Docs added in 9a76e31422088f5a1231e8671063cb0ed3da1581 :)

primitives/traits/src/lib.rs
Agusrodri: Docs added for this in 9a76e31422088f5a1231e8671063cb0ed3da1581.

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: Done in 9a76e31422088f5a1231e8671063cb0ed3da1581.

pallets/registrar/src/lib.rs
Agusrodri: From what I understand, as the `BufferedParasToDeregister` storage is set inside `on_initialize` and then drained inside `on_finalize`, we don't need to account for an extra write, but maybe @tmpolaczyk can confirm.

pallets/registrar/src/mock.rs
Agusrodri: Mmm not sure how could we do that to be honest. I set the empty tuple here as this mock is supposed to reflect the standard behavior of the pallet (as its usage in dancebox and flashbox).

In case we would like this type to be dynamic, we could try introducing the `pallet_parameters` to the mock somehow, but still not sure about the implementation. 

If we want to investigate that, I think it would worth doing it in a separate PR, as depending on the final implementation, we could end up having a lot of extra code. What do you think?

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: Sure, issue created https://opslayer.atlassian.net/browse/MD-517?atlOrigin=eyJpIjoiYjlmYjkxMmZlNzI4NDJmZmFmMDkwM2UxMTdkZDliYTUiLCJwIjoiaiJ9

solo-chains/runtime/starlight/src/tests/collator_assignment_tests.rs
Agusrodri: Sure, I added a test in cf97156e020b5066cefd0db772b3913ff3e98966. 

The result shows that no collators are assigned to the container chain. This is because if we don't mark the validation code as trusted, when we call `mark_valid_for_collating`, it will simply fail with `NotParathread` error (as the paraId has not yet been onboarded and thus `mark_valid_for_collating` cannot upgrade it to a parachain in the relay context).

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1604,6 +1609,73 @@ parameter_types! {
...
+    fn register(
+        who: AccountId,
+        id: ParaId,
+        genesis_storage: Vec<ContainerChainGenesisDataItem>,
+        head_data: Option<HeadData>,
+    ) -> DispatchResult {
+        // Return early if head_data is not specified
+        let genesis_head = match head_data {
+            Some(data) => data,
+            None => return Err(DispatchError::Other("HeadData not specified!")),
```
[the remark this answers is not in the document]
Agusrodri: Yes we can do it. I didn't take this approach initially because I wanted the `ContainerRegistrar` pallet to be as isolated as possible from any possible implementation of `InnerRegistrar`, but I'm also fine with adding them.

Applied in 1da160133ce626b398e73c7672930540e8662fa1.

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1604,6 +1609,73 @@ parameter_types! {
...
+        let key_values: Vec<(Vec<u8>, Vec<u8>)> =
+            genesis_storage.into_iter().map(Into::into).collect();
+
+        // Check if the wasm code is present in storage
+        let validation_code = match key_values
+            .into_iter()
+            .find(|(key, _)| key == &StorageWellKnownKeys::CODE.to_vec())
+        {
+            Some((_, code)) => ValidationCode(code),
+            None => return Err(DispatchError::Other("Code not found")),
```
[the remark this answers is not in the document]
Agusrodri: Applied in 1da160133ce626b398e73c7672930540e8662fa1 as well :)

pallets/registrar/src/lib.rs
Agusrodri: _fixed_

primitives/traits/src/lib.rs
Agusrodri: _fixed_

solo-chains/runtime/starlight/src/lib.rs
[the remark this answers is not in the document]
Agusrodri: _fixed_

solo-chains/runtime/starlight/src/lib.rs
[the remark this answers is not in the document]
Agusrodri: _fixed_

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: _fixed_

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: _resolved_

pallets/registrar/src/mock.rs
girazoki: I think a way to test it is:
- you can have a wrapper type that can be set to () or to something where at least a variable is updated. The goal is to at least know those funcions are being called succesfully

solo-chains/runtime/starlight/src/tests/collator_assignment_tests.rs
girazoki: fixed

pallets/registrar/src/mock.rs
Agusrodri: At the end I followed another approach that I think it's better. Basically I just leveraged the existing `mock_data` pallet to also include and verify the calls to `InnerRegistrar`.

I updated some tests of `register` and `deregister` where I added extra checks to see if the `Mock` storage of `mock_data` pallet is being updated properly.

The changes are included in 29bf2d8365d566de33bb360f6ec4a98af15f0888. Let me know if this approach works for you, otherwise we can go with the wrapper one :)

girazoki: very nice Agustin (and thanks @tmpolaczyk for inspiring this).

Fixed

girazoki (approved):
Good job @Agusrodri

pallets/registrar/src/lib.rs
tmpolaczyk: `take` is a write, but if I'm correct storage items that only exist inside a block doesn't count as writes because they don't affect the state root and they don't need PoV. Would be nice if substrate had some explicit "transactional" storage, but I think it doesn't. 

So the current implementation looks good to me.

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1606,6 +1611,76 @@ parameter_types! {
...
+        head_data: Option<HeadData>,
+    ) -> DispatchResult {
+        // Return early if head_data is not specified
+        let genesis_head = match head_data {
+            Some(data) => data,
+            None => return Err(ContainerRegistrarError::<Runtime>::HeadDataNecessary.into()),
+        };
+
+        let key_values: Vec<(Vec<u8>, Vec<u8>)> =
+            genesis_storage.into_iter().map(Into::into).collect();
```
tmpolaczyk, line 1638: You don't need to collect here if you just need to find the code, try something like this:

```rust
// Check if the wasm code is present in storage
let validation_code = match genesis_storage
    .into_iter()
    .map(Into::into)
    .find(|(key, _): &(Vec<u8>, Vec<u8>)| key == &StorageWellKnownKeys::CODE)
{
    Some((_, code)) => ValidationCode(code),
    None => return Err(ContainerRegistrarError::<Runtime>::WasmCodeNecessary.into()),
};
```

Also don't call `CODE.to_vec()` on every iteration, you should be able to compare references

solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -1606,6 +1611,76 @@ parameter_types! {
...
+    for InnerStarlightRegistrar<Runtime, AccountId, RegistrarManager, RegistrarWeightInfo>
+where
+    RegistrarManager: RegistrarInterface<AccountId = AccountId>,
+    RegistrarWeightInfo: paras_registrar::WeightInfo,
+    Runtime: pallet_registrar::Config,
+{
+    fn register(
+        who: AccountId,
+        id: ParaId,
+        genesis_storage: Vec<ContainerChainGenesisDataItem>,
```
tmpolaczyk, line 1628: Actually if this method doesn't need the full genesis storage, it would be much better to only take the code as input.

```suggestion
        code: &[u8],
```

pallets/registrar/src/tests.rs
```diff
@@ -1017,10 +1098,20 @@ fn deregister_2_container_chains_in_same_block() {
...
                 HookCall::MarkedValid(43.into()),
+                HookCall::InnerScheduleParaUpgrade(43.into()),
+                HookCall::InnerScheduleParaDowngrade(42.into()),
+                HookCall::InnerScheduleParaDowngrade(43.into()),
                 HookCall::Deregistered(42.into()),
                 HookCall::Deregistered(43.into()),
+                HookCall::InnerDeregisterWeight,
+                HookCall::InnerDeregisterWeight,
+                HookCall::InnerDeregister(42.into()),
+                HookCall::InnerDeregister(43.into()),
```
tmpolaczyk, line 1114: This is nice, it shows the entire parachain cycle and which hooks are executed in which order.

solo-chains/runtime/starlight/src/tests/relay_registrar.rs
```diff
@@ -95,3 +100,505 @@ fn registrar_needs_a_reserved_para_id() {
...
+            // in the relay Registrar pallet.
+
+            assert_eq!(Runtime::registered_paras(), vec![1001.into(), 1002.into()]);
+            assert_eq!(Runtime::genesis_data(1003.into()).as_ref(), None);
+            run_to_session(1u32);
+
+            // Change max_head_data_size config.
+            assert_ok!(
+                Configuration::set_max_head_data_size(root_origin(), 20500),
+                ()
```
tmpolaczyk, line 124: Is this needed? Because I see you also set it in genesis

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: Yeah that would be nice, but the problem is that we need to take the code from that storage anyway, either inside this function or before calling it (inside `ContainerRegistrar::do_register` function). 

In my opinion is better to pass the genesis storage as a parameter and then search for the code inside the function. This way, we will perform this search only when `InnerRegistrar` has actually an implementation. Otherwise we would perform the search without much sense (like in Dancebox or Flashbox for instance).

Another solution is to receive the code as a parameter as we did with `head_data`, but I don't see any extra value on that, given that it's much more breaking and without much sense, as we already have the code inside `ContainerChainGenesisData`. What do you think?

solo-chains/runtime/starlight/src/tests/relay_registrar.rs
Agusrodri: Nop, this is not needed anymore, as it's being already set in genesis. 

Fixed in 73daab1933b5cfc51de53b7c7e8512420e7907c0.

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: Thanks for the suggestion! 

Fixed in 73daab1933b5cfc51de53b7c7e8512420e7907c0.

pallets/registrar/src/lib.rs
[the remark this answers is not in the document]
tmpolaczyk: *fixed*

solo-chains/runtime/starlight/src/lib.rs
Agusrodri: Great, yes at least we can avoid cloning the storage in that section. 

Mock fixed in e83f4a4c817cd829c039cb983531875273aa9f6b.

tmpolaczyk (approved):

PR #663, opened 2024-08-23
Add force-debug feature to starlight

tmpolaczyk:
So we won't see `wasm:stripped` in zombienet logs if compiled with `--features force-debug`


girazoki:
Please provide at least small description on the PR @tmpolaczyk

girazoki (approved):

PR #664, opened 2024-08-26
WIP Support collators in starlight

tmpolaczyk:


client/consensus/src/collators/lookahead.rs
```diff
@@ -637,7 +638,12 @@ where
                                 let slot = inherent_providers.slot();
                                 let container_chain_slot_duration = (params.get_current_slot_duration)(parent_header.hash());
 
-                                match try_to_buy_core::<_, _, <<OBlock as BlockT>::Header as HeaderT>::Number, _, CIDP, _, _>(params.para_id, aux_data, inherent_providers, &params.keystore, params.orchestrator_client.clone(), params.orchestrator_tx_pool.clone(), parent_header, params.orchestrator_slot_duration, container_chain_slot_duration).await {
+                                let buy_core_result = if params.solochain {
```
girazoki, line 641: comment! or even add todo

client/consensus/src/lib.rs
```diff
@@ -245,7 +245,147 @@ where
...
+/// and makes no guarantees which one as that depends on the keystore's iterator behavior.
+/// This is the standard way of determining which key to author with.
+/// It also returns its ParaId assignment
+pub async fn first_eligible_key_next_session_solochain<B: BlockT, C, P>(
+    client: &C,
+    parent_hash: &H256,
+    keystore: KeystorePtr,
+) -> Option<(AuthorityId<P>, ParaId)>
+where
+    C: RelayChainInterface + ?Sized,
```
girazoki, line 316: The only thing that changes here with respect to the previous function is the bounds on C?

tmpolaczyk: The bounds on C and this one is async while the old one is not. So we can't use `find_map` here. If we wanted to unify them we would have to make them both async, I can try.

client/service-container-chain/src/service.rs
```diff
@@ -322,13 +314,35 @@ fn start_consensus_container(
...
-        .expect("start_consensus_container: slot duration should exist");
+    let crate::spawner::CollationParams {
+        collator_key,
+        orchestrator_tx_pool,
+        orchestrator_client,
+        orchestrator_para_id,
+        solochain,
+    } = collation_params;
+    let slot_duration = if solochain {
+        // TODO: always use 6s for aura, not sure why I was getting an error when trying to use
```
girazoki, line 328: remove this TODO, aura is not implemented in the relay chain (babe being used)

solo-chains/runtime/starlight/src/genesis_config_presets.rs
```diff
@@ -317,7 +317,7 @@ fn starlight_testnet_genesis(
             "config": runtime_parachains::configuration::HostConfiguration {
                 scheduler_params: SchedulerParams {
                     max_validators_per_core: Some(1),
-                    num_cores: para_ids.len() as u32,
+                    num_cores: 4,
```
girazoki, line 320: why this?

tmpolaczyk: To support registering new chains in zombienet. Without this, chains 2000 and 2001 work, but when we register 2002 it won't have a core. I set it to 4 because we have 4 validators.

girazoki: alright, if it does not break anything else..

client/consensus/src/lib.rs
girazoki: that would be great

tmpolaczyk: I can remove this function if we add some methods to the OrchestratorChainInterface: https://github.com/moondance-labs/dancekit/pull/33

tmpolaczyk: *fixed*

client/service-container-chain/src/service.rs
tmpolaczyk: *fixed*

test/suites/zombie-tanssi-relay/test-tanssi-relay.ts
```diff
@@ -1,21 +1,442 @@
...
         });
+
+        it({
+            id: "T03",
+            timeout: 600000,
+            title: "Test assignation did not change",
+            test: async function () {
+                // TODO: starlight collator assignment is not set properly on genesis, so we need to wait 2 sessions
+                // for the collators to be assigned, and a bit more for container chains to start producing blocks
+                await waitSessions(context, relayApi, 4, null, "Tanssi-relay");
```
tmpolaczyk, line 75: TODO: Set this back down to 3, or even 2 if it works

girazoki: yes please!

client/consensus/src/lib.rs
```diff
@@ -48,6 +50,7 @@ pub use {
     sp_application_crypto::AppPublic,
     sp_consensus::Error as ConsensusError,
     sp_core::crypto::{ByteArray, Public},
```
girazoki, line 52:
```suggestion
    sp_core::{crypto::{ByteArray, Public}, H256}
```

client/consensus/src/lib.rs
```diff
@@ -48,6 +50,7 @@ pub use {
...
+    sp_core::H256,
```
girazoki, line 53:
```suggestion

```

client/consensus/src/lib.rs
```diff
@@ -145,14 +148,13 @@ use {
 /// and makes no guarantees which one as that depends on the keystore's iterator behavior.
 /// This is the standard way of determining which key to author with.
 /// It also returns its ParaId assignment
-pub fn first_eligible_key<B: BlockT, C, P>(
+pub async fn first_eligible_key<C, P>(
     client: &C,
-    parent_hash: &B::Hash,
+    parent_hash: &H256,
```
girazoki, line 153: why this change? isnt this forcing us to use one specific hash?

girazoki: Removing the first generic I mean

client/service-container-chain/src/spawner.rs
```diff
@@ -638,13 +639,19 @@ impl<SelectSyncMode: TSelectSyncMode> Spawner for ContainerChainSpawner<SelectSy
...
+        validator: bool,
+        solochain: bool,
+    ) {
         // The node always starts as an orchestrator chain collator.
         // This is because the assignment is detected after importing a new block, so if all
         // collators stop at the same time, when they start again nobody will produce the new block.
         // So all nodes start as orchestrator chain collators, until the first block is imported,
         // then the real assignment is used.
-        if validator {
+        // Except in solochain mode, then the initial assignment is None.
```
girazoki, line 653: why is this? because we dont have "orchestrator" parachain?

solo-chains/runtime/starlight/src/genesis_config_presets.rs
girazoki: fixed

girazoki: Why dont we control the `num_cores` from withing zombienet itself?

test/configs/zombieStarlight.json
```diff
@@ -17,8 +17,7 @@
                                 "max_candidate_depth": 3
                             },
                             "scheduler_params": {
-                                "scheduling_lookahead": 2,
-                                "num_cores": 1
+                                "scheduling_lookahead": 2
```
girazoki, line 20: Here, for instance

girazoki:
I general this PR is as good as it can be for now. There is work to be done on the node-side, but that is planned for another PR.

@tmpolaczyk let's fix the small things, and let's see what the rest has to say

client/consensus/src/lib.rs
tmpolaczyk: This fixes a compile error, I guess now we are forced to always use H256.

Caused by `check_para_id_assignment` function requiring H256.

client/service-container-chain/src/spawner.rs
tmpolaczyk: Yes, without this change it panics when calling the `collate_on` closure

test/configs/zombieStarlight.json
tmpolaczyk: What do you mean?

client/consensus/src/lib.rs
girazoki: can we change `check_para_id_assignment` to accept a generic? or is it too much of a change? cause this feels "bad" ina  way

client/service-container-chain/src/spawner.rs
girazoki: disregard

test/configs/zombieStarlight.json
girazoki: instead of hardcoding the number of blocks to 4 into genesis, I would rather have it as it was and then control it in zombienet through here, what do you think?

client/consensus/src/lib.rs
tmpolaczyk: We can if we make the `OrchestratorChainInterface` trait generic, but I think it's a bad idea.

test/configs/zombieStarlight.json
tmpolaczyk: I can add it here as well, but I would also leave the value of 4 in genesis.

client/consensus/src/lib.rs
girazoki: Alright, let's leave it as is for ow..

girazoki: disregard

girazoki (approved):
it's an approval for me, but I want more eyes on it @ParthDesai @nanocryk @Agusrodri

client/service-container-chain/src/service.rs
```diff
@@ -322,13 +314,23 @@ fn start_consensus_container(
...
+    let crate::spawner::CollationParams {
+        collator_key,
+        orchestrator_tx_pool,
+        orchestrator_client,
+        orchestrator_para_id,
+        solochain,
+    } = collation_params;
+    let slot_duration = if solochain {
+        // Solochains use Babe instead of Aura, which has 6s slot duration
+        SlotDuration::from_millis(6_000)
```
ParthDesai, line 329: Better to read this from babe configuration then hard code it.

test/configs/zombieStarlight.json
girazoki: fine, but yes let's add it in both places

client/service-container-chain/src/service.rs
girazoki: can we read it from config? I am not sure if there is a runtime-api from which we can read it

test/configs/zombieStarlight.json
girazoki: fixed

client/service-container-chain/src/service.rs
ParthDesai: Yes. There is a runtime-api. Example: https://github.com/paritytech/polkadot-sdk/blob/f0e420a031b2b0d4c2d3c065ab3e4c43a160b0d7/substrate/bin/node/runtime/src/lib.rs#L2860

tmpolaczyk: Actually we already have a variable here called `relay_chain_slot_duration`, I will use that instead. That's also hardcoded to 6s, I will change it to use the runtime api.

tmpolaczyk: I didn't find a way to get that using the overseer. We could call the runtime-api directly using the new method, but it seems too much effort, and I need to import sp_consensus_babe to decode the result.

So I'm keeping it hardcoded to 6s.

Leaving the code here for future reference

```rust
let relay_chain_slot_duration = {
    let phash = relay_chain_interface.finalized_block_hash().await.unwrap();
    let babe_config = relay_chain_interface.call_remote_runtime_function_encoded("BabeApi_configuration", phash, &[]).await.unwrap();
    let babe_config: sp_consensus_babe::BabeConfiguration = Decode::decode(&mut babe_config.as_slice()).unwrap();
    
    babe_config.slot_duration
}
```

client/consensus/src/collators/lookahead.rs
```diff
@@ -636,7 +637,13 @@ where
...
+                                    // TODO: implement parathread support for solochain
+                                    unimplemented!("Cannot buy core for parathread in solochain")
```
ParthDesai, line 642: We should avoid this panic as it is possible for sudo to register parathread which can basically crash the collator with that assignment. 

I would suggest to break the loop or if that also result in error, print a warning and `continue` on next iteration.

tmpolaczyk: *fixed*

client/consensus/src/collators/lookahead.rs
```diff
@@ -636,7 +637,14 @@ where
                                 let slot = inherent_providers.slot();
                                 let container_chain_slot_duration = (params.get_current_slot_duration)(parent_header.hash());
 
-                                match try_to_buy_core::<_, _, <<OBlock as BlockT>::Header as HeaderT>::Number, _, CIDP, _, _>(params.para_id, aux_data, inherent_providers, &params.keystore, params.orchestrator_client.clone(), params.orchestrator_tx_pool.clone(), parent_header, params.orchestrator_slot_duration, container_chain_slot_duration).await {
+                                let buy_core_result = if params.solochain {
+                                    // TODO: implement parathread support for solochain
+                                    log::warn!("Unimplemented: cannot buy core for parathread in solochain");
```
ParthDesai, line 642: I would prefer if we can remove this panic. Since, as of now, it is possible to register parathreads in the solo chain using sudo, which can potentially trigger this. 

I would suggest instead of panic, if we can make the loop `no-op` by `continue` on next iteration?

ParthDesai: *duplicate*

node/src/service.rs
```diff
@@ -678,6 +689,227 @@ pub async fn start_parachain_node(
...
+        .build_cumulus_network::<_, sc_network::NetworkWorker<_, _>>(
+            &parachain_config,
+            orchestrator_para_id,
+            import_queue,
+            relay_chain_interface.clone(),
+        )
+        .await?;
+
+    /*
+    let rpc_builder = {
```
ParthDesai, line 753: Why was this commented out?

node/src/service.rs
```diff
@@ -678,6 +689,227 @@ pub async fn start_parachain_node(
...
+    if validator {
+        // Start task which detects para id assignment, and starts/stops container chains.
+        build_check_assigned_para_id(
+            orchestrator_chain_interface.clone(),
+            sync_keystore.clone(),
+            cc_spawn_tx.clone(),
+            node_builder.task_manager.spawn_essential_handle(),
+        );
+    }
+    //node_builder.network.start_network.start_network();
```
ParthDesai, line 826: This as well.

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+        para_id: ParaId,
+    ) -> OrchestratorChainResult<Option<ContainerChainGenesisData>> {
+        let res: Option<ContainerChainGenesisData> = call_remote_runtime_function(
+            &self.relay_chain_interface,
+            "RegistrarApi_genesis_data",
+            relay_parent,
+            &para_id,
+        )
+        .await
+        .unwrap();
```
ParthDesai, line 1429: Can you remove this unwrap?

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+        para_id: ParaId,
+    ) -> OrchestratorChainResult<Vec<Vec<u8>>> {
+        let res: Vec<Vec<u8>> = call_remote_runtime_function(
+            &self.relay_chain_interface,
+            "RegistrarApi_boot_nodes",
+            relay_parent,
+            &para_id,
+        )
+        .await
+        .unwrap();
```
ParthDesai, line 1446: Can you remove this unwrap?

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+        para_id: ParaId,
+    ) -> OrchestratorChainResult<Option<BlockNumber>> {
+        let res: Option<BlockNumber> = call_remote_runtime_function(
+            &self.relay_chain_interface,
+            "AuthorNotingApi_latest_block_number",
+            relay_parent,
+            &para_id,
+        )
+        .await
+        .unwrap();
```
ParthDesai, line 1463: Can you remove this unwrap?

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+            .await
+            .map_err(|e| OrchestratorChainError::Application(Box::new(e)))
+    }
+
+    async fn data_preserver_active_assignment(
+        &self,
+        _orchestrator_parent: PHash,
+        _profile_id: DataPreserverProfileId,
+    ) -> OrchestratorChainResult<DataPreserverAssignment<ParaId>> {
+        unimplemented!("Data preserver node does not support Starlight yet")
```
ParthDesai, line 1487: If possible, remove this `unimplemented` as well.

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+        authority: NimbusId,
+    ) -> OrchestratorChainResult<Option<ParaId>> {
+        let res: Option<ParaId> = call_remote_runtime_function(
+            &self.relay_chain_interface,
+            "TanssiAuthorityAssignmentApi_check_para_id_assignment",
+            relay_parent,
+            &authority,
+        )
+        .await
+        .unwrap();
```
ParthDesai, line 1502: Is it possible to remove this unwrap?

node/src/service.rs
```diff
@@ -1063,4 +1315,209 @@ where
...
+        authority: NimbusId,
+    ) -> OrchestratorChainResult<Option<ParaId>> {
+        let res: Option<ParaId> = call_remote_runtime_function(
+            &self.relay_chain_interface,
+            "TanssiAuthorityAssignmentApi_check_para_id_assignment_next_session",
+            relay_parent,
+            &authority,
+        )
+        .await
+        .unwrap();
```
ParthDesai, line 1519: Is it possible to remove this unwrap?

node/src/service.rs
tmpolaczyk: Because I wasn't sure if it's needed or not, will remove it now

node/src/service.rs
tmpolaczyk: This will only happen if someone runs a node with `--solochain --data-preserver` flags at the same time, which is not implemented yet, so I think this is the best to handle that case

node/src/service.rs
tmpolaczyk: Probably yes for all of them, good catch

client/service-container-chain/src/service.rs
```diff
@@ -308,8 +300,8 @@ pub fn container_chain_import_queue(
 fn start_consensus_container(
     client: Arc<ContainerChainClient>,
     backend: Arc<FullBackend>,
-    orchestrator_client: Arc<ParachainClient>,
-    orchestrator_tx_pool: Arc<FullPool<Block, ParachainClient>>,
+    // TODO: this should be generic, so we can use RelayChainInterface for solochain
+    collation_params: crate::spawner::CollationParams,
```
Agusrodri, line 304: What does this mean exactly? I see that we are already using the same `relay_chain_interface` object for both solochain and parachain modes, right? I didn't fully get the TODO meaning.

tmpolaczyk: That's comment is outdated, it was referring to the `orchestrator_client` field which I ended up removing, so I will remove the TODO as well.

ParthDesai (approved):

node/src/service.rs
tmpolaczyk: *fixed*

node/src/service.rs
tmpolaczyk: *fixed*

node/src/service.rs
tmpolaczyk: *fixed*

node/src/service.rs
tmpolaczyk: *fixed*

node/src/service.rs
tmpolaczyk: *fixed*

node/src/service.rs
tmpolaczyk: *roadmap*

node/src/service.rs
tmpolaczyk: *fixed*

client/service-container-chain/src/service.rs
tmpolaczyk: *fixed*

client/service-container-chain/src/service.rs
tmpolaczyk: *disregard*

client/consensus/src/lib.rs
tmpolaczyk: *fixed*

client/consensus/src/lib.rs
tmpolaczyk: *fixed*

PR #666, opened 2024-08-27
[POC] Spawn embeded Container chain node on data preservers assignment

nanocryk:
Implements the logic to detect assignments and spawn emebed container chain service corresponding to assignment.

This is a proof of concept and will need further work in later PRs:
- this is only implemented in `simple` container chain node for now. Code in `command.rs` should be refactored out in a separate crate to be used by `frontier` container, which should support customizing the RPC (to support frontier RPC).
- only the spawning of the first assignment is tested. Additional work may be required to properly react to unassignment and re-assignment.

Blocked by https://github.com/moondance-labs/dancekit/pull/32 which adds an `OrchestratorChainInterface` function.


client/service-container-chain/src/spawner.rs
```diff
@@ -256,16 +256,19 @@ async fn try_spawn<SelectSyncMode: TSelectSyncMode>(
     if !start_collation {
         collation_params = None;
 
-        log::info!("This is a syncing container chain, using random ports");
+        log::info!("This is a syncing container chain, using random ports unless they are explicitly provided as CLI args");
         // Use random ports to avoid conflicts with the other running container chain
+        // Don't override provided port.
+        // TODO: How does this prevent conflicts? All containers will have the same ports.
```
tmpolaczyk, line 262: It's because in this branch we are spawning a container chain that will only be alive for 1 session, and will restart later with the correct ports. So we only need the ports to not conflict with the CLI ports, because that can cause conflict with the other chain that is running. At first I used different ports every time (`rng.gen()`) but then I realized that a fixed port will never conflict while a random port may conflict sometimes.

If you need to run a container chain for more than 1 session, you shouldn't enter this branch, the default is to use CLI ports.

client/service-container-chain/src/spawner.rs
```diff
@@ -256,16 +256,19 @@ async fn try_spawn<SelectSyncMode: TSelectSyncMode>(
     if !start_collation {
```
tmpolaczyk, line 256: Add something like this as per previous comment

```suggestion
    if !start_collation && not_data_preserver {
```

test/configs/zombieDataPreservers.json
```diff
@@ -0,0 +1,114 @@
...
+                    "command": "../target/release/container-chain-simple-node",
+                    "args": ["--no-hardware-benchmarks", "--database=paritydb", "--wasmtime-precompiled=wasm"],
+                    "ws_port": 9949,
+                    "p2p_port": 33049,
+                    "prometheus_port": 33102
+                },
+                {
+                    "name": "DataPreserver",
+                    "validator": false,
+                    "command": "scripts/container-chain-args.sh --rpc-port 9950 -- ../target/release/container-chain-simple-node",
```
tmpolaczyk, line 95: Not sure if you need this script in the end. If you only want to add some container args, you can do it like `keep-db` in this test:

https://github.com/moondance-labs/tanssi/blob/57a05db22eb408aa3d6bab146f38c1fbb128df23/test/configs/zombieTanssiKeepDb.json#L77C26-L77C37

So I guess `"-- --rpc-port 9950"`

The reason why container args come before relay chain args is because that's the order you get when you add `-- --arg` here in zombienet. Originally I tried to add container args after relay args but that didn't work.

client/service-container-chain/src/spawner.rs
tmpolaczyk: *fixed*

client/service-container-chain/src/spawner.rs
tmpolaczyk: *fixed*

client/orchestrator-chain-rpc-interface/src/lib.rs
```diff
@@ -318,4 +319,17 @@ impl OrchestratorChainInterface for OrchestratorChainRpcClient {
     async fn finalized_block_hash(&self) -> OrchestratorChainResult<PHash> {
         self.request("chain_getFinalizedHead", rpc_params![]).await
     }
+
+    async fn data_preserver_active_assignment(
```
girazoki, line 323: @tmpolaczyk how well will this plug with your changes?

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,90 @@
...
+    crate::spawner::{ContainerChainSpawner, TSelectSyncMode},
+    dc_orchestrator_chain_interface::{
+        DataPreserverAssignment, OrchestratorChainInterface, OrchestratorChainResult,
+    },
+    futures::stream::StreamExt,
+    std::{future::Future},
+    tc_consensus::ParaId,
+};
+
+pub type ProfileId = <dancebox_runtime::Runtime as pallet_data_preservers::Config>::ProfileId;
```
girazoki, line 27: I would try to make this generic, i.e. for any runtime that implements `pallet_data_preservers::Config`, I am thinking of starlight here

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,90 @@
...
+            match (current_assignment, new_assignment) {
+                // no change
+                (x, y) if x == y => continue,
+                (
+                    Assignment::NotAssigned | Assignment::Inactive(_),
+                    Assignment::Active(para_id),
+                ) => {
+                    spawner.spawn(para_id, false).await;
+                }
+                (Assignment::Active(para_id), Assignment::Inactive(x)) if para_id == x => {
```
girazoki, line 68: let's add comments in each of these

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,90 @@
...
+
+pub type ProfileId = <dancebox_runtime::Runtime as pallet_data_preservers::Config>::ProfileId;
+
+async fn try_fut<T, E>(fut: impl Future<Output = Result<T, E>>) -> Result<T, E> {
+    fut.await
+}
+
+/// Watch assignements by indefinitly listening to finalized block notifications and switching to
+/// the chain the profile is assigned to.
+pub async fn task_watch_assignment<S: TSelectSyncMode>(
```
girazoki, line 35: it would be good if we can unitest this function. Do you think it is very difficult?

container-chains/nodes/simple/src/command.rs
```diff
@@ -592,3 +499,152 @@ impl CliConfiguration<Self> for RelayChainCli {
         self.base.base.node_name()
     }
 }
+
+fn rpc_provider_mode(cli: Cli, profile_id: u64) -> Result<()> {
```
girazoki, line 503: maybe this belongs more in service.rs?

test/suites/data-preservers/test_data_preservers.ts
```diff
@@ -0,0 +1,142 @@
+import { beforeAll, describeSuite, expect } from "@moonwall/cli";
+import { ApiPromise, Keyring } from "@polkadot/api";
+import { signAndSendAndInclude } from "../../util/block";
+import { getHeaderFromRelay } from "../../util/relayInterface";
+import fs from "fs/promises";
+
+describeSuite({
```
girazoki, line 7: Let's put this test in the CI

client/service-container-chain/src/data_preservers.rs
tmpolaczyk: I was going to recommend to use `handle_update_assignment` because that's already tested, but not sure if it can be easily integrated as that function is quite complex.

client/service-container-chain/src/data_preservers.rs
nanocryk: Since it interacts with the orchestrator interface it should probably use this type:
https://github.com/moondance-labs/dancekit/pull/32/files#diff-cc0640977493f1e56ff4eec9045ead68475bd6d2d41284ffa919745c7e726bb5R101

The linked type cannot be generic since we can't know in advance to what runtime we're connecting to I think.

tmpolaczyk: This is not needed, somehow I managed to support starlight without adding starlight as a dependency.

client/orchestrator-chain-rpc-interface/src/lib.rs
nanocryk: Note that now "data preserver mode" is no longer implemented using a subcommand but by a normal CLI argument. You'll thus probably have to add this argument to your subcommand.

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,90 @@
...
+    profile_id: ProfileId,
+) {
+    use dc_orchestrator_chain_interface::DataPreserverAssignment as Assignment;
+
+    if let OrchestratorChainResult::Err(e) = try_fut(async move {
+        let orchestrator_chain_interface = spawner.params.orchestrator_chain_interface.clone();
+
+        let mut current_assignment = DataPreserverAssignment::<ParaId>::NotAssigned;
+
+        let mut stream = orchestrator_chain_interface
```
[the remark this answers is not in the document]
nanocryk: Yes

client/orchestrator-chain-rpc-interface/src/lib.rs
tmpolaczyk: Yeah, basically we need a `rpc_provider_mode_solochain` function

container-chains/nodes/simple/src/command.rs
```diff
@@ -592,3 +499,152 @@ impl CliConfiguration<Self> for RelayChainCli {
...
+                    tokio_handle: config.tokio_handle.clone(),
+                    chain_type: config.chain_spec.chain_type(),
+                    relay_chain,
+                    relay_chain_interface,
+                    sync_keystore: keystore_container.keystore(),
+                    orchestrator_para_id: para_id,
+                    collation_params: None,
+                    spawn_handle: task_manager.spawn_handle().clone(),
+                    // TODO: Which sync mode to use?
+                    sync_mode: { move |_db_exists, _para_id| Ok(sc_cli::SyncMode::Warp) },
```
tmpolaczyk, line 621: You can reuse the same logic as the other function, right?

```rust
move |db_exists, para_id| {
    spawner::select_sync_mode_using_client(
        db_exists,
        &orchestrator_client.clone(),
        para_id,
    )
}
```

node/src/cli.rs
```diff
@@ -184,7 +184,7 @@ pub struct Cli {
     #[arg(long)]
     pub para_id: Option<u32>,
 
-    /// Relay chain arguments, optionally followed by "--" and orchestrator chain arguments
+    /// Relay chain arguments, optionally followed by "--" and container chain arguments
```
tmpolaczyk, line 187: This comment was wrong, maybe something like this:

```suggestion
    /// Extra arguments, `container-args -- relay-args` or `relay-args` if no `--`
```

container-chains/nodes/simple/src/command.rs
nanocryk: No because it depends on the client which we don't have.

tmpolaczyk: Since this is not a collator it can safely use warp sync.

Because the bug is collators getting stuck trying to warp sync a chain with 0 blocks and not producing blocks. But if the data preserver node gets stuck if will fix itself when another node creates a new block.

tmpolaczyk: *disregard*

container-chains/nodes/simple/src/command.rs
nanocryk: No other functions taking `Cli` are in service. In the future we'll likely refactor generic part into a common crate and keep the code in commands.rs minimal.

client/service-container-chain/src/data_preservers.rs
girazoki: resolved

client/service-container-chain/src/data_preservers.rs
girazoki: @nanocryk how difficult would it be? what is your opinion?

nanocryk: Not sure it is a good idea to merge it with `handle_update_assignment`, especially since `task_watch_assignment` is relatively short. I can take a look at how you test the other function to write similar tests :)

client/service-container-chain/src/data_preservers.rs
girazoki: alright, then let's disregard

container-chains/nodes/simple/src/command.rs
girazoki: roadmap

container-chains/nodes/simple/src/command.rs
```diff
@@ -592,3 +499,152 @@ impl CliConfiguration<Self> for RelayChainCli {
...
+                    container_chain_cli,
+                    tokio_handle: config.tokio_handle.clone(),
+                    chain_type: config.chain_spec.chain_type(),
+                    relay_chain,
+                    relay_chain_interface,
+                    sync_keystore: keystore_container.keystore(),
+                    orchestrator_para_id: para_id,
+                    collation_params: None,
+                    spawn_handle: task_manager.spawn_handle().clone(),
+                    // TODO: Which sync mode to use?
```
tmpolaczyk, line 620:
```suggestion
                    // We can use warp sync because the warp sync bug only affects collators
```

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,360 @@
...
+                (
+                    Assignment::Active(para_id),
+                    Assignment::Inactive(_) | Assignment::NotAssigned,
+                ) => {
+                    spawner.stop(para_id, false); // don't keep db
+                }
+                // Changed para id, remove previous node and start new one
+                (Assignment::Active(previous_para_id), Assignment::Active(para_id)) => {
+                    spawner.stop(previous_para_id, false); // don't keep db
+                    spawner.spawn(para_id, false).await;
```
tmpolaczyk, line 81: This case could cause issues with the ports being unavailable, because `stop` doesn't wait for the chain to stop. You can use `wait_for_paritydb_lock` to wait for the previous chain to stop.

client/service-container-chain/src/data_preservers.rs
girazoki: fixed

girazoki (approved):

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,420 @@
...
+        spawner.ensure_all_events_were_emitted();
+
+        spawner.set_expectations(vec![]);
+        spawner.chain_interface.mock_block({
+            let mut map = BTreeMap::new();
+            map.insert(profile_id, DataPreserverAssignment::NotAssigned);
+            map
+        });
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        spawner.ensure_all_events_were_emitted();
```
tmpolaczyk, line 418: Active(1) -> Active(2) case is not tested.

client/service-container-chain/src/data_preservers.rs
```diff
@@ -0,0 +1,420 @@
...
+        spawner.set_expectations(vec![SpawnerEvent::Started(para_id2, false)]);
+        spawner.chain_interface.mock_block({
+            let mut map = BTreeMap::new();
+            map.insert(profile_id, DataPreserverAssignment::Active(para_id2));
+            map
+        });
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        spawner.ensure_all_events_were_emitted();
+
+        spawner.set_expectations(vec![SpawnerEvent::Stopped(para_id2, true)]);
```
tmpolaczyk, line 393: It feels weird to have the expected events before the tests, in most other tests we put them in the final assert. But you can ignore this.

nanocryk: Instead each function could collect the events then expect will ensure it matches :)

tmpolaczyk (approved):
Thank you! Don't forget to merge back master again

client/service-container-chain/src/data_preservers.rs
girazoki: fixed

client/service-container-chain/src/data_preservers.rs
girazoki: disregard

PR #671, opened 2024-08-29
new tools for developers: testing a custom runtime upgrade against released binaries

girazoki:
Adds two new jobs to Moonwall, which dont execute in the CI, with the following purpose:
- Download the latest released runtime, and create a chain-spec based of it
- Download the latest release node, and start a network composed of relay + tanssi + a container with this node & the aforementioned generated specs
- Perform a runtime upgrade against a locally built runtime

This job is similar to the one we have running in the CI, however this runs with the latest released node binaries, while the CI one runs with the compiled binary.

Developers can test it doing from the `test` folder doing:
- `pnpm moonwall test zombie_simple_template_upgrade_with_latest_releases` if they are upgrading a non-frontier runtime or:
- `pnpm moonwall test zombie_frontier_template_upgrade_with_latest_releases` if they are upgrading a frontier runtime

This should prevent developers from pushing runtimes that are not compatible with the latest release node (as it happened in the past)
CC @augustocollado


Agusrodri:
I checked locally and both jobs work fine! However, the Flashbox node was panicking with:

```
panicked at /home/gh-actions/.cargo/git/checkouts/polkadot-sdk-df3be1d6828443a1/be526b0/cumulus/pallets/parachain-system/src/lib.rs:140:13:
Relay chain block number needs to strictly increase between Parachain blocks!
```
 This is because Flashbox config for `parachain_system` pallet was using `RelayNumberStrictlyIncreases` for `CheckAssociatedRelayNumber` type, and should use `RelayNumberMonotonicallyIncreases`. 

Fixed in 89ac55ae1e7a9d50fdf2b6cbdea4f17170d7b67c.

Agusrodri (approved):

PR #675, opened 2024-09-09
Starlight benchmarking

chexware:
This PR adds Starlight runtime to the benchmarking workflow scripts.


girazoki:
@chexware please provide a PR description and also update with master

girazoki:
Doing BINARY=./target/release/tanssi-relay CHAIN=dev ./tools/benchmarking.sh fails on pallet_referenda. And it seems it is because there is two instances of that pallet and they write to the same output file. I suggest:

- we remove FellowshipReferenda from the benchmark list:
- we update the script to use another output file for it

what do you think  @chexware ? I guess the easiest is just to remove it

tools/dev-benchmark-pallet.sh
```diff
@@ -79,6 +79,13 @@ TEMPLATE_PATH=benchmarking/frame-weight-runtime-template.hbs \
     tools/benchmarking.sh "$PALLET" "*" --check
 cp -v tmp/flashbox_weights/$PALLET.rs runtime/flashbox/src/weights/$PALLET.rs
 
+# Starlight weights
+BINARY=target/release/tanssi-relay \
+    TEMPLATE_PATH=benchmarking/frame-weight-runtime-template.hbs \
+    CHAIN=starlight-dev \
+    OUTPUT_PATH=tmp/starlight_weights \
+    tools/benchmarking.sh "$PALLET" "*" --check
```
tmpolaczyk, line 87: This script is for quickly updating weights after adding a new extrinsic to a pallet. It currently won't work for Starlight because we are not using custom weights yet. Therefore, you can ignore it in this PR.

chexware: Sure thing removed it in https://github.com/moondance-labs/tanssi/pull/675/commits/c8c1a7e05c8448e578ff180720f11c04a1f813f5

girazoki (approved):
LGTM

solo-chains/client/cli/src/command.rs
```diff
@@ -89,6 +90,11 @@ impl SubstrateCli for Cli {
             ]),
         )
     }
+
+    #[cfg(feature = "runtime-benchmarks")]
```
tmpolaczyk, line 94: Why do we need this in starlight but not in the other runtimes? Add a comment

tmpolaczyk (approved):
Script looks good. I haven't tried to run the benchmarks, and I don't understand the `HrmpChannelSizeAndCapacityWithSystemRatio` change, someone else pls review that

solo-chains/client/cli/src/command.rs
chexware: Starlight benchmarking for paras_registrar fails if we use the default config since it tries to register the parachains twice if use the standard `load_spec`

chexware:
> Script looks good. I haven't tried to run the benchmarks, and I don't understand the `HrmpChannelSizeAndCapacityWithSystemRatio` change, someone else pls review that

Needed to update that because if we open channel with system parachain  it was set to a default value of 500 which is more than the  one we set in Starlight's [genesis_config_presets](https://github.com/moondance-labs/tanssi/blob/aleks-starlight-benchmarking/solo-chains/runtime/starlight/src/genesis_config_presets.rs) . This resulted in failing benchmarking case for `runtime_parachains::hrmp::establish_channel_with_system`. I decided to use the same config format as Rococo since  we need to care about only the value stored at `genesis_config_presets` and set the channel capacity to the max value in that configuration. In that way if we need bigger channel capacity we could just increase  the values in `genesis_config_presets` without breaking any benchmarking cases

PR #676, opened 2024-09-10
Remove should_end_session check in Starlight, it always returns false

tmpolaczyk:
Remove should_end_session check in Starlight because it always returns false

### DIFF
--- a/solo-chains/runtime/starlight/src/lib.rs
+++ b/solo-chains/runtime/starlight/src/lib.rs
@@ -2313,6 +2313,10 @@ sp_api::impl_runtime_apis! {
         /// Return the registered para ids
         fn registered_paras() -> Vec<ParaId> {
             // We should return the container-chains for the session in which we are kicking in
+            // We could potentially predict whether the next block will yield a session change as in dancebox but this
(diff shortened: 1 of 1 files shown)


solo-chains/runtime/starlight/src/lib.rs
```diff
@@ -2314,16 +2313,7 @@ sp_api::impl_runtime_apis! {
...
-            let should_end_session = <Runtime as pallet_session::Config>::ShouldEndSession::should_end_session(parent_number + 1);
-
-            let session_index = if should_end_session {
-                Session::current_index() +1
-            }
-            else {
-                Session::current_index()
-            };
-
+            let session_index = Session::current_index();
```
girazoki, line 2316: Let's add a comment please since this a different behavior from dancebox:

- We could potentially predict whether the next block will yield a session change as in dancebox but this is innecesary: the starlight blocks are being produced by validators, and therefore it should never stall because of any collator-rotation. Therefore it suffices for collators to predict the chain in which they have to collate after the session-change block

tmpolaczyk: *fixed*

girazoki (approved):

PR #680, opened 2024-09-13
Ethereum RPC Data Providers

nanocryk:
Makes the spawning logic generic over the RuntimeApi, which allows to inject a `GenerateRpcBuilder` that can call runtime apis only present in specific runtimes (different in simple and frontier templates).


client/service-container-chain/src/monitor.rs
```diff
@@ -59,10 +60,10 @@ pub struct SpawnedContainer {
     /// Used to check the reference count, if it's 0 it means the database has been closed
     pub backend: std::sync::Weak<ContainerChainBackend>,
     /// Used to check the reference count, if it's 0 it means that the client has been closed.
-    pub client: std::sync::Weak<ContainerChainClient>,
```
tmpolaczyk, line 62: This is unfortunate, this module only cares about the refcount so it shouldn't need this generic param everywhere... But I believe we would need unsafe code to remove it.

rust-toolchain
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.78.0"
+channel = "1.79.0"
```
tmpolaczyk, line 2: This alone should be a separate PR, let's use the same version as polkadot-sdk has in the stable2407 branch

client/service-container-chain/src/monitor.rs
tmpolaczyk: Actually, would it work if you set the type to `dyn Any`? Try initializing it like this

```
client.clone() as Arc<dyn Any + Send + Sync>;
```

rust-toolchain
nanocryk: Which version is used in polkadot-sdk? I cannot find a `rust-toolchain` file :/

I bumped to this version to take advantage of [this new feature](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position), which make trait aliases way more powerful as they can now imply associated types bounds (no need to repeat long where clauses everywhere).

tmpolaczyk: You can find a way to find the rust version here: https://github.com/paritytech/polkadot-sdk/issues/5335

nanocryk: They seem to target 1.77 which doesn't have the feature above. Not using this feature will propagate big where clauses everywhere there is a `RuntimeApi` :/ It should be fine to use a different stable version of Rust than the polkadot repo as long as we all use the same (thanks to `rust-toolchain` file)

tmpolaczyk: That's quite old...

I see they are updating directly to 1.81: https://github.com/paritytech/polkadot-sdk/pull/5676

So you can choose 1.78 or 1.81, but let's do it in a separate PR, and let's merge it after the starlight branch is created (I don't want this to break the release process).

nanocryk: https://github.com/moondance-labs/tanssi/pull/691

client/service-container-chain/src/rpc.rs
```diff
@@ -114,3 +119,103 @@ where
...
+#[derive(CloneNoBound, DefaultNoBound)]
+pub struct GenerateSubstrateRpcBuilder<RuntimeApi>(pub PhantomData<RuntimeApi>);
+impl<RuntimeApi> GenerateSubstrateRpcBuilder<RuntimeApi> {
+    /// Creates a new instance.
+    pub fn new() -> Self {
+        Self(PhantomData)
+    }
+}
+
+const _: () = {
```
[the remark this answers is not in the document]
tmpolaczyk: My guess: to avoid the `use generate_rpc_builder::*;` below from infecting the global scope. I've seen this pattern in polkadot-sdk.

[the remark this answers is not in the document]
nanocryk: Exactly. It is commonly found in macros to import types without polluting the surrounding scope. I tried to experiment here with a module that re-exports all the types needed to impl the trait without needing a lot of imports.

We can keep it or throw it away depending of what you think about this design.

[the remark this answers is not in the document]
tmpolaczyk: I prefer `mod trait_impl` or similar. Using `const` to create a code block is a hack that should only be used in macros, and as far as I'm aware it has no benefits.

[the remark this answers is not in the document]
nanocryk: So like this?
```
mod trait_impl {
    use {super::*, trait_mod::*};
    
    // impl Trait for ...
}
```

tmpolaczyk (approved):

girazoki (approved):
good job @nanocryk

PR #682, opened 2024-09-16
Change pallet_identity deposit in Starlight to match other runtimes

tmpolaczyk:


tmpolaczyk:
Closing because it's not clear what's the correct deposit

PR #683, opened 2024-09-17
Change solochain flag into subcommand

tmpolaczyk:
Instead of using `tanssi-node --solochain` to start a starlight collator, we will use `tanssi-node solochain`.

This allows us to simplify the argument parsing logic, and remove all the orchestrator args from `start_solochain_node`.

A major change is the new keystore path. Instead of putting keystore in

```
Collator1000-01/data/chains/simple_container_2000/keystore
```

We put it in

```
Collator1000-01/data/config/keystore
```

But zombienet will put the keys in the old path, so we need to manually copy it if we are running under zombienet. See `copy_zombienet_keystore` function.

To generate collator keys in the new location, collators will need to add a `--keystore-path` argument to the `tanssi-node key` command.


client/service-container-chain/src/cli.rs
```diff
@@ -45,16 +48,82 @@ pub struct ContainerChainRunCmd {
...
+    /// reduces bandwidth use.
+    #[arg(
+		long,
+		value_parser = validate_relay_chain_url,
+		num_args = 0..,
+		alias = "relay-chain-rpc-url"
+    )]
+    pub relay_chain_rpc_urls: Vec<Url>,
+
+    /// EXPERIMENTAL: Embed a light client for the relay chain. Only supported for full-nodes.
```
girazoki, line 68: is this new?

client/service-container-chain/src/spawner.rs
```diff
@@ -115,9 +115,10 @@ pub struct ContainerChainSpawnParams {
 #[derive(Clone)]
 pub struct CollationParams {
     pub collator_key: CollatorPair,
-    pub orchestrator_tx_pool: Arc<FullPool<OpaqueBlock, ParachainClient>>,
-    pub orchestrator_client: Arc<ParachainClient>,
+    pub orchestrator_tx_pool: Option<Arc<FullPool<OpaqueBlock, ParachainClient>>>,
+    pub orchestrator_client: Option<Arc<ParachainClient>>,
     pub orchestrator_para_id: ParaId,
+    /// If this is `false`, then `orchestrator_tx_pool` and `orchestrator_client` must be `Some`.
```
girazoki, line 121: Could we make this receive `BuyCoreParams` instead? like you did a bit before

node/src/command/solochain.rs
```diff
@@ -0,0 +1,415 @@
...
+            "ParityDb",
+            // Print base path instead of db path because each container will have its own db in a
+            // different subdirectory.
+            self.config.base_path.path().display(),
+        );
+    }
+
+    /// A helper function that runs a node with tokio and stops if the process receives the signal
+    /// `SIGTERM` or `SIGINT`.
+    pub fn run_node_until_exit<F, E>(
```
girazoki, line 100: is this copied from polkadot?

node/src/command/solochain.rs
```diff
@@ -0,0 +1,415 @@
...
+    let signals = tokio_runtime.block_on(async { Signals::capture() })?;
+
+    init_cmd(command, &Cli::support_url(), &Cli::impl_version())?;
+
+    let base_path = command.base_path()?.unwrap();
+    let network_node_name = command.node_name()?;
+    let is_dev = command.is_dev()?;
+    let role = command.role(is_dev)?;
+    // TODO: where to get relay_chain_id from?
+    let relay_chain_id = "starlight_local_testnet".to_string();
```
girazoki, line 176: from the specs I think you can take the relay chain id

girazoki:
Please  remove the comment about the TODO, as this is the behavior we observe in reegular tanssi

client/service-container-chain/src/cli.rs
tmpolaczyk: Copied from polkadot-sdk, I didn't test it but it should "just work"

client/service-container-chain/src/spawner.rs
tmpolaczyk: I tried and it resulted in too many changes for this PR. So if you don't mind we can do it when adding the buy_core functionality for starlight.

node/src/command/solochain.rs
tmpolaczyk: Yes, most of the functions in this file are

client/service-container-chain/src/spawner.rs
girazoki: roadmap

client/service-container-chain/src/cli.rs
```diff
@@ -45,16 +48,82 @@ pub struct ContainerChainRunCmd {
...
+        // So if the user passes `--base-path /tmp/node`, we want the ephemeral container data in
+        // `/tmp/node/containers`, and the persistent storage in `/tmp/node/config`.
+        // TODO: there should be a way to avoid this if we refactor the code that creates the db,
+        // but maybe that breaks dancebox
+        let base_path = match self.base.base_path() {
+            Ok(Some(x)) => x,
+            _ => {
+                // This is maybe unreachable. There is always a default base path, and if run in
+                // `--dev` or `--tmp` mode, a temporary base path is created.
+                panic!("No base path")
```
girazoki, line 93: This panic is not unreachable, it can be reached by `./target/release/tanssi-node solo-chain`

tmpolaczyk: *fixed*

node/src/command.rs
```diff
@@ -338,6 +340,74 @@ pub fn run() -> Result<()> {
...
+            let container_chain_cli = cmd.run.normalize();
+            let runner = solochain::create_runner(&container_chain_cli)?;
+
+            // The expected usage is
+            // `tanssi-node solochain --flag`
+            // So `cmd` stores the flags from after `solochain`, and `cli` has the flags from between
+            // `tanssi-node` and `solo-chain`. We are ignoring the flags from `cli` intentionally.
+            // Would be nice to error if the user passes any flag there, but it's not easy to detect.
+
+            // Zombienet appends a --chain flag after "solo-chain" subcommand, which is ignored, so it's fine,
```
ParthDesai, line 355: Why are we ignoring `--chain` command? Can we instead use that to decide which relay chain id and based on that chain spec?

client/service-container-chain/src/service.rs
```diff
@@ -481,7 +489,7 @@ fn start_consensus_container(
                     })?;
 
                     let authorities = tc_consensus::authorities::<Block, ParachainClient, NimbusPair>(
-                        orchestrator_client_for_cidp.as_ref(),
+                        orchestrator_client_for_cidp.as_ref().unwrap(),
```
ParthDesai, line 492: Is it possible to remove the `unwrap` here?

client/service-container-chain/src/service.rs
```diff
@@ -499,7 +507,7 @@ fn start_consensus_container(
                     );
 
                     let slot_freq = tc_consensus::min_slot_freq::<Block, ParachainClient, NimbusPair>(
-                        orchestrator_client_for_cidp.as_ref(),
+                        orchestrator_client_for_cidp.as_ref().unwrap(),
```
ParthDesai, line 510: Is it possible to remove the `unwrap` here?

node/src/command.rs
tmpolaczyk: This is the `--chain` from container chain args, which should be empty because it will change depending on which container chain the collator is assigned to.

I believe the relay chain id for which I have a TODO is only needed when using an external relay (rpc or light node).

client/service-container-chain/src/service.rs
tmpolaczyk: I tried but it was too many changes. This is related to the comment about `BuyCoreParams`, will be fixed in the future.

node/src/command/solochain.rs
```diff
@@ -0,0 +1,419 @@
...
+    // Collator1000-01/data/chains/simple_container_2000/keystore/
+
+    zombienet_path
+}
+
+/// When running under zombienet, collator keys are injected in a different folder from what we
+/// expect. This function will check if the zombienet folder exists, and if so, copy all the keys
+/// from there into the expected folder.
+pub fn copy_zombienet_keystore(keystore: &KeystoreConfig) -> std::io::Result<()> {
+    let keystore_path = keystore.path().unwrap();
```
ParthDesai, line 379: What is the reasoning behind changing the key path?

node/src/command/solochain.rs
```diff
@@ -0,0 +1,419 @@
...
+            if to < RECOMMENDED_OPEN_FILE_DESCRIPTOR_LIMIT {
+                warn!(
+                    "Low open file descriptor limit configured for the process. \
+                        Current value: {:?}, recommended value: {:?}.",
+                    to, RECOMMENDED_OPEN_FILE_DESCRIPTOR_LIMIT,
+                );
+            }
+        }
+        Ok(fdlimit::Outcome::Unsupported) => {
+            // Unsupported platform (non-Linux)
```
ParthDesai, line 231: Should we be printing a warning here?

client/service-container-chain/src/service.rs
ParthDesai: What about `expect` message?

node/src/command/solochain.rs
tmpolaczyk: In zombienet, the keystore path depends on the chain id, which I already decided to ignore. So I'm hardcoding `simple_container_2000` as the chain id in this case. I tried changing the keystore path in zombienet config but that is not possible.

Outside of zombienet, the default keystore path is `$base_path/chains/dancebox_local/keystore/`, which has the `dancebox_local` chain id but that's confusing because this is not dancebox. So I have set the keystore path to `$base_path/config/keystore/`.

Essentially we need the keystore to be in a separate folder outside of `containers`, because that gets deleted when the collator rotates. The exact path does not matter, and can be changed by the user by using `--keystore-path`.

node/src/command.rs
ParthDesai: So, zombienet appends `--chain` in container chain args correct? 

Okay. In that case why don't we have solochain's `--chain_id` or something (depending upon how relay chain is going to be used.)?

node/src/command/solochain.rs
tmpolaczyk: This is copied from polkadot-sdk, so I'm not touching it

node/src/command/solochain.rs
ParthDesai: Make sense.

*disregard*

node/src/command.rs
tmpolaczyk: But in zombienet we don't need the relay chain id, and outside of zombienet we won't have `--chain`.

node/src/command/solochain.rs
ParthDesai: *disregard*

node/src/command.rs
ParthDesai: I am confused. Why not? Can you explain the reason behind both of them?

tmpolaczyk: Sure. So in zombienet the relay chain id is obtained from the relay chain args appended by zombienet, so we have these args: `--chain simple-template.json -- --chain rococo-local.json`. And the relay chain id is obtained from the `rococo-local` arg. This will also apply to non-zombienet collators that pass the chain spec of starlight in the future.

Outside of zombienet we don't have the first `--chain`, so the args will be like `-- --chain starlight-spec.json`. And again, we already have the relay chain id from the relay chain spec.

The only case this may break is when using an external rpc as relay chain, so we don't have the starlight chain spec as an argument, but I'm not even sure if that's broken and I don't think it's worth to test it at this stage. If it's broken we can fix it in the future.

ParthDesai: Make sense. So, how `-- --chain starlight-spec.json` is this handled? 
Since, We have hardcoded solo chain id right now.

Will you be addressing that as part of your todo?

tmpolaczyk: Oh my bad, that hardcoded relay chain id is actually not used, you can set it to `""` and it will still work. I can improve the TODO to mention this.

ParthDesai: But it seems to be used by `RelayChainCli`?

ParthDesai: It is clear now. Please mention this in the todo as this is not obvious. :)

Agusrodri (approved):

client/service-container-chain/src/service.rs
tmpolaczyk: *fixed*

client/service-container-chain/src/service.rs
tmpolaczyk: *fixed*

node/src/command.rs
tmpolaczyk: *fixed*

node/src/command/solochain.rs
tmpolaczyk: It's complicated, see the other discussion.

*disregard*

node/src/command/solochain.rs
tmpolaczyk: *disregard*

ParthDesai (approved):

PR #690, opened 2024-09-24
Starlight: Updated integration test run_block flow

chexware:
Updating Starlight runtime integration tests to use start_block + end_block run_block flow instead of the current end_block + start_block

### DIFF
--- a/solo-chains/runtime/starlight/src/tests/common/mod.rs
+++ b/solo-chains/runtime/starlight/src/tests/common/mod.rs
@@ -599,10 +599,6 @@ impl ExtBuilder {
         if let Some(keystore) = keystore {
             ext.register_extension(KeystoreExt(keystore));
         }
-        ext.execute_with(|| {
-            // Start block 1
-            start_block();
-        });
         ext
     }
 }
--- a/solo-chains/runtime/starlight/src/tests/common/mod.rs
+++ b/solo-chains/runtime/starlight/src/tests/common/mod.rs
@@ -240,9 +240,8 @@ pub fn end_block() {
 }
 
 pub fn run_block() {
-    end_block();
-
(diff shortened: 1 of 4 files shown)


solo-chains/runtime/starlight/src/tests/common/mod.rs
```diff
@@ -240,9 +240,8 @@ pub fn end_block() {
 }
 
 pub fn run_block() {
-    end_block();
-
-    start_block()
+    start_block();
+    end_block()
```
tmpolaczyk, line 244: I am not sure if we want this change, what was the reasoning behind it?

girazoki: We want to mimic the closest block production routine possible. WIthout this change we have situations where we have tests that require to manually end blocks because there are changes that happen on_finalize. So yes I think this change is worth pursuing

solo-chains/runtime/starlight/src/tests/relay_registrar.rs
```diff
@@ -45,7 +45,7 @@ fn registrar_needs_a_reserved_para_id() {
         .with_next_free_para_id(2000u32.into())
         .build()
         .execute_with(|| {
-            run_to_block(2);
+            //run_to_block(3);
```
girazoki, line 48: why this comment?

girazoki:
After analizing a bit more, it might make more sense maintain them as we had as it is the way parity also tests things..even if it means finalizing the state manually in some tests, soryr @chexware. At least we understand the implications now.

Maybe one more test we can add that this PR discovered: there is a ttl for voting for the valid wasm, and if it is not respected, the paralifecycle goes back to None. Let's add a test for that

PR #691, opened 2024-09-25
Bump Rust version to 1.81.0

nanocryk:
Using same version as https://github.com/paritytech/polkadot-sdk/pull/5676

We're currently on 1.78.0, and I'd like to use [this feature](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#bounds-in-associated-type-position) introduced in 1.79.0, which will allow to shorten a lot our `where`-clauses with trait aliases that can imply/infer associated types bounds.


solo-chains/client/cli/src/command.rs
```diff
@@ -547,11 +547,10 @@ fn load_spec(
...
             let path = std::path::PathBuf::from(path);
 
-            let chain_spec = Box::new(polkadot_service::GenericChainSpec::from_json_file(
+            (Box::new(polkadot_service::GenericChainSpec::from_json_file(
                 path.clone(),
-            )?) as Box<dyn polkadot_service::ChainSpec>;
-
-            chain_spec
+            )?) as Box<dyn polkadot_service::ChainSpec>)
+                as std::boxed::Box<dyn sc_cli::ChainSpec>
```
tmpolaczyk, line 553:
```suggestion
```

tmpolaczyk (commented):
Looking good, merge back master please

tmpolaczyk (approved):

nanocryk:
We seem to hit the same issue as described [in this Rust issue](https://github.com/rust-lang/rust/issues/125353), which seems to be a compiler regression. Until it is fixed, we can make progress by adding `--exclude tc-consensus` to `cargo test` inside the coverage CI job.

girazoki (approved):

PR #692, opened 2024-09-25
Combine `RemoveParaIdsWithNoCredit` and `CollatorAssignmentHook`

ParthDesai:
# Description
Combine `RemoveParaIdsWithNoCredit` and `CollatorAssignmentHook` in one trait with two methods.

Both method executes same logic. Only difference being `pre_assignment_*` method always rollback while `post_assignment_*` only rollback in case it errors out.

# Note
Naming and tests are still remaining. Just want to get some early feedback on overall design.

### DIFF
--- a/runtime/dancebox/src/lib.rs
+++ b/runtime/dancebox/src/lib.rs
@@ -115,8 +115,8 @@ use {
     },
     tp_traits::{
         apply, derive_storage_traits, GetContainerChainAuthor, GetHostConfiguration,
-        GetSessionContainerChains, MaybeSelfChainBlockAuthor, RelayStorageRootProvider,
-        RemoveInvulnerables, RemoveParaIdsWithNoCredits, SlotFrequency,
+        GetSessionContainerChains, MaybeSelfChainBlockAuthor, ParaIdAssignmentHooks,
+        RelayStorageRootProvider, RemoveInvulnerables, SlotFrequency,
     },
     tp_xcm_core_buyer::BuyCoreCollatorProof,
     xcm_runtime_apis::{
@@ -812,9 +812,9 @@ impl RemoveInvulnerables<CollatorId> for RemoveInvulnerablesImpl {
     }
 }
 
-pub struct RemoveParaIdsWithNoCreditsImpl;
+pub struct ParaIdAssignmentHooksImpl;
 
-impl RemoveParaIdsWithNoCreditsImpl {
+impl ParaIdAssignmentHooksImpl {
     fn charge_para_ids_internal(
         blocks_per_session: tp_traits::BlockNumber,
         para_id: ParaId,
@@ -888,11 +888,8 @@ impl RemoveParaIdsWithNoCreditsImpl {
     }
 }
 
-impl<AC> RemoveParaIdsWithNoCredits<BalanceOf<Runtime>, AC> for RemoveParaIdsWithNoCreditsImpl {
-    fn pre_assignment_remove_para_ids_with_no_credits(
-        para_ids: &mut Vec<ParaId>,
-        currently_assigned: &BTreeSet<ParaId>,
-    ) {
+impl<AC> ParaIdAssignmentHooks<BalanceOf<Runtime>, AC> for ParaIdAssignmentHooksImpl {
+    fn pre_assignment(para_ids: &mut Vec<ParaId>, currently_assigned: &BTreeSet<ParaId>) {
         let blocks_per_session = Period::get();
         para_ids.retain(|para_id| {
             with_transaction(|| {
@@ -909,7 +906,7 @@ impl<AC> RemoveParaIdsWithNoCredits<BalanceOf<Runtime>, AC> for RemoveParaIdsWit
         });
     }
 
-    fn post_assignment_remove_para_ids_with_no_credits(
+    fn post_assignment(
         current_assigned: &BTreeSet<ParaId>,
         new_assigned: &mut BTreeMap<ParaId, Vec<AC>>,
         maybe_tip: &Option<BalanceOf<Runtime>>,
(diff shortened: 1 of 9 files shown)


solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -2909,54 +2911,127 @@ impl frame_support::traits::Randomness<Hash, BlockNumber> for BabeCurrentBlockRa
 
 pub struct RemoveParaIdsWithNoCreditsImpl;
 
-impl RemoveParaIdsWithNoCredits for RemoveParaIdsWithNoCreditsImpl {
-    fn remove_para_ids_with_no_credits(
-        para_ids: &mut Vec<ParaId>,
+impl RemoveParaIdsWithNoCreditsImpl {
+    fn charge_para_ids_internal(
```
girazoki, line 2915: this is becoming hard to read, I would probably split it into different subfunctions that are public in the services payment pallet so that all this code is not in the runtime.

girazoki: also the naming of the function is not ideal, I would call it something like `charge_services_for_para_id`

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -2909,54 +2911,127 @@ impl frame_support::traits::Randomness<Hash, BlockNumber> for BabeCurrentBlockRa
...
+        )
+        .into_result(true)?;
+        // TODO: Have proper weight
+        Ok(Weight::zero())
+    }
+}
 
-            let remaining_to_pay = remaining_block_credits_to_pay.saturating_add(remaining_session_credits_to_pay).saturating_add(max_tip);
+impl<AC> RemoveParaIdsWithNoCredits<BalanceOf<Runtime>, AC> for RemoveParaIdsWithNoCreditsImpl {
+    fn pre_assignment_remove_para_ids_with_no_credits(
```
girazoki, line 2989: Here I would rename `pre_assignment_remove_para_ids_with_no_credits` to simply `pre_assignment_hook` or  `pre_assignment_filtering`

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -2909,54 +2911,127 @@ impl frame_support::traits::Randomness<Hash, BlockNumber> for BabeCurrentBlockRa
...
+                ))
+            })
+            .is_ok()
+        });
+    }
 
-            // This should take into account whether we tank goes below ED
-            // The true refers to keepAlive
-            Balances::can_withdraw(&pallet_services_payment::Pallet::<Runtime>::parachain_tank(*para_id), remaining_to_pay).into_result(true).is_ok()
+    fn post_assignment_remove_para_ids_with_no_credits(
```
girazoki, line 3009: same thing here, the naming is very specific to what this is doing. I would call it something like `post_assignmnet_hook`

runtime/dancebox/src/lib.rs
```diff
@@ -811,54 +816,127 @@ impl RemoveInvulnerables<CollatorId> for RemoveInvulnerablesImpl {
...
+        maybe_tip: &Option<BalanceOf<Runtime>>,
+    ) -> Weight {
+        let blocks_per_session = Period::get();
+        let mut total_weight = Weight::zero();
+        new_assigned.retain(|&para_id, collators| {
+            // Short-circuit in case collators are empty
+            if collators.is_empty() {
+                return true;
+            }
+            with_storage_layer(|| {
```
girazoki, line 926: what is `with_storage_layer`¿?

ParthDesai: So, `with_storage_layer` is similar to `with_transaction` but it automatically commits the storage changes into parent overlay if the closure returns `Ok` and do the rollback otherwise. 

Parity added this function since it is more convenient to use. (i.e you don't have to manually return `Ok` or `Err` for commit and rollback respectively.

runtime/dancebox/src/lib.rs
```diff
@@ -811,54 +816,127 @@ impl RemoveInvulnerables<CollatorId> for RemoveInvulnerablesImpl {
...
+                ))
+            })
+            .is_ok()
+        });
+    }
 
-            // This should take into account whether we tank goes below ED
-            // The true refers to keepAlive
-            Balances::can_withdraw(&pallet_services_payment::Pallet::<Runtime>::parachain_tank(*para_id), remaining_to_pay).into_result(true).is_ok()
+    fn post_assignment_remove_para_ids_with_no_credits(
```
tmpolaczyk, line 914: So can we remove this method? `pre_assignment` already uses `with_transaction` to remove para_ids that would have failed if `Self::charge_para_ids_internal` returns `Err`, so if I'm not wrong, calling `post_assignment` will never remove any additional para id?

runtime/dancebox/src/lib.rs
tmpolaczyk: Are there any performance concerns when running that function inside a loop? I guess not?

ParthDesai: Fairly low cost. See here: https://github.com/paritytech/substrate/pull/10809#issuecomment-1078343281

runtime/dancebox/src/lib.rs
ParthDesai: In perfect world yes. But, since we are changing state in between calling `pre` and `post` we need to run it again to eliminate even a tiniest possibility that some para_id escapes the restrictions.

girazoki (approved):
I like better the way it looks now. I am still missing some doc on what each of the lines does. If we can fix that would be great

runtime/dancebox/src/lib.rs
ParthDesai: *disregard*

tmpolaczyk (approved):

PR #703, opened 2024-09-30
fix author noting bench

girazoki:
Fixes the author-noting benchmark that was not able to run properly for dancelight.

We also forgot to kill the author-data upon deregistration, something that this PR fixes too. It should not impact that much unless we re-register the same container, in which case we can remove this data manually


solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1737,9 +1737,6 @@ impl pallet_registrar::RegistrarHooks for DancelightRegistrarHooks {
     }
 
     fn para_deregistered(para_id: ParaId) -> Weight {
-        // Clear pallet_author_noting storage
-        // TODO: uncomment when pallets exists
-        /*
```
tmpolaczyk, line 1742: Mention this change in the description

pallets/author-noting/src/benchmarks.rs
```diff
@@ -55,26 +58,26 @@ mod test_sproof {
...
+                for para_id in 1..x {
+                    let para_id = para_id.into();
+                    container_chains.push(para_id);
+                    // Mock assigned authors for this para id
+                    let author: T::AccountId = account("account id", 0u32, 0u32);
+                    // Use the max allowed value for num_each_container_chain
+                    let num_each_container_chain = 2;
+                    T::ContainerChainAuthor::set_authors_for_para_id(para_id, vec![author; num_each_container_chain]);
+                    sproof_builder.num_items += 1;
+                }
```
tmpolaczyk, line 78: There's some whitespace here

tmpolaczyk (approved):

solo-chains/runtime/dancelight/src/lib.rs
girazoki: yep

girazoki: fixed

pallets/author-noting/src/benchmarks.rs
girazoki: fixed

PR #706, opened 2024-10-01
Improved display of startup errors in node binaries

chexware:
Added `color_eyre` to all node binary entry points for improved display of startup errors.


container-chains/nodes/frontier/Cargo.toml
```diff
@@ -14,6 +14,7 @@ workspace = true
 async-io = { workspace = true }
 async-trait = { workspace = true }
 clap = { workspace = true, features = [ "derive" ] }
+color-eyre = { workspace = true, default-features = false }
```
tmpolaczyk, line 17:
```suggestion
color-eyre = { workspace = true }
```

Not needed here, it's already defined in the root Cargo.toml

tmpolaczyk (approved):

container-chains/nodes/frontier/Cargo.toml
chexware: Good point removed in [b16f442](https://github.com/moondance-labs/tanssi/pull/706/commits/b16f442af5544f9b4e9e13816a57cf39f75ec64b)

PR #708, opened 2024-10-01
Avoid ZeroCollators error in solochains

tmpolaczyk:
In dancebox, the pallet_collator_assignment has a check to ensure that we never return an assignment with 0 collators, because that would brick the chain. But in dancelight, that's not a problem because validators are stored in another pallet. So we can safely handle the case of 0 collators, by returning an empty assignment.

So we will no longer see this log when starting a dancelight node:

```
2024-10-01 17:03:07 Error in collator assignment, will keep previous assignment. ZeroCollators    
```

Also add tests to ensure this doesn't panic.


girazoki (approved):
thank you @tmpolaczyk looks good to me

PR #715, opened 2024-10-04
CI job compare runtime sizes

girazoki:


.github/workflows/release.yml
```diff
@@ -282,6 +283,90 @@ jobs:
...
+          run: cargo install twiggy
+        - name: Lookup for latest target branch build
+          run: |
+            TARGET_BRANCH=${{ github.event.pull_request.base.ref }}
+            LATEST_TARGET_BRANCH_BUILD=$(gh run -R moondance-labs/tanssi list -w CI --limit=100 --json databaseId,url,headBranch,event,status,conclusion,createdAt --jq ".[] | select(.headBranch == \"$TARGET_BRANCH\" and .event == \"push\" and .status == \"completed\" and .conclusion == \"success\") | .databaseId" | head -n 1)
+            echo "LATEST_TARGET_BRANCH_BUILD=$LATEST_TARGET_BRANCH_BUILD" >> $GITHUB_OUTPUT
+        - name: Download latest target branch build artifacts
+          run: |
+            gh run download $LATEST_TARGET_BRANCH_BUILD -n runtimes --dir runtimes-target-branch
+
```
ParthDesai, line 315: Extra space?

.github/workflows/release.yml
```diff
@@ -282,6 +283,90 @@ jobs:
...
+        - name: "Download branch built runtime"
+          uses: actions/download-artifact@v4
+          with:
+            name: runtimes
+            path: runtimes-current-branch
+        - name: Prepare file
+          run: |
+            # Create or truncate the file
+            echo "" > runtime_size_report.md
```
ParthDesai, line 324: If you are always overwriting the file. Is there any reason for appending?

.github/workflows/release.yml
```diff
@@ -282,6 +283,90 @@ jobs:
...
+            echo "" >> runtime_size_report.md
+            echo "#### Compared to target branch" >> runtime_size_report.md
+            echo "" >> runtime_size_report.md
+        - name: Retrieve runtime diff
+          run: |
+            IFS="," read -a RUNTIMES_TO_COMPARE <<< ${{ env.RUNTIMES_AS_STRING }}
+            get_status_emoji() {
+              local size=$1
+              local diff=$2
+              if [ $size -gt 2400 ]; then
```
ParthDesai, line 335: Should this be $diff?

girazoki: no, this warns you either when the runtime exceeds the 2.4MB or the diff is bigger than 0

.github/workflows/release.yml
girazoki: makes it easier to write "individual" messages

.github/workflows/release.yml
girazoki: I think we should only care in that case about the first one haha

ParthDesai (approved):

PR #720, opened 2024-10-10
Support Stream Payment for Data Preservers assignement

nanocryk:
Wires StreamPayment pallet to DataPreservers, allowing data provider to set a StreamPayment config in their DataPreservers profile, which can then be accepted by a para manager which will create a StreamPayment with that config. The data preserver will provide service as long as the stream payment is not stalled.

This PR includes a small change in pallet_registrar in set_para_manager: it now allows Root to set the para manager. In the test suites the paras are registered in genesis and thus they don't have a deposit with a creator, which prevents to set a para manager which is required to make a DataPreservers assignment (without forcing it with Root, which would skip the stream payment creation).


pallets/registrar/src/lib.rs
```diff
@@ -626,12 +626,15 @@ pub mod pallet {
             para_id: ParaId,
             manager_address: T::AccountId,
         ) -> DispatchResult {
-            let origin = ensure_signed(origin)?;
+            // Allow root to force set para manager.
+            if let Err(_) = ensure_root(origin.clone()) {
+                let origin = ensure_signed(origin)?;
```
tmpolaczyk, line 631: Maybe let's use `ensure_signed_or_root` here, to keep the pattern of the first line always being `let origin = ...`.

https://paritytech.github.io/polkadot-sdk/master/frame_system/fn.ensure_signed_or_root.html

tmpolaczyk (approved):
Tested using zombienet, works

client/service-container-chain/src/data_preservers.rs
```diff
@@ -37,6 +37,8 @@ pub async fn task_watch_assignment(spawner: impl Spawner, profile_id: ProfileId)
     use dc_orchestrator_chain_interface::DataPreserverAssignment as Assignment;
 
     if let OrchestratorChainResult::Err(e) = try_fut(async move {
+        log::info!("Starting data preserver assignment watcher");
```
girazoki, line 40: does this log in every block import? if so we maybe should consider moving it to debug?

client/service-container-chain/src/data_preservers.rs
```diff
@@ -52,11 +54,15 @@ pub async fn task_watch_assignment(spawner: impl Spawner, profile_id: ProfileId)
                 .data_preserver_active_assignment(hash, profile_id)
                 .await?;
 
-            log::info!("Assignement for block {hash}: {new_assignment:?}");
+            if current_assignment == new_assignment {
+                continue;
+            }
+
+            log::info!(
```
girazoki, line 61: same thing, or maybe even create a log target called data_preservers

tmpolaczyk: This only logs when the state changes, and the other log only once when the node starts, it's ok for me.

test/suites/data-preservers/test_data_preservers.ts
```diff
@@ -175,6 +214,122 @@ describeSuite({
...
+
+                const logFilePath = getTmpZombiePath() + "/DataPreserver-2001.log";
+                await waitForLogs(logFilePath, 300, ["NotAssigned => Active(Id(2000))"]);
+
+                const wsProvider = new WsProvider("ws://127.0.0.1:9952");
+                dataProvider2000BApi = await ApiPromise.create({ provider: wsProvider });
+
+                const newContainerNetwork = dataProvider2000BApi.consts.system.version.specName.toString();
+                const newParaId = (await dataProvider2000BApi.query.parachainInfo.parachainId()).toString();
+                expect(newContainerNetwork, "Container2000 API incorrect").to.contain("container-chain-template");
```
girazoki, line 320: can we check here that the payment is possible to trigger?

girazoki (approved):
good job jeremy, just minor comments!

client/service-container-chain/src/data_preservers.rs
nanocryk: It logs only at the start of data preserver mode. I can move it to debug tho.

client/service-container-chain/src/data_preservers.rs
nanocryk: I react on this log in the typescript tests. I can add a log target however

PR #723, opened 2024-10-16
Test: Offchain workers disabled by default for container chains 

chexware:
Adding tests to verify that offchain workers are disabled by default for container chains. Created dummy offchain worker pallet  that  submits an unsigned extrinsic and emits an event. The pallet is added it to the  simple container chain template and offchain actions must be disabled by default.

**TO DO:**

- [x] Add storage variable that enables disables calling offchain workers extrinsics. Set it to `false` by default
- [x] Add usigned extrinsics that emits event an can be called from the offchain worker if enabled
- [x] Update the tests to verify that the offchain worker pallet won't emit events after it is enabled


pallets/ocw-simple/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl<T: Config> Pallet<T> {
+    /// Simple event emitter
+    fn emit_offchain_event() {
+        Self::deposit_event(Event::SimpleOffchainEvent);
+    }
+
+    // Remove comment to enable http requests
+    /*
+    /// Fetch current price and return the result in cents.
```
girazoki, line 76: this should be removed right?

container-chains/runtime-templates/simple/src/lib.rs
```diff
@@ -701,6 +705,7 @@ construct_runtime!(
         RootTesting: pallet_root_testing = 100,
         AsyncBacking: pallet_async_backing::{Pallet, Storage} = 110,
 
+        OffchainWorker: pallet_ocw_simple::{Pallet,Call,Event<T>} = 120,
```
girazoki, line 708: What would happen if a data preserver enables this? would it just throw an event?

It's a pity that we cannot enable this just for testing :(

girazoki: I would call it pallet_roo_ocw

girazoki: Perhaps one way in which we can improve this is by adding a extrinsic that enables the offchain worker (by having a storage item in the pallet OffchainWorkerTestEnabled.

chexware: Good points... I will add the storage item and rename the pallet

pallets/ocw-simple/src/lib.rs
chexware: Yep not needed unless we want to specifically test http interactions

girazoki:
I think we need to test this better. I think we trully need to understand what the offchain worker does, which for starters should call a extrinsic of some sort to emit an event and you be able to detect it. In its current state I dont think we are testing much...

I think the "easiest" way to test this would be:

- add a storage item that enables the offchainWorker call or not in the pallet.
- add a extrinsic that the offchain worker calls and emits the event
- call this from unsigned?

container-chains/nodes/frontier/Cargo.toml
```diff
@@ -118,6 +118,8 @@ fc-storage = { workspace = true }
 fp-evm = { workspace = true }
 fp-rpc = { workspace = true }
 pallet-ethereum = { workspace = true }
+
+
```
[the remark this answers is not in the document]
chexware: Removed in https://github.com/moondance-labs/tanssi/pull/723/commits/2520247da194a15a0365e36f90070a636bdb7804

pallets/ocw-simple/src/lib.rs
chexware: Remove in https://github.com/moondance-labs/tanssi/pull/723/commits/e7ff792fad52b899b70dfb355ee15387442125bb

container-chains/runtime-templates/simple/src/lib.rs
chexware: Pallet renamed in https://github.com/moondance-labs/tanssi/pull/723/commits/e7ff792fad52b899b70dfb355ee15387442125bb . Added storage variable and extrinsic to switch on/off the offchain worker in https://github.com/moondance-labs/tanssi/pull/723/commits/bbee0863271446d03998a3cb071381f190c0a39d . Disabled offchain worker activity by default in [d6afff8](https://github.com/moondance-labs/tanssi/pull/723/commits/d6afff86e4149c6667af05d99e207638ae7922c6) .

chexware: Event emission using unsigned extrinic in [527989c](https://github.com/moondance-labs/tanssi/pull/723/commits/527989cdd1475f273200e0f6fa9fd9944ced680b)

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl<T: Config> Pallet<T> {
+    /// A helper function to sign payload and send an unsigned transaction
+    fn send_raw_unsigned_transaction(block_number: BlockNumberFor<T>) -> Result<(), &'static str> {
+        // Make sure transaction can be sent
+        let next_unsigned_at = NextUnsignedAt::<T>::get();
+        if next_unsigned_at > block_number {
+            return Err("Too early to send unsigned transaction");
+        }
+        let is_offchain_worker_enabled = OffchainWorkerTestEnabled::<T>::get();
```
girazoki, line 164: I would make this the first assertion

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            OffchainWorkerTestEnabled::<T>::put(!OffchainWorkerTestEnabled::<T>::get());
+            Ok(().into())
+        }
+
+        /// Submits unsigned transaction that emits an event
+        ///
+        /// Can be triggered only by an offchain worker
+        #[pallet::call_index(1)]
+        #[pallet::weight({0})]
```
girazoki, line 110: lets add at least a db write

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        pub fn submit_event_unsigned(
+            origin: OriginFor<T>,
+            _block_number: BlockNumberFor<T>,
+        ) -> DispatchResultWithPostInfo {
+            // This ensures that the function can only be called via unsigned transaction.
+            ensure_none(origin)?;
```
girazoki, line 116: we should only accept this if offchain worker mode is enabled?

test/suites/zombie-offchain/test-simple-template-offchain.ts
```diff
@@ -0,0 +1,79 @@
...
+            },
+        });
+
+        it({
+            id: "T02",
+            title: "Offchain events are not emitted for simple container chain when offchain pallet testing is enabled",
+            test: async function () {
+                const keyring = new Keyring({ type: "sr25519" });
+                const alice = keyring.addFromUri("//Alice", { name: "Alice default" });
+
```
girazoki, line 53: I am thinking whether we should add this to our main existing zombie test, I think i might be overkill to just add one single suite for this. What do you think @tmpolaczyk ?

girazoki: if we want this to be a separate suite, we shoudl at least run it when we merge to master in the CI

tmpolaczyk: Either way is ok, we have a job to manually run all the zombienet tests in ci.

If the offchain worker prints a log, we can use the existing test "Check collator logs to ensure common errors are fixed" to ensure that that particular log is not printed.

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        /// Number of blocks of cooldown after unsigned transaction is included.
+        ///
+        /// This ensures that we only accept unsigned transactions once, every `UnsignedInterval`
+        /// blocks.
+        #[pallet::constant]
+        type UnsignedInterval: Get<BlockNumberFor<Self>>;
+    }
+
+    #[pallet::storage]
+    pub(super) type OffchainWorkerTestEnabled<T> = StorageValue<_, bool, ValueQuery>;
```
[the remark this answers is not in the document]
tmpolaczyk: Yes

test/configs/zombieDancelightSingleContainer.json
```diff
@@ -0,0 +1,108 @@
...
+                    "prometheus_port": 33102
+                },
+                {
+                    "name": "Collator-05",
+                    "command": "../target/release/tanssi-node solo-chain",
+                    "args": ["--no-hardware-benchmarks", "--database=paritydb", "--wasmtime-precompiled=wasm"],
+                    "prometheus_port": 33102
+                },
+                {
+                    "name": "Collator-06",
```
tmpolaczyk, line 93: You can remove some collators to make the test startup faster, I believe the default is 2 collators per container chain?

test/suites/zombie-offchain/test-simple-template-offchain.ts
```diff
@@ -0,0 +1,79 @@
...
+                const blockNum = (await container2000Api.rpc.chain.getBlock()).block.header.number.toNumber();
+                expect(blockNum).to.be.greaterThan(0);
+                const switchTx = container2000Api.tx.offchainWorker.switchOffchainWorker();
+
+                // Enable off-chain worker test event emission
+                await signAndSendAndInclude(container2000Api.tx.sudo.sudo(switchTx), alice);
+                await context.waitBlock(20, "Container2000");
+
+                const events = await container2000Api.query.system.events();
```
tmpolaczyk, line 62: I believe these events will only be the events of the last block, so even if the offchain worker event is emitted this test may pass. Not sure if we have some util function to find events in the latest N blocks.

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        /// Switches on or off the offchain worker
+        ///
+        /// Only root (or specified authority account) should be able to switch
+        /// the off-chain worker on and off to avoid enabling it by default in production
+        #[pallet::call_index(0)]
+        #[pallet::weight({0})]
+        pub fn switch_offchain_worker(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
tmpolaczyk, line 99: Adding a bool parameter here will make this easier to use, `set` instead of `switch`

pallets/ocw-testing/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        let call = Call::submit_event_unsigned { block_number };
+
+        SubmitTransaction::<T, Call<T>>::submit_unsigned_transaction(call.into())
+            .map_err(|()| "Unable to submit unsigned transaction.")?;
+
+        Ok(())
+    }
+
+    fn validate_transaction_parameters(block_number: &BlockNumberFor<T>) -> TransactionValidity {
+        // Now let's check if the transaction has any chance to succeed.
```
girazoki, line 178: Add check that the testing mode is enabled

pallets/ocw-testing/src/lib.rs
tmpolaczyk: Why ignore? A collator can decide to include this extrinsic in any block, no?

pallets/ocw-testing/src/lib.rs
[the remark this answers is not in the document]
chexware: yes it is set to false on genesis [here](https://github.com/moondance-labs/tanssi/blob/1ef792a0e3e65cc962666d08ab72f7aa988e0e94/pallets/ocw-testing/src/lib.rs#L66)

pallets/ocw-testing/src/lib.rs
chexware: Good point... updated it here https://github.com/moondance-labs/tanssi/pull/723/commits/e66e8733d74019b4df5d905fb63cf8faa3423d17

test/configs/zombieDancelightSingleContainer.json
chexware: Updated here https://github.com/moondance-labs/tanssi/pull/723/commits/d263ed4eba8e7bfe55da406d567248f45b882f46

pallets/ocw-testing/src/lib.rs
chexware: Added weights to the extrinisics in [e6bcbff](https://github.com/moondance-labs/tanssi/pull/723/commits/e6bcbfff035a2474025ec268ba99f63d7c6ee722)

pallets/ocw-testing/src/lib.rs
chexware: Added in [2497476](https://github.com/moondance-labs/tanssi/pull/723/commits/2497476aad7fc12254795810cbc4ae078fab99b9)

pallets/ocw-testing/src/lib.rs
chexware: Updated order of checks in [2497476](https://github.com/moondance-labs/tanssi/pull/723/commits/2497476aad7fc12254795810cbc4ae078fab99b9)

test/suites/zombie-offchain/test-simple-template-offchain.ts
chexware: Yeah good point... I don't think we have a utility function for that but will create one!

test/suites/zombie-offchain/test-simple-template-offchain.ts
girazoki: so @chexware whatever you prefer to be honest,

test/suites/zombie-offchain/test-simple-template-offchain.ts
chexware: Added it in https://github.com/moondance-labs/tanssi/pull/723/commits/3faffe08fb84e13a580662ab13bf420dfcfdd40a

pallets/ocw-testing/src/lib.rs
tmpolaczyk: @chexware can you add this check here as well? Just in case

```rust
        // Make sure offchain worker testing is enabled
        let is_offchain_worker_enabled = OffchainWorkerTestEnabled::<T>::get();
        if !is_offchain_worker_enabled {
            return InvalidTransaction::Call.into();
        }
```

tmpolaczyk:
I was able to trigger a test failure by enabling offchain workers like this, good job!

```diff
diff --git a/client/node-common/src/service.rs b/client/node-common/src/service.rs
index 9500a431..cad274a0 100644
--- a/client/node-common/src/service.rs
+++ b/client/node-common/src/service.rs
@@ -552,7 +552,7 @@ where
 
         let collator = parachain_config.role.is_authority();
 
-        if parachain_config.offchain_worker.enabled {
+        if parachain_config.offchain_worker.enabled || true {
             task_manager.spawn_handle().spawn(
                 "offchain-workers-runner",
                 "offchain-work",
```

test/moonwall.config.json
```diff
@@ -646,6 +643,34 @@
                 }
             ]
         },
+        {
+            "name": "zombie_simple_template_offchain",
+            "timeout": 600000,
+            "testFileDir": ["suites/zombie-offchain"],
+            "runScripts": [
+                "build-spec-dancelight.sh",
```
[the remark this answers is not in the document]
chexware: Good spot... Created a new build script for it in https://github.com/moondance-labs/tanssi/pull/723/commits/7834cadcd441bba4c35371501090fc0cd85bcc10

test/suites/zombie-offchain/test-simple-template-offchain.ts
chexware: @girazoki seems that the current action for manually  running zombienet tests doesn't recognise the new moonwall config when ran it [here](https://github.com/moondance-labs/tanssi/actions/runs/11683481284). I assume that should work fine after merging the branch?

pallets/ocw-testing/src/lib.rs
chexware: Added error handling inside the extrinsic in https://github.com/moondance-labs/tanssi/pull/723/commits/d9ddc08ac7a5c17427c4cd4cf1763ff60218b52e. However, this should be handled by [validate_unsigned](https://github.com/moondance-labs/tanssi/blob/53fac57f7effd184f9d74aed3c0b32a90ef03af9/pallets/ocw-testing/src/lib.rs#L149).

tmpolaczyk (approved):

test/suites/zombie-offchain/test-simple-template-offchain.ts
girazoki: it is very likely that you need to checkout to the appropriate github commit, otherwise it is probably using master?

girazoki: where can I see the job you guys are referring to?

girazoki: alright the problem is clear, when you use the workflow dispatch from your branch, it only takes the action from that branch but not the codebase (I think). you can try to add as input the sha like we do with https://github.com/moondance-labs/tanssi/blob/f8b66f38f93dc8fb2a1bbe9d2e44d14d21cbd1ca/.github/workflows/prepare-binary.yml#L7

then check how the action goes into the sha you input

PR #725, opened 2024-10-21
Add wasm-precompiled feature to tanssi-relay binary

tmpolaczyk:
Speeds up running `dev_tanssi_relay` tests because now the wasm is only compiled once before running the tests, instead of once for every test suite.

Also speeds up `zombie_tanssi_relay`, but there we only have 4 tanssi-relay nodes so the effect is not noticeable.


tmpolaczyk:
[flamegraph_before](https://github.com/user-attachments/assets/65266dfe-f6dc-4178-994d-81dc2a5c9c8b)

[flamegraph_after](https://github.com/user-attachments/assets/d5aa7092-2a9d-499e-b176-ad0c8be03bd8)

test/moonwall.config.json
```diff
@@ -68,8 +71,9 @@
                             "--reserved-only",
                             "--alice",
                             "--tmp",
-                            "--chain=dancelight-dev",
+                            "--chain=tmp/dancelight-dev-raw.json",
```
girazoki, line 74: nice! maybe we should consider it adding it in the zombie tests?

tmpolaczyk: Done

*fixed*

girazoki (approved):
good job @tmpolaczyk

PR #726, opened 2024-10-21
MD603 - add SudoRegistrar proxy

augustocollado:
This PR adds a new proxy type _SudoRegistrar_ to the starlight runtime.
It allows proxy calls on DataPreservers (pallet_data_preservers), Registrar (paras_registrar), and ContainerRegistrar (pallet_registrar).

It also adds coverage tests.
Note: Some unrelated files are also modified to comply with fmt/prettier rules


test/suites/dev-tanssi-relay/proxy/test-proxy-registrar.ts
```diff
@@ -0,0 +1,252 @@
...
+                    genesisData,
+                    GENESIS_HEAD
+                );
+                await context.createBlock([await txRegisterTanssi.signAsync(charlie)]);
+
+                await jumpSessions(context, 1);
+
+                // Accept validation code so that para is onboarded after 2 sessions
+                const txAddsCode = polkadotJs.tx.sudo.sudo(
+                    polkadotJs.tx.paras.addTrustedValidationCode(VALIDATION_CODE)
```
tmpolaczyk, line 126: This call needs also needs sudo but is not supported by this proxy, should we add it @girazoki ?

girazoki: yes

girazoki: Although it is unlikely that we need to use it (nodes to this by default) better add it

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -806,6 +807,17 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
             }
             ProxyType::OnDemandOrdering => matches!(c, RuntimeCall::OnDemandAssignmentProvider(..)),
+            ProxyType::SudoRegistrar => match c {
+                RuntimeCall::Sudo(pallet_sudo::Call::sudo { call: ref x }) => {
+                    matches!(
+                        x.as_ref(),
+                        &RuntimeCall::DataPreservers(..)
+                            | &RuntimeCall::Registrar(..)
+                            | &RuntimeCall::ContainerRegistrar(..)
+                    )
```
girazoki, line 817: A tomasz says, I see a couple of them missing just in case,
- paras
- paraSudoWrapper

With those 2 I would be happy

girazoki (approved):
LGTM

girazoki:
the failing job is because we dont support it for external collaborators, so it's ok if we merge

tmpolaczyk (approved):

PR #727, opened 2024-10-22
External operator slashes info pallet

girazoki:
It adds the external operator slashes pallet:

**_Associated config traits_**:
- **_EraIndexProvider_**: which gives information about active era and when the era started
- **_Defer Period_**: a constant that represents the time that we have to cancel a slash
- **_BondingDuration_**: a constant that represents the time for which we have to maintain certain structures to be able to prove slashes. Obviously slashes are not maintained longer than the bonding duration.

**_Extrinsics_**:
- **_cancel_deferred_slash_**: an extrinsic that allows us to cancel an existing slash within the deferred period
- **_force_inject_slash_**: an extrinsic that allows us to forcedly inject a slash without having to verify it. Mostly for testing purposes.

**Pallet purpose**:
- Store a mapping between **eraIndex -> vec![Slash]**, where Slash contains information about the percentage to be slashed, the account, etc.

**Testing**:
Offences are tested by using both babe and grandpa equivocations, both in rust and typescript


pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+            let last_item = slash_indices[slash_indices.len() - 1];
+            ensure!(
+                (last_item as usize) < era_slashes.len(),
+                Error::<T>::InvalidSlashIndex
+            );
+
+            for (removed, index) in slash_indices.into_iter().enumerate() {
+                let index = (index as usize) - removed;
+                era_slashes.remove(index);
+            }
```
tmpolaczyk, line 198: This is a weird way to remove multiple elements from a vec. I know it's copied from polkadot but just do this instead:

```suggestion
            // Remove elements starting from the highest index to avoid shifting issues.
            for index in slash_indices.into_iter().rev() {
                era_slashes.remove(index);
            }
```

Performance could be improved by using Vec::retain, but that would be too complex.

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+                era
+            } else {
+                era.saturating_add(slash_defer_duration)
+                    .saturating_add(One::one())
+            };
+
+            let mut era_slashes = Slashes::<T>::get(&era_to_consider);
+            era_slashes.push(slash);
+
+            Slashes::<T>::insert(era_to_consider, &era_slashes);
```
tmpolaczyk, line 245: Slashes::mutate?

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+{
+    fn on_offence(
+        offenders: &[OffenceDetails<
+            T::AccountId,
+            pallet_session::historical::IdentificationTuple<T>,
+        >],
+        slash_fraction: &[Perbill],
+        slash_session: SessionIndex,
+    ) -> Weight {
+        let mut consumed_weight = Weight::from_parts(0, 0);
```
tmpolaczyk, line 276: Can we avoid the need to return a weight from this function? Not a fan of manually counting storage reads and writes.

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+// (at your option) any later version.
+
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
```
Agusrodri, line 16: We could add a few lines of docs here, giving an overview of the pallet's purpose.

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+            + Eq
+            + Saturating
+            + One
+            + Ord
+            + MaxEncodedLen;
+
+        /// Interface for interacting with a session pallet.
+        type SessionInterface: SessionInterface<Self::AccountId>;
+
+        /// Era index provider, used to fetch the active era among other thins
```
Agusrodri, line 108:
```suggestion
        /// Era index provider, used to fetch the active era among other things
```

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        /// The weight information of this pallet.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        EmptyTargets,
+        InvalidSlashIndex,
+        NotSortedAndUnique,
```
Agusrodri, line 122: Would be nice to have at least one short line of explanation docs for each error.

pallets/external-validator-slashes/src/mock.rs
```diff
@@ -0,0 +1,240 @@
...
+        1,
+        2,
+        3,
+    ]);
+}
+
+pub struct TestSessionManager;
+impl pallet_session::SessionManager<u64> for TestSessionManager {
+    fn new_session(_new_index: SessionIndex) -> Option<Vec<u64>> {
+        Validators::mutate(|l| l.take())
```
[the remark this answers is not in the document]
girazoki: yes I think it can be a constant

pallets/external-validator-slashes/src/lib.rs
girazoki: I can check but if it is meaningful to be used (i.e., ther weight charged when you insert an offence) then there is not much we can do. At best I can try to benchmark it, would that be better?

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+        });
+
+        Self::confirm_unconfirmed_slashes(era_index);
+    }
+}
+
+impl<T: Config> Pallet<T> {
+    /// Apply previously-unapplied slashes on the beginning of a new era, after a delay.
+    fn confirm_unconfirmed_slashes(active_era: EraIndex) {
+        let mut era_slashes = Slashes::<T>::take(&active_era);
```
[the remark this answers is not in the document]
girazoki: fixed

pallets/external-validator-slashes/src/mock.rs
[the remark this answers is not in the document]
girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+
+                // Prune out everything that's from before the first-kept index.
+                let n_to_prune = bonded
+                    .iter()
+                    .take_while(|&&(era_idx, _)| era_idx < first_kept)
+                    .count();
+
+                // Kill slashing metadata.
+                for (pruned_era, _) in bonded.drain(..n_to_prune) {
+                    let _ = ValidatorSlashInEra::<T>::clear_prefix(&pruned_era, REMOVE_LIMIT, None);
```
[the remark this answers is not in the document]
girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
tmpolaczyk: Nice so let's remove it. And no need to benchmark this here, the weight will be included in the benchmark of whatever calls `on_offence` (right?).

pallets/external-validator-slashes/src/lib.rs
girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
girazoki: yes in theory yes (if those extrinsics have benchmarks)

girazoki: https://github.com/paritytech/polkadot-sdk/blob/b326540184a1c87c141c169896c2f0e58d41e24f/polkadot/runtime/rococo/src/lib.rs#L368

Those pallets use the default weightInfo in polkadot runtimes (and so are we)

girazoki: so I will remove it for now, but we need to track whether we need to track whether we need to properly benchmark those extrinsics

girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
girazoki: fixed

pallets/external-validator-slashes/src/lib.rs
girazoki: fixed

Agusrodri (approved):
Looks good to me!

tmpolaczyk (approved):

PR #731, opened 2024-10-24
add first job for testing starliht ugrade through zombie

girazoki:
Adds the first job to test starlight upgrades through zombie, but I encountered the following issues:

- first, I need to specify a parachain otherwise moonwall does not work with non-parachain zombienet configs.
- second, I needed to specify as provider name : "parachain" otherwise it does not allow me to do `runtimeUpgrade`


test/configs/zombieDancelightUpgrade.json
```diff
@@ -0,0 +1,66 @@
...
+                "ws_port": "9947",
+                "validator": true
+            },
+            {
+                "name": "bob",
+                "validator": true
+            }
+        ]
+    },
+    "parachains": [
```
girazoki, line 40: this is needed because moonwall does not accept zombienet configs without parachains

tmpolaczyk: Would be nice to have a comment in the file, but this is json

tmpolaczyk (approved):

test/configs/zombieDancelightUpgrade.json
girazoki: I will make sure @timbrinded is aware once he comes back

PR #737, opened 2024-10-31
Add bridge testing environment

ParthDesai:
# Description
This PR adds bridge testing environment through series of shell scripts. The environment can be used to write bridging related tests in moonwall.


test/suites/zombie-tanssi-relay-eth-bridge/test-consensus-bridge.ts
```diff
@@ -0,0 +1,156 @@
...
+            title: "Ethereum Blocks are being recognized on tanssi-relay",
+            test: async function () {
+                await sleep(10000);
+                const firstFinalizedBlockRoot = (
+                    await relayApi.query.ethereumBeaconClient.latestFinalizedBlockRoot()
+                ).toJSON();
+                expect(firstFinalizedBlockRoot).to.not.equal(
+                    "0x0000000000000000000000000000000000000000000000000000000000000000"
+                );
+                await sleep(20000);
```
girazoki, line 114: we can possibly wait for blocks instead of time here. Approximately every how many blocks is this injected?

ParthDesai: *fixed*

girazoki (approved):
looks good parth, good job!

test/suites/zombie-tanssi-relay-eth-bridge/test-consensus-bridge.ts
```diff
@@ -0,0 +1,156 @@
...
+            const beaconRelay = keyring.addFromUri("//BeaconRelay", { name: "Beacon relay default" });
+
+            const txHash = await relayApi.tx.balances
+                .transferAllowDeath(beaconRelay.address, 1_000_000_000_000)
+                .signAndSend(alice);
+            console.log("Transferred money to beacon relay", txHash.toHex());
+
+            ethereumNodeChildProcess = spawn("./scripts/bridge/start-ethereum-node.sh", {
+                shell: true,
+                detached: true,
```
tmpolaczyk, line 47: Why detached? I believe this only affects whether the child process is killed when the parent process dies. We don't want any processes running after the tests are finished, right?

```suggestion
```

test/scripts/bridge/cleanup.sh
```diff
@@ -0,0 +1,45 @@
...
+
+  beacon_relay=""
+  beefy_relay=""
+
+  # Source daemons.pid if it exists
+  source $artifacts_dir/daemons.pid 2> /dev/null || true
+
+  # Using interrupt instead of kill signal for process to cleanup
+  kill -s INT $beacon_relay 2> /dev/null || true
+  kill -s INT $beefy_relay 2> /dev/null || true
```
tmpolaczyk, line 35: Instead of storing the PID in a file I would have implemented the `start-*.sh` scripts as javascript functions, then this would have been much easier.

But if it works don't touch it

test/scripts/bridge/set-env.sh
```diff
@@ -0,0 +1,242 @@
...
+              exit 1
+          fi
+      else
+          if ! [ -x "$(command -v date)" ]; then
+              echo 'Error: date is not installed.'
+              exit 1
+          fi
+    fi
+
+    check_node_version 22
```
tmpolaczyk, line 230: I have node 20 so I changed this line and it works, where does the requirement to use node 22 come from?

test/suites/zombie-tanssi-relay-eth-bridge/test-consensus-bridge.ts
ParthDesai: The shell script has its own cleanup mechanism that traps some signals and upon receiving the signal, kills all the process in the same process group via `kill 0`. 

I am adding `detached: true` so that process group id is different than the moonwall process. So, that all process except moonwall can be terminated by `kill 0`. If I do not pass it, the `kill 0` in script will terminate moonwall process as well.

test/scripts/bridge/set-env.sh
ParthDesai: I just put the latest LTS version.

tmpolaczyk (approved):

test/suites/zombie-tanssi-relay-eth-bridge/test-consensus-bridge.ts
tmpolaczyk: First time I hear about process group ids, ok let's keep it detached then

ParthDesai: *resolved*

PR #738, opened 2024-10-31
Remove unneeded storage version from pallet invulnerables

tmpolaczyk:
TODO: make sure this doesn't break anything


girazoki:
it's not unneeded, it makes running migrations a bit more robust

tmpolaczyk:
Not clear if we should remove this constant or add it to all the other pallets

PR #745, opened 2024-11-14
Starlight: add smoke test for BEEFY and MMR digests

Agusrodri:
This PR adds a smoke test which checks that both BEEFY and MMR root digests are correct and consistent.


girazoki (approved):
good job @Agusrodri

PR #746, opened 2024-11-14
Implement container db folder cleanup on startup

tmpolaczyk:
Container database folders are normally removed when the collator gets unassigned. However, if the collator stops the node right before being unassigned, and then starts it later, the container db will not be removed. The db will be removed the next time the collator is assigned and unassigned from that container chain, which may never happen. 

To fix this edge case, this PR introduces a new db folder cleanup functionality. When the node starts, it will check the container folders and delete all of them except the ones it is assigned to as a collator. This check is only performed once, the first time a finalized block is imported.


girazoki:
` When the node starts, it will check the container folders and delete all of them except the ones it is assigned to as a collator.`

What happens if I stop it while it is syncing a chain (2000) but is assigned to another one (2001), then I boot it up again and we are still in the same situation? would I remove 2000?

tmpolaczyk:
> What happens if I stop it while it is syncing a chain (2000) but is assigned to another one (2001), then I boot it up again and we are still in the same situation? would I remove 2000?

"assigned" means "current or next", so obviously it won't be removed

nanocryk (approved):

PR #747, opened 2024-11-15
Container candidates in dev service

girazoki:
Adds the ability to inject container candidates on dancelights dev service. For now this remains as an opt-in that you can enable in your tests by calling the following rpc in dev mode:

- `mock_enableParaInherentCandidate`

similarly you can disable it by calling:
- `mock_disableParaInherentCandidate`

I also added a test that proves that when enabled, this mode makes the para-head progress in the relay, indicating the candidates are being inserted.

TODO in this PR:
- [x] Cleanup
- [x] tests with rewards being stored

Follow-up tests for future PRs would be:
- AuthorNoting properly tested: now that we can inject proper headers, we should be able to properly test author-noting inherent, since we need the header to retrieve the author.
- XCM tests: now that we have the ability to inject candidates we should also be able to put UMP messages in the candidates.


solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[method(name = "mock_enableParaInherentCandidate")]
+    async fn enable_para_inherent_candidate(&self) -> RpcResult<()>;
+
+    /// Indicate the mock parachain candidate insertion to be disabled
+    #[method(name = "mock_disableParaInherentCandidate")]
+    async fn disable_para_inherent_candidate(&self) -> RpcResult<()>;
+}
+
+pub struct DevRpc {
+    pub mock_para_inherent_channel: flume::Sender<Vec<u8>>,
```
tmpolaczyk, line 44: I don't understand why does this need to be a channel, I would simply use an `AtomicBool`, or if you need to clone it `Arc<AtomicBool>`

```suggestion
    pub mock_para_inherent_enabled: AtomicBool,
```

solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -630,3 +945,44 @@ fn new_partial_basics(
...
+) -> Vec<u8> {
+    // equivalent to (`real_payload`, context).encode()
+    let mut out = payload.encode_as();
+    out.extend(context.encode());
+    out
+}
+
+/// Create an `AvailabilityBitfield` where `concluding` is a map where each key is a core index
+/// that is concluding and `cores` is the total number of cores in the system.
+fn availability_bitvec(used_cores: usize, cores: usize) -> AvailabilityBitfield {
```
tmpolaczyk, line 977:
```suggestion
/// Create an `AvailabilityBitfield` with size `total_cores`. The first `used_cores` set to true (occupied),
/// and the remaining to false (available).
fn availability_bitvec(used_cores: usize, total_cores: usize) -> AvailabilityBitfield {
```

solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -630,3 +945,44 @@ fn new_partial_basics(
...
+        .sr25519_sign(ValidatorId::ID, key.as_ref(), &data)?
+        .map(|sig| UncheckedSigned::new(payload, validator_index, sig.into()));
+    Ok(signature)
+}
+
+fn payload_data<H: Encode, Payload: Encode>(
+    payload: &Payload,
+    context: &SigningContext<H>,
+) -> Vec<u8> {
+    // equivalent to (`real_payload`, context).encode()
```
tmpolaczyk, line 969: Then why use the other way?

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
girazoki: The intention of this channel is to not write to the storage item every block, but rather only when the user requests it through the rpc. hence the channel

solo-chains/node/tanssi-relay-service/src/dev_service.rs
girazoki: I think in the end it does not really matter much, but I copied this from polkadot, so I guess let's maintain it this way

solo-chains/node/tanssi-relay-service/src/dev_service.rs
girazoki: solved

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
tmpolaczyk: Ok makes sense

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
```diff
@@ -0,0 +1,83 @@
...
+        log::info!("entering here");
+        let mock_para_inherent_channel = self.mock_para_inherent_channel.clone();
+        // Push the message to the shared channel where it will be queued up
+        // to be injected in to an upcoming block.
+        mock_para_inherent_channel
+            .send_async(true.encode())
+            .await
+            .map_err(|err| internal_err(err.to_string()))?;
+
+        log::info!("SENEDING ENABLE");
```
tmpolaczyk, line 59: Remove debug logs

tmpolaczyk (approved):

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
girazoki: ignore

solo-chains/node/tanssi-relay-service/src/dev_service.rs
girazoki: ignore

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
girazoki: resolved

PR #749, opened 2024-11-15
Add smoke tests for BABE

tmpolaczyk:


test/suites/smoke-test-dancelight/test-babe.ts
```diff
@@ -0,0 +1,62 @@
...
+                    (log) => log.isConsensus === true && log.asConsensus[0].toHex() == stringToHex("BABE")
+                );
+                expect(filteredDigests.length).to.eq(1);
+
+                // 0x01 corresponds to ConsensusLog::NextEpochData enum variant.
+                expect(filteredDigests[0].asConsensus[1].toHex().startsWith("0x01")).to.be.true;
+
+                // Assert that authorities from log == authorities from pallet
+                const babeAuthoritiesFromPallet = await api.query.babe.authorities();
+                const asdas = api.registry.createType(
```
girazoki, line 36: please the variable name

girazoki:
Can we go back 100 blocks to verify the authors are correct? I think we do something similar with tanssi and aura, see https://github.com/moondance-labs/tanssi/blob/master/test/suites/smoke-test-common-parachain/test-block-author-logs.ts

However I guess you will need to compute a vrf verification, so not sure how easy this is from typescript..

girazoki (approved):
Description of the PR please!

tmpolaczyk:
> However I guess you will need to compute a vrf verification, so not sure how easy this is from typescript..

I was not able to do this because I don't know how to create the message to be verified. It's not a simple concatenation of bytes but something more complex, and with no javascript impl, see https://docs.rs/crate/merlin/latest/source/src/transcript.rs#69

girazoki:
> > However I guess you will need to compute a vrf verification, so not sure how easy this is from typescript..
> 
> I was not able to do this because I don't know how to create the message to be verified. It's not a simple concatenation of bytes but something more complex, and with no javascript impl, see https://docs.rs/crate/merlin/latest/source/src/transcript.rs#69

yes for now let's leave it out of scope

test/moonwall.config.json
```diff
@@ -919,7 +919,6 @@
             "foundation": {
                 "type": "read_only"
             },
-            "reporters": ["html"],
```
tmpolaczyk, line 922: undo this

test/suites/smoke-test-dancelight/test-babe.ts
tmpolaczyk: *fixed*

PR #756, opened 2024-11-19
Add smoke tests para inclusion

tmpolaczyk:


test/moonwall.config.json
```diff
@@ -919,7 +919,6 @@
             "foundation": {
                 "type": "read_only"
             },
-            "reporters": ["html"],
```
tmpolaczyk, line 922: undo before merge

girazoki (approved):
nice one!

test/moonwall.config.json
tmpolaczyk: *fixed*

PR #760, opened 2024-11-25
add services payment test

girazoki:
Adds tests with respect to services payment because now we have a way to simulate container candidates. The tests checks that the block credits are getting spent and container chains deassigned when they run out of creds


test/suites/dev-tanssi-relay/services-payment/test_services_payment_block_credits.ts
```diff
@@ -0,0 +1,275 @@
...
+
+                const credits2 = (await polkadotJs.query.servicesPayment.blockProductionCredits(paraId)).toJSON();
+                const containerBlockNum2 = await (await polkadotJs.query.authorNoting.latestAuthor(paraId)).toJSON()
+                    .blockNumber;
+                expect(containerBlockNum1, "container chain 2000 did not create a block").toBeLessThan(
+                    containerBlockNum2
+                );
+                expect(credits1, "container chain 2000 created a block without burning any credits").toBeGreaterThan(
+                    credits2
+                );
```
tmpolaczyk, line 90: Can we also test that `credits1 - credits2 == containerBlockNum2 - containerBlockNum1`? I'm assuming that 1 block created is always 1 credit

girazoki: done

tmpolaczyk (approved):

PR #761, opened 2024-11-25
Starlight <> Ethereum: Rewards mapping

Agusrodri:
### What does it do?

This PR takes care of generating a specific command called `ReportRewards`, which is used to map and encode rewards data happening in Starlight. This information will then be used by Symbiotic smart contracts on the Ethereum side to perform specific computation/checks related to rewards.

More specifically, the `ReportRewards` command has the following structure:

```rust
    ReportRewards {
        // block timestamp
        timestamp: u64,
        // index of the era we are sending info of
        era_index: u32,
        // total_points for the era
        total_points: u128,
        // new tokens inflated during the era
        tokens_inflated: u128,
        // merkle root of vec![(validatorId, rewardPoints)]
        rewards_merkle_root: H256,
    }
```

### Addition of runtime api

In this PR, a `ExternalValidatorsRewardsApi` runtime api was added, containing two main functions:

```rust
sp_api::decl_runtime_apis! {
    pub trait ExternalValidatorsRewardsApi<AccountId, EraIndex>
    where
        AccountId: parity_scale_codec::Codec,
        EraIndex: parity_scale_codec::Codec,
    {
        fn generate_rewards_merkle_proof(account_id: AccountId, era_index: EraIndex) -> Option<MerkleProof>;
        fn verify_rewards_merkle_proof(merkle_proof: MerkleProof) -> bool;
    }
}
```

Where:
- `generate_rewards_merkle_proof`: generates a merkle proof for a given `account_id` and `era_index`. If the `account_id` was not rewarded in that era or was not present in the validator set for some reason, it returns `None`.
- `verify_rewards_merkle_proof`: verifies a merkle proof (that could be previously generated by calling _generate_rewards_merkle_proof_).


pallets/external-validators-rewards/src/lib.rs
```diff
@@ -104,6 +138,75 @@ pub mod pallet {
...
+        //  - rewards_merkle_root: merkle root corresponding [(validatorId, rewardPoints)]
+        //      for the era_index specified.
+        //  - leaves: that were used to generate the previous merkle root.
+        //  - leaf_index: index of the validatorId's leaf in the previous leaves array (if any).
+        //  - total_points: number of total points of the era_index specified.
+        pub fn generate_era_rewards_utils(
+            era_index: EraIndex,
+            maybe_account_id_check: Option<T::AccountId>,
+        ) -> Option<EraRewardsUtils> {
+            let era_rewards = RewardPointsForEra::<T>::get(&era_index);
```
tmpolaczyk, line 152: We could check if the account exists before entering the loop:

```rust
if let Some(account) = &maybe_account_id_check {
    if !era_rewards.individual.contains_key(account) {
        log::error!(
            target: "ext_validators_rewards",
            "AccountId {:?} not found for era {:?}!",
            account,
            era_index
        );
        return None;
    }
}
```

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1335,17 +1336,39 @@ impl pallet_external_validators::Config for Runtime {
     type UnixTime = Timestamp;
     type SessionsPerEra = SessionsPerEra;
     type OnEraStart = (ExternalValidatorSlashes, ExternalValidatorsRewards);
-    type OnEraEnd = ();
+    // reminder to add on_initialize weight
+    type OnEraEnd = ExternalValidatorsRewards;
```
tmpolaczyk, line 1340: How do we account for this? OnEraEnd is run in on_initialize but we have no way to return the weight, and the pallet doesn't know when does the era end so we can't just add the weight to on_initialize.

Should we return the weight from the OnEraEnd method? Or just benchmark it and call register_extra_weight_unchecked?

Agusrodri: I was just thinking about that. In my opinion I would return the weight from the `OnEraEnd` method, and I guess in that case we would also need to modify the `on_initialize` method where this is called.

pallets/external-validators-rewards/src/lib.rs
Agusrodri: fixed in b36213efa4e8fbc36c6b6642d8bb98f00cd9b239.

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -104,6 +138,75 @@ pub mod pallet {
...
+            let total_points: u128 = era_rewards.total as u128;
```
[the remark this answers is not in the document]
Agusrodri: This conversion is basically to retrieve a `u128` in the `total_points` field of `ReportRewards` command (as it was discussed internally). 

Should we change the `total_points` field to `u32`?

solo-chains/runtime/dancelight/src/lib.rs
Agusrodri: After analyzing it a bit more, I think it's better to go for the benchmarking option. I'll be working on that.

Agusrodri: Added the register_extra_weight_unchecked function with the proper benchmarked weight in 74537a6c8ae42cd99f4ae5a0e6f2c50ff971d8a0.

Agusrodri: _fixed_

primitives/bridge/src/lib.rs
```diff
@@ -67,6 +67,18 @@ mod custom_send_message;
 pub enum Command {
     // TODO: add real commands here
     Test(Vec<u8>),
```
ParthDesai, line 69: Should we remove `Test`?

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -115,6 +228,53 @@ pub mod pallet {
...
+                        if let Err(err) = T::OutboundQueue::deliver(ticket) {
+                            log::error!(target: "xcm::ethereum_blob_exporter", "OutboundQueue delivery of message failed. {err:?}");
+                        }
+                    }
+                    Err(err) => {
+                        log::error!(target: "xcm::ethereum_blob_exporter", "OutboundQueue validation of message failed. {err:?}");
+                    }
+                }
+
+                frame_system::Pallet::<T>::register_extra_weight_unchecked(
```
ParthDesai, line 263: This call is dangerous. But since the weight is static and this being a relay chain, it should not cause an issue in practice.

pallets/external-validators-rewards/src/lib.rs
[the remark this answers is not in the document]
ParthDesai: I would say keep it at `u128`. It's better to have big datatype rather than having small type first and then making it bigger.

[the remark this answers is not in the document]
Agusrodri: _disregard_

primitives/bridge/src/lib.rs
Agusrodri: I would keep it for now. I think it's still useful if we want to test message sending in a general way.

pallets/external-validators-rewards/src/lib.rs
Agusrodri: Yes, I agree.

Agusrodri: _disregard_

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -115,6 +228,53 @@ pub mod pallet {
...
+                    id: None,
+                    channel_id,
+                    command,
+                };
+
+                // Validate and deliver the message
+                match T::ValidateMessage::validate(&outbound_message) {
+                    Ok((ticket, _fee)) => {
+                        if let Err(err) = T::OutboundQueue::deliver(ticket) {
+                            log::error!(target: "xcm::ethereum_blob_exporter", "OutboundQueue delivery of message failed. {err:?}");
```
girazoki, line 255: this is not ethereum blob exporter, we should change the target

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -115,6 +228,53 @@ pub mod pallet {
...
+
+                frame_system::Pallet::<T>::register_extra_weight_unchecked(
+                    T::WeightInfo::on_era_end(),
+                    DispatchClass::Mandatory,
+                );
+            } else {
+                // Unreachable, this should never happen as we are sending
+                // None as the second param in Self::generate_era_rewards_utils.
+                log::error!(
+                    target: "ext_validators_rewards",
```
girazoki, line 271: here you have the correct target haha

girazoki (approved):
minor comments, the rest looks good to me

nanocryk (approved):

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1335,17 +1336,39 @@ impl pallet_external_validators::Config for Runtime {
...
+pub struct TimestampProvider;
+impl Get<u64> for TimestampProvider {
+    fn get() -> u64 {
+        Timestamp::get()
+    }
+}
+
+// TODO: add a proper way to retrieve the inflated tokens.
+// Will likely be through InflationRewards.
+pub struct EraInflationProvider;
```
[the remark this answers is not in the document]
Agusrodri: Yes it does. I put the explicit implementation to emphasize on the TODO, but I'll remove it.

pallets/external-validators-rewards/src/lib.rs
Agusrodri: _fixed_ in 96da5636be9480807c41c365fb26f9b6c12c2e7a.

solo-chains/runtime/dancelight/src/lib.rs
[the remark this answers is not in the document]
Agusrodri: _fixed_ in 96da5636be9480807c41c365fb26f9b6c12c2e7a.

primitives/bridge/src/lib.rs
Agusrodri: @ParthDesai what do you think?

ParthDesai: Okay. Let's keep it for now.

ParthDesai: *fixed*

primitives/bridge/src/custom_do_process_message.rs
```diff
@@ -151,6 +151,7 @@ impl ConstantGasMeter {
     fn maximum_dispatch_gas_used_at_most(command: &Command) -> u64 {
         match command {
             Command::Test { .. } => 60_000,
+            Command::ReportRewards { .. } => 60_000,
```
ParthDesai, line 154: Minor: But add TODO here to revisit the gas cost value.

Agusrodri: _fixed_

PR #765, opened 2024-11-27
Fork-aware tx-pool polkadot2409

chexware:


girazoki:
Can we add --txpool=fork-aware in all zombienet jobs where we have collators?

PR #766, opened 2024-11-28
Add zombienet suite for unneeded para bug

tmpolaczyk:
Does not run in CI because it fails.

To reproduce this bug we need a collator that:
* Creates block for a para
* That para runs out of credits
* Validator keeps banning that collator


girazoki:
UPdate description and labels please,

girazoki (approved):

PR #770, opened 2024-11-28
Allow rotating a subset of collators instead of all of them

tmpolaczyk:
Adds new field to pallet_configuration: full_rotation_mode. Allows to specify how to perform the rotation, allowing a different value for each kind of chain (orchestrator, parachain, parathread):

* Default: RotateAll, will rotate all collators (as before this PR)
* KeepAll (same as disabling rotation, but now can be set only for parathreads or only for orchestrator for example)
* KeepCollators: will keep N collators assigned on every rotation. If the collators are still eligible, this guarantees there will be N collators that don't rotate.
* KeepPerbill: same but allows to easily set it to 50%, allowing to change the number of collators per chain without having to also update this value. This refers to the % respect to the max number of collators, so if max_orchestrator = 5 but there are only 2 collators assigned to the orchestrator chain, and we try to keep 50%, then it will try to keep 2.5 collators, so all of them.

Contains migration for configuration pallet. I had to fix the migration because it used a hardcoded key that depends on pallet name, thus failing in dancelight.


pallets/collator-assignment/src/assignment.rs
```diff
@@ -142,6 +139,48 @@ where
...
+            None => return,
+        };
+
+        let num_to_keep = match full_rotation_mode {
+            FullRotationMode::RotateAll => 0,
+            FullRotationMode::KeepAll => {
+                return;
+            }
+            FullRotationMode::KeepCollators { keep } => keep,
+            FullRotationMode::KeepPerbill { keep } => keep * max_collators,
```
girazoki, line 162:
```suggestion
            FullRotationMode::KeepPerbill { percentage } => percentage * max_collators,
```

pallets/collator-assignment/src/assignment.rs
```diff
@@ -492,4 +531,5 @@ pub struct ChainNumCollators {
     pub min_collators: u32,
     // This will only be filled if all the other min have been reached
     pub max_collators: u32,
+    pub parathread: bool,
```
girazoki, line 534: what does this boolean indicate? whether the chain is a parathread? if so add comment

pallets/collator-assignment/src/lib.rs
```diff
@@ -398,47 +402,43 @@ pub mod pallet {
...
-                        chains,
-                        shuffle_collators,
-                    )
-                } else {
-                    log::debug!(
-                        "Collator assignment: keep old assigned. Session {:?}, Seed: {:?}",
-                        current_session_index.encode(),
-                        random_seed
-                    );
+            Self::deposit_event(Event::NewPendingAssignment {
```
girazoki, line 421: it might be a breaking change, but I think it would be good to see in the event the rotation type that happened

girazoki: Also it feels a little bit awkward to through the event before all changes happen no? I know we were doing this in the past, but maybe it's a good time to push the event a little be downwards

pallets/collator-assignment/src/assignment.rs
tmpolaczyk: What else could it indicate? xD

pallets/collator-assignment/src/lib.rs
tmpolaczyk: Yeah, I was going to change it and emit the event after, but let's do it in a separate PR

runtime/common/src/migrations.rs
```diff
@@ -145,20 +150,7 @@ where
...
-                .expect("configuration.activeConfig should have value");
-        // This works because there is no enum in the v2
-        assert_eq!(
-            old_config_bytes.len(),
-            HostConfigurationV2::default().encoded_size()
-        );
-
-        use parity_scale_codec::Encode;
-        Ok((old_config_bytes).encode())
+        Ok(vec![])
```
girazoki, line 153: let's try to do the pre and post upgrade hooks no?

pallets/collator-assignment/src/lib.rs
tmpolaczyk: The type depends on the chain, do you want the event to contain the full "FullRotationModes" struct?

solo-chains/runtime/dancelight/src/tests/mod.rs
```diff
@@ -68,3 +73,116 @@ fn check_treasury_pallet_id() {
         dancelight_runtime_constants::TREASURY_PALLET_ID
     );
 }
+
+#[test]
```
girazoki, line 77: I dont think this test should belong to mod.rs, specially when we dont have any test (that I know of) in that file. Either move it to its own test file or put it in integration_tests.rs

pallets/collator-assignment/src/assignment.rs
girazoki: with our knowledge probably nothing else, but adding a comment does not hurt

runtime/common/src/migrations.rs
```diff
@@ -961,33 +833,35 @@ where
...
             //Box::new(migrate_boot_nodes),
             // Applied in runtime 400
-            Box::new(migrate_config_parathread_params),
-            Box::new(migrate_add_collator_assignment_credits),
-            Box::new(migrate_registrar_pending_verification),
-            Box::new(migrate_registrar_manager),
-            Box::new(migrate_data_preservers_assignments),
-            Box::new(migrate_registrar_reserves),
-            Box::new(migrate_config_max_parachain_percentage),
+            //Box::new(migrate_config_parathread_params),
```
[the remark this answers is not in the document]
tmpolaczyk: What's the easiest way to get that info?

runtime/common/src/migrations.rs
tmpolaczyk: Not sure what to do here

pallets/collator-assignment/src/assignment.rs
tmpolaczyk: *fixed*

pallets/collator-assignment/src/lib.rs
tmpolaczyk: *fixed*

solo-chains/runtime/dancelight/src/tests/mod.rs
tmpolaczyk: *fixed*

runtime/common/src/migrations.rs
girazoki: easy: you take whatever you have before the migration and you check in the post hook that the migration has completed as you expected

Example: you output the old_config_bytes, which dont include your new rotationmethod

then in the post hook you recive that, you decode it accordingly, and then you read the value after the migration. If everything goes well,the post migration hook should contain the same values that you output in the pre but with the addition of the new rotation values

runtime/common/src/migrations.rs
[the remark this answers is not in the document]
girazoki: I would say go to each of the runtime tags and see whether the migration is there (perm-runtime-xx)

chexware (commented):
Besides the unresolved conversations all looks good to me

runtime/common/src/migrations.rs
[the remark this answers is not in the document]
tmpolaczyk: Can I just remove them instead?

runtime/common/src/migrations.rs
tmpolaczyk: Added this to the pre_upgrade, I believe it's enough:

```rust
        let old_config: Result<HostConfigurationV3, _> =
            DecodeAll::decode_all(&mut old_config_bytes.as_ref());
        let new_config: Result<HostConfiguration, _> =
            DecodeAll::decode_all(&mut old_config_bytes.as_ref());

        assert!(old_config.is_ok());
        assert!(new_config.is_err());
```

girazoki: no with this you are only checking that you cannot decode with the new config before the runtime-upgrade haha. At least add the opposite in the post upgrade

runtime/common/src/migrations.rs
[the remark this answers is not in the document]
tmpolaczyk: I went to runtime 400 and some of the migrations are missing from the list, so we are not doing a good job tracking them.

girazoki (approved):
in general I think it looks good, just a few nitpicks from open conversations!

runtime/common/src/migrations.rs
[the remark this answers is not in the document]
tmpolaczyk: Nevermind, I was confusing flashbox migrations with dancebox, the list is correct. Also, updated commented migrations with runtime version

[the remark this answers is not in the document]
tmpolaczyk: *fixed*

runtime/common/src/migrations.rs
tmpolaczyk: *fixed*

chexware (approved):

PR #771, opened 2024-12-02
Rename zombienet config files and test folders to match suite name

tmpolaczyk:
Simplify naming scheme of zombienet tests. Make suite name the only thing you need to remember, because it's already part of the command you run. Changes:

* Config file (configs/zombie_tanssi.json)
* Suite folder (suites/zombie_tanssi)
* Test filename (suites/zombie_tanssi/test_zombie_tanssi.ts)

For the rare cases where two different suites use the same config, I choose the most significant one. For example, both "zombie_tanssi" and "zombie_flashbox" use the "suites/zombie_tanssi" folder.


girazoki (approved):
I think it's a great change. Thanks @tmpolaczyk !

PR #772, opened 2024-12-02
Add support for eth to starlight message passing

ParthDesai:
# Description
This PR extends the bridge e2e shell scripts to do following: 
1. Configure execution relay (execution relay == ethereum -> starlight).
2. Start execution relay.
3. Use our own fork of snowbridge repository so that we can use updated Gateway contract.

It also extends the typescript test:
1.  Fund the address corresponding to `//ExecutionRelay` phrase so that relayer can send transaction.
2. Set correct gateway address read by the inbound queue pallet.
3. Write a test to send a tx to a dummy function on gateway contract to trigger the bridge, wait for 2 tanssi sessions and make sure that we received the message.


test/scripts/bridge/build-relayer.sh
```diff
@@ -15,7 +15,7 @@ if [ -d "$relayer_root_dir" ];
 then
   echo "Relayer seems to be already setup. Skipping git fetch"
 else
-  git clone https://github.com/Snowfork/snowbridge $relayer_root_dir
+  git clone https://github.com/moondance-labs/snowbridge $relayer_root_dir
```
tmpolaczyk, line 18: People who have already run this script need to remove the snowbridge folder, otherwise it will fail, as it will use the old remote. CI works?

test/scripts/bridge/generate-contract-info.sh
```diff
@@ -10,4 +10,6 @@ pnpm generateContracts "$output_dir/contracts.json" > /dev/null
 popd > /dev/null
 
 # Output the file so that invoker can read it
-cat "$output_dir/contracts.json"
+contract_output=$(cat "$output_dir/contracts.json")
+
+echo "{ \"data\": $contract_output, \"ethereum_key\": \"$ethereum_key\" }"
```
tmpolaczyk, line 15: Let's use jq? (untested)

```suggestion
jq --arg ethereum_key "$ethereum_key" '{data: ., ethereum_key: $ethereum_key}' "$output_dir/contracts.json"
```

tmpolaczyk (approved):

test/scripts/bridge/build-relayer.sh
ParthDesai: *disregard*

test/scripts/bridge/generate-contract-info.sh
ParthDesai: *disregard*

PR #773, opened 2024-12-02
Revisit xcm config for dancelight

dimartiro:
# Description

This PR addresses the following issues with the dancelight XCM configuration:

- Disables teleported asset transfers
- Enables reserve asset transfers
- Updates waived locations
- Additionally, it includes the required configuration for the mocknets to test these changes.


solo-chains/runtime/dancelight/src/xcm_config.rs
```diff
@@ -136,20 +154,10 @@ parameter_types! {
     pub StarForBridgeHub: (AssetFilter, Location) = (Star::get(), BridgeHub::get());
     pub StarForPeople: (AssetFilter, Location) = (Star::get(), People::get());
     pub StarForBroker: (AssetFilter, Location) = (Star::get(), Broker::get());
+    pub const RelayNetwork: NetworkId = NetworkId::Westend;
```
girazoki, line 157: I would change this, but it is true that we need to think what to put..

girazoki: for now you can leave it as is though

girazoki (approved):
looks good to me!

solo-chains/runtime/dancelight/src/tests/common/xcm/expected_event_checker.rs
```diff
@@ -0,0 +1,103 @@
...
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+#[macro_export]
+macro_rules! assert_expected_events {
```
tmpolaczyk, line 18: Why is this here? Can't we use the one from polkadot-sdk?

girazoki: I am sure there was a good reason for bringing it. But we can try using the polkadot one

primitives/xcm-commons/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn chain_part(&self) -> Option<Location>;
+}
+
+impl Parse for Location {
+    fn chain_part(&self) -> Option<Location> {
+        match (self.parents, self.first_interior()) {
+            // sibling parachain
+            (1, Some(Parachain(id))) => Some(Location::new(1, [Parachain(*id)])),
+            // parent
+            (1, _) => Some(Location::parent()),
```
Agusrodri, line 32: We should re-check these variants for the case of dancelight. Given that it's a relay context, not sure if we want to return something for "parent" or "sibling parachains".

Should we create a new trait/struct for dancelight? What do you think @girazoki @dimartiro?

solo-chains/runtime/dancelight/src/xcm_config.rs
```diff
@@ -116,6 +117,23 @@ pub type XcmRouter = WithUniqueTopic<
     ChildParachainRouter<Runtime, XcmPallet, PriceForChildParachainDelivery>,
 >;
 
+/// Type for specifying how a `Location` can be converted into an `AccountId`. This is used
+/// when determining ownership of accounts for asset transacting and when attempting to use XCM
+/// `Transact` in order to determine the dispatch Origin.
+pub type LocationToAccountId = (
```
Agusrodri, line 123: I guess we should remove this. For dancelight, we already have the proper `LocationConverter` configured.

See: https://github.com/moondance-labs/tanssi/blob/master/solo-chains/runtime/dancelight/src/xcm_config.rs#L64

primitives/xcm-commons/src/lib.rs
girazoki: I think the goal is to re-use them for both sibling parachain and relay mode. So I think it's ok to keep them

Agusrodri (approved):

PR #774, opened 2024-12-03
Add pallet_pooled_staking to Dancelight

tmpolaczyk:
Add pallet_pooled_staking to Dancelight. Container chain collators will be selected first from invulnerables, and then from staking.


nanocryk (approved):

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1634,10 +1637,97 @@ impl pallet_inflation_rewards::Config for Runtime {
...
+    pub StakingAccount: AccountId32 = PalletId(*b"POOLSTAK").into_account_truncating();
+    pub const InitialManualClaimShareValue: u128 = MILLIUNITS;
+    pub const InitialAutoCompoundingShareValue: u128 = MILLIUNITS;
+    pub const MinimumSelfDelegation: u128 = 10_000 * UNITS;
+    pub const RewardsCollatorCommission: Perbill = Perbill::from_percent(20);
+    // Need to wait 2 sessions before being able to join or leave staking pools
+    pub const StakingSessionDelay: u32 = 2;
+}
+
+pub struct SessionTimer<G>(PhantomData<G>);
```
ParthDesai, line 1654: Can you put something more descriptive than `G`? Like `SessionDelay` or `Delay`?

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -3007,8 +3099,27 @@ impl tanssi_initializer::ApplyNewSession<Runtime> for OwnApplySession {
...
+        let target_session_index = session_index.saturating_add(1);
+        let max_collators = <CollatorConfiguration as GetHostConfiguration<u32>>::max_collators(
+            target_session_index,
+        );
+        let next_collators: Vec<_> = invulnerables
+            .iter()
+            .cloned()
+            .chain(candidates_staking.into_iter().filter_map(|elig| {
+                let cand = elig.candidate;
+                if invulnerables.contains(&cand) {
```
ParthDesai, line 3123: Just curious, what is the possibility of this happening?

Agusrodri:
Changes look good to me! Could we add at least one integration test to check the overall functioning of the pooled staking inside dancelight?

girazoki:
And a description please @tmpolaczyk

solo-chains/runtime/dancelight/src/lib.rs
tmpolaczyk: It's as easy as looking at the list of staking candidates and adding one of them as an invulnerable. Obviously shouldn't happen intentionally, but that can be prevented by this filter_map.

ParthDesai: I think it would be better design if we prevent that from happening. What do you think? 

Let's leave that right now, if it is out of scope for this PR.

tmpolaczyk: We would need to prevent both marking as invulnerable an identity from staking, as well as prevent adding to staking an identity that's already an invulnerable. Would be more error prone than this.

solo-chains/runtime/dancelight/src/lib.rs
tmpolaczyk: *fixed*

solo-chains/runtime/dancelight/src/lib.rs
tmpolaczyk: *disregard*

tmpolaczyk:
I separated the typescript tests to a new PR because they are not trivial, and not ready yet. https://github.com/moondance-labs/tanssi/pull/782

Agusrodri (approved):

PR #775, opened 2024-12-04
Add iforgor config and readme

nanocryk:
Add config file and README for tool `iforgor`, which will allow us to share scripts for commands commonly run when working on Tanssi.


tools/iforgor/tanssi.toml
```diff
@@ -0,0 +1,189 @@
...
+script = "git fetch origin && git pull"
+
+[[entries]]
+name = "[Git] Push"
+script = "git push"
+
+[[entries]]
+name = "[Git] Delete local branches (select)"
+script = """
+git branch | ichoose --multi --title "Select branches to delete" | while read x ; do git branch -D $x ; done
```
ParthDesai, line 188: `ichoose` does not exists on macos. Is it built-in for linux?

tools/iforgor/tanssi.toml
```diff
@@ -0,0 +1,189 @@
...
+echo "\nℹ️ Running 'cargo fmt'\n"
+cargo fmt
+
+echo "\nℹ️ Running 'toml-maid'\n"
+toml-maid
+
+echo "\nℹ️ Running 'zepter'\n"
+zepter
+
+cd test
```
ParthDesai, line 159: You might want to use `pushd` and `popd` here. So, that user is back in the main directory once the command is run instead in the test directory.

tools/iforgor/tanssi.toml
```diff
@@ -0,0 +1,189 @@
...
+
+# Typescript
+
+## API
+
+[[entries]]
+name = "[Tanssi] Generate TS-API"
+only_in_dir = "**/tanssi"
+script = """
+cd typescript-api
```
ParthDesai, line 114: Same here. Use `pushd` and `popd` so user is returned to the same directory.

tools/iforgor/tanssi.toml
nanocryk: It is another binary similar to iforgor. I talk about it in the README.

tools/iforgor/tanssi.toml
nanocryk: Directory change doesn't persist after script has finished running. Script is run in a new bash process and doesn't affect the current directory of the user.

tools/iforgor/tanssi.toml
ParthDesai: *disregard*

tools/iforgor/tanssi.toml
ParthDesai: Okay. So, both needs to be installed on both platform, correct?

girazoki (approved):

PR #779, opened 2024-12-09
Update moonwall to 5.9.1 and add --bail=1

tmpolaczyk:
The --bail=1 flag makes the tests stop on the first failure, as opposed to always executing all the tests. The zombie_data_preservers tests takes 40 minutes to fail because it waits for several tests to fail with timeout, with this it should fail in 10 minutes.

https://stackoverflow.com/a/78969381

**Requires node 22**


ParthDesai:
@tmpolaczyk What does `--bail=1` do?

tmpolaczyk:
> @tmpolaczyk What does `--bail=1` do?

Updated PR description

ParthDesai (approved):
Much needed. Thanks.

girazoki (approved):
thank you @tmpolaczyk

PR #780, opened 2024-12-10
Bridge test to work with sha

girazoki:
UPdates download job so and adds more debuggeability to e2e bridge tests


.github/workflows/e2e-test-bridge.yml
```diff
@@ -93,13 +93,18 @@ jobs:
...
-              uses: dawidd6/action-download-artifact@v2
+              uses: dawidd6/action-download-artifact@v7
               with:
-                run_id: ${{ steps.retrieve-run-id.outputs.run_id }}
+                commit: ${{ steps.retrieve-run-id.outputs.run_sha }}
+                workflow: release.yml
                 name: binaries
                 path: target/release
-                search_artifacts: true
+                allow_forks: true
```
ParthDesai, line 107: Why do you need to have `allow_forks: true`?

girazoki: I read the guy recommends it

girazoki: besides we were running with allow_forks already, it's just in the new version the default is false

ParthDesai: If we do not need it, let's set it to false as making it true explicitly allows this attack: https://github.com/dawidd6/action-download-artifact/security/advisories/GHSA-5xr6-xhww-33m4

Which is the reason why it is by default false.

girazoki: fixed

NindoK (approved):

PR #782, opened 2024-12-11
Add pallet staking tests for starlight (typescript)

tmpolaczyk:
Add typescript tests for pallet_pooled_staking in Dancelight. They need some changes from the Dancebox tests because now we must create container chain blocks, as the relay block author is not rewarded through staking.


girazoki (approved):
good job @tmpolaczyk

PR #783, opened 2024-12-11
Fix: benchmarking for pallet-data-preservers and pallet-registrar on  Dancelight

chexware:
This PR adds benchmarking weights for `pallet-data-preservers`  and `pallet-registrar` to Dancelight.

Changes:
- Implemented additional benchmarking helper function for `RegistrarHandler` which reserves  `ParaId
` on Dancelight.
-  Updated the `ParaId`  in failing benchmarking cases - Any value lower than 2000 will be overwritten by the current `LOWEST_PUBLIC_ID` defined in Polkadot SDK and the benchmarks will still fail after reserving  a `ParaId `


solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -2045,6 +2045,9 @@ impl pallet_registrar::Config for Runtime {
     type DepositAmount = DepositAmount;
     type RegistrarHooks = DancelightRegistrarHooks;
     type RuntimeHoldReason = RuntimeHoldReason;
+    #[cfg(feature = "runtime-benchmarks")]
+    type InnerRegistrar = ();
+    #[cfg(not(feature = "runtime-benchmarks"))]
```
tmpolaczyk, line 2050: I believe the weight of `T::InnerRegistrar::register` is not accounted for if we do it like this, so we need to fix the benchmark in some other way

girazoki: yes indeed

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -2045,6 +2045,9 @@ impl pallet_registrar::Config for Runtime {
...
     type InnerRegistrar =
         InnerDancelightRegistrar<Runtime, AccountId, Registrar, paras_registrar::TestWeightInfo>;
```
girazoki, line 2052: Now that we are touching this, we should probably change `paras_registrar::TestWeightInfo` for weights::runtime_common_paras_registrar::SubstrateWeight<Runtime>;

solo-chains/runtime/dancelight/src/lib.rs
girazoki: however, why would we need to change this? why does pallet-registrar influence whatsoever on the benchmarks of data preservers? I get that you need a chain registered for this to work, but we can probably set it up as a benchmark helper in pallet-data-preservers

tmpolaczyk: Either we benchmark the pallet including the hooks, or we return the weight from `T::InnerRegistrar::register` and benchmark that separately.

I find it simpler to benchmark the pallet including the hooks, this way we don't need to do manual weight tracking, and also the benchmark is more similar to production so more accurate in my opinion.

tmpolaczyk: What do you mean? Dancebox always has `InnerRegistrar = ()`. But if we touch the benchmark it can break so we should make sure it still works after any changes we end up doing.

chexware: Cool thanks for clarifying that !

solo-chains/runtime/dancelight/src/lib.rs
chexware: Updated in https://github.com/moondance-labs/tanssi/pull/783/commits/f0c48c3e3ddb550cac505da66391875803d9cfde

solo-chains/runtime/dancelight/src/lib.rs
chexware: Resolved in https://github.com/moondance-labs/tanssi/pull/783/commits/955f90f481d86d5cc9c6447178e443ff4c2d94b5, [075b3f7](https://github.com/moondance-labs/tanssi/pull/783/commits/075b3f7470afed009295a3d8e5b99da334f15e9d), and https://github.com/moondance-labs/tanssi/pull/783/commits/a010dc878707db1a4670266ce9823cb6783ca76b

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1618,11 +1620,54 @@ impl pallet_data_preservers::AssignmentPayment<AccountId> for PreserversAssignme
...
+        let genesis_data = ContainerChainGenesisData {
+            storage,
+            name: Default::default(),
+            id: Default::default(),
+            fork_id: Default::default(),
+            extensions: Default::default(),
+            properties: Default::default(),
+        };
+
+        assert_ok!(pallet_registrar::Pallet::<T>::register(
```
girazoki, line 1657: do we need to both reserve and register here?

solo-chains/runtime/dancelight/src/lib.rs
```diff
@@ -1618,11 +1620,54 @@ impl pallet_data_preservers::AssignmentPayment<AccountId> for PreserversAssignme
...
+#[cfg(feature = "runtime-benchmarks")]
+impl<T> ArgumentFactory<T::AccountId> for DataPreserversBenchmarkHelper<T>
+where
+    T: pallet_registrar::Config + pallet_balances::Config + paras_registrar::Config,
+{
+    fn reserve_para_id(para_id: ParaId) {
+        use frame_support::{assert_ok, dispatch::RawOrigin, traits::fungible::Mutate};
+
+        let account: T::AccountId = frame_benchmarking::account("para_manager", 0, 1);
+        assert_ok!(<T as pallet_registrar::Config>::Currency::mint_into(
```
girazoki, line 1634: Here you can probably call `force_register` instead of register. This should avoid:
- sending tokens to the register account
- need to reserve the para-id (maybe I am wrong here).
- all the nextFreeParaId stuff.
-

chexware: I tried using `paras_registrar::force_register` however  I either get 'NotOwner' error if I don't use `pallet_registrar::register`  after or `AlreadyRegistered `error if I do call it

solo-chains/runtime/dancelight/src/lib.rs
chexware: Moved reserving ParaId to `InnerDancelightRegister::prepare_chain_registration ` to avoid duplicating code in https://github.com/moondance-labs/tanssi/pull/783/commits/674169290cfa373cac5047de830553dc0e85d3d3

solo-chains/runtime/dancelight/src/lib.rs
chexware: Moved reserving ParaId to `InnerDancelightRegister::prepare_chain_registration ` to avoid duplicating code in https://github.com/moondance-labs/tanssi/pull/783/commits/674169290cfa373cac5047de830553dc0e85d3d3

pallets/data-preservers/src/benchmarks.rs
```diff
@@ -36,6 +36,17 @@ use {
     tp_traits::{ParaId, StorageDeposit},
 };
 
+/// Trait describing factory function for para_id.
+pub trait ArgumentFactory<AcocuntId> {
+    /// Factory function reserving parachain
+    fn reserve_para_id(para_id: ParaId);
```
tmpolaczyk, line 42: Is this unused?

chexware: Yes removed in https://github.com/moondance-labs/tanssi/pull/783/commits/73b28788d8abcc176e94c59120d83842b0086065

tmpolaczyk (approved):
Thanks!

PR #784, opened 2024-12-16
Reward block author in Dancelight

tmpolaczyk:
Previously we were only rewarding the Dancelight block author for including parachain candidates, but not for the block per se. So an empty block would get no reward.

This PR fixed that by rewarding 20 points to the block author, same as in polkadot.

https://github.com/paritytech/polkadot-sdk/blob/48c6574b1a32893e6a2113622d009cefda0a5f21/substrate/frame/staking/src/pallet/impls.rs#L1411


test/suites/dev-tanssi-relay/external-validators-rewards/test_external_validator_reward_candidates.ts
```diff
@@ -0,0 +1,67 @@
...
+                await jumpToSession(context, 3);
+                await context.createBlock();
+
+                // we are still in era 0
+                const validatorRewards = await context
+                    .polkadotJs()
+                    .query.externalValidatorsRewards.rewardPointsForEra(0);
+                const totalRewards = validatorRewards.total.toBigInt();
+
+                expect(totalRewards).to.be.greaterThan(0n);
```
girazoki, line 37: we should make sure here that rewards are higher than what the validator gets for producing a block. Otherwise we are not testing specifically the rewards given by candidates.

For that you probably need to fetch the previous blocks reward accumulation points, the current one, and the difference should be higher than 20

tmpolaczyk: *fixed*

test/suites/dev-tanssi-relay/external-validators-rewards/test_external_validator_reward_candidates.ts
```diff
@@ -0,0 +1,68 @@
...
+                await context.createBlock();
+
+                // we are still in era 0
+                const validatorRewards = await context
+                    .polkadotJs()
+                    .query.externalValidatorsRewards.rewardPointsForEra(0);
+                const totalRewards = validatorRewards.total.toBigInt();
+
+                // Validators get 20 points for creating a block, so if they included a candidate, they will get more than 20
+                expect(totalRewards).to.be.greaterThan(20n);
```
girazoki, line 38: wait they get 20 per candidate. So you will always have higher than 20?

girazoki: I guess if so, you need to check the rewards `totalRewards > num_blocks*20`

tmpolaczyk: *fixed*

Agusrodri (approved):

PR #785, opened 2024-12-17
Test operator from Ethereum bridge produces blocks

nanocryk:
Adds a node in the moonwall config that'll produce blocks. It is not automatically added to invulnerables/whitelisted validators (that's good). We rotate its nimbus key and register it so it'll be eligible to produce blocks. After the operators are bridged, we wait a bit before checking it is the author of a block.


tmpolaczyk (approved):

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -44,17 +45,19 @@ describeSuite({
...
             const beaconRelay = keyring.addFromUri("//BeaconRelay", { name: "Beacon relay default" });
             const executionRelay = keyring.addFromUri("//ExecutionRelay", { name: "Execution relay default" });
 
             // Operator keys
-            operatorAccount = keyring.addFromUri("//" + "Bob", { name: "COLLATOR" + " ACCOUNT" });
-            operatorNimbusKey = keyring.addFromUri("//" + "COLLATOR_NIMBUS", { name: "COLLATOR" + " NIMBUS" });
-
-            await relayApi.tx.session.setKeys(u8aToHex(operatorNimbusKey), []).signAndSend(operatorAccount);
+            operatorAccount = keyring.addFromUri("//Charlie", { name: "Charlie default" });
+            operatorNimbusKey = await relayCharlieApi.rpc.author.rotateKeys();
```
girazoki, line 58:
```suggestion
            // We rotate the keys for charlie so that we have access to them from this test as well as the node
            operatorNimbusKey = await relayCharlieApi.rpc.author.rotateKeys();
```

girazoki (approved):
small coment

PR #787, opened 2024-12-18
Refactor on_container_author_noted hook to take a Vec

tmpolaczyk:
To get a more accurate weight, it's better to have a function that takes a list of n chains to reward than to call a function n times.

Should be a refactor but there are some observable changes:

* Block 0 of each container chain is no longer being rewarded. It's the genesis block and it's not created by a collator, so the collator shouldn't get any reward
* The event order changes. In pseudocode:

```rust
// Previous order:
for para_id in paras { for pallet in pallets { pallet.emit_event(para_id) }}

// New order:
for pallet in pallets { for para_id in paras { pallet.emit_event(para_id) }}
```


girazoki (approved):
I think it looks good. Just double checking:

- we dont need migrations
- changing the order (iterate over paras vs iterate over pallets first) does not affect

chexware (approved):
Looks good to me!

Agusrodri (approved):

PR #792, opened 2024-12-23
Add xcm upm mock endpoint and tests

dimartiro:
# Description

Add support for xcm upward passing messages in dancelight

- [x] Add mock endpoint to send upward passing messages
- [x] Add typescript test to check all is working ok
    - [x] Create a bunch of utils method

### DIFF
--- a/solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
+++ b/solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
@@ -44,6 +44,7 @@ pub trait DevApi {
     async fn inject_upward_message(&self, message: Vec<u8>) -> RpcResult<()>;
 }
 
+#[derive(Clone)]
 pub struct DevRpc {
     pub mock_para_inherent_channel: flume::Sender<Vec<u8>>,
     pub upward_message_channel: flume::Sender<Vec<u8>>,
--- a/solo-chains/node/tanssi-relay-service/src/dev_service.rs
+++ b/solo-chains/node/tanssi-relay-service/src/dev_service.rs
@@ -108,11 +108,8 @@ struct DevDeps<C, P> {
     pub pool: Arc<P>,
     /// Manual seal command sink
     pub command_sink: Option<futures::channel::mpsc::Sender<EngineCommand<Hash>>>,
-    /// Channels for dev rpcs
-    pub dev_rpc_data: (
-        Option<flume::Sender<Vec<u8>>>, //downward
-        Option<flume::Sender<Vec<u8>>>, //upward
-    ),
+    /// Dev rpcs
+    pub dev_rpc: Option<DevRpc>,
 }
 
 fn create_dev_rpc_extension<C, P>(
@@ -120,7 +117,7 @@ fn create_dev_rpc_extension<C, P>(
         client,
         pool,
         command_sink: maybe_command_sink,
-        dev_rpc_data: maybe_dev_rpc_data,
+        dev_rpc: maybe_dev_rpc,
     }: DevDeps<C, P>,
 ) -> Result<RpcExtension, Box<dyn std::error::Error + Send + Sync>>
 where
@@ -149,14 +146,8 @@ where
         io.merge(ManualSeal::new(command_sink).into_rpc())?;
     }
 
-    if let Some(mock_para_inherent_channel) = maybe_dev_rpc_data.0 {
-        io.merge(
-            DevRpc {
-                mock_para_inherent_channel,
-                upward_message_channel: maybe_dev_rpc_data.1.unwrap(),
-            }
-            .into_rpc(),
-        )?;
+    if let Some(dev_rpc_data) = maybe_dev_rpc {
+        io.merge(dev_rpc_data.into_rpc())?;
     }
 
     Ok(io)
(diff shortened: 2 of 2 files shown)


solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -388,6 +397,16 @@ where
                                 &validation_code_hash,
                             );
                             let collator_signature = collator_pair.sign(&payload);
+
+                            let mut upm_messages = UpwardMessages::new();
```
Agusrodri, line 401:
```suggestion
                            let mut ump_messages = UpwardMessages::new();
```

solo-chains/runtime/dancelight/src/tests/common/xcm/expected_event_checker.rs
Agusrodri, line 1: Hadn't we removed this and started using the polkadot-sdk one? Maybe it was a mistake while rebasing the branch.

solo-chains/runtime/dancelight/src/xcm_config.rs
```diff
@@ -117,6 +117,23 @@ pub type XcmRouter = WithUniqueTopic<
     ChildParachainRouter<Runtime, XcmPallet, PriceForChildParachainDelivery>,
 >;
 
+/// Type for specifying how a `Location` can be converted into an `AccountId`. This is used
+/// when determining ownership of accounts for asset transacting and when attempting to use XCM
+/// `Transact` in order to determine the dispatch Origin.
+pub type LocationToAccountId = (
```
Agusrodri, line 123: Same with this, it should be removed.

test/moonwall.config.json
```diff
@@ -50,7 +50,7 @@
         {
             "name": "dev_tanssi_relay",
             "envVars": ["DEBUG_COLORS=1"],
-            "testFileDir": ["suites/dev-tanssi-relay", "suites/common-all", "suites/common-tanssi"],
```
Agusrodri, line 53: Let's undo this change.

test/moonwall.config.json
```diff
@@ -75,7 +75,8 @@
                             "--chain=tmp/dancelight-dev-raw.json",
                             "--no-hardware-benchmarks",
                             "--wasmtime-precompiled=wasm",
-                            "--unsafe-force-node-key-generation"
+                            "--unsafe-force-node-key-generation",
+                            "-lxcm=trace"
```
Agusrodri, line 79: And this one as well, keep it as before.

solo-chains/node/tanssi-relay-service/src/dev_rpcs.rs
```diff
@@ -69,6 +74,30 @@ impl DevApiServer for DevRpc {
...
         Ok(())
     }
+
+    async fn inject_upward_message(&self, msg: Vec<u8>) -> RpcResult<()> {
+        let upward_message_channel = self.upward_message_channel.clone();
+        // If no message is supplied, inject a default one.
+        let msg = if msg.is_empty() {
+            // TODO: recheck this
+            xcm::VersionedXcm::<()>::V4(Xcm(vec![
+                ReserveAssetDeposited((Parent, 10000000000000u128).into()),
```
Agusrodri, line 84: Let's not forget to build the proper message here :)

solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -719,7 +742,15 @@ fn new_full<
                                 &[],
                             )
                             .expect("Should be able to write to aux storage; qed");
+                        }
 
+                        if let Some(value) = upward_message.last() {
```
girazoki, line 747: I would take a different approach here, I would not sore it in an auxiliary message (as this is not a temporary thing we need to read). Instead, I would read all (and not only the last) message we send from this channel and inject it into the create function for `MockParachainsInherentDataProvider` as an argument

solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -108,7 +109,10 @@ struct DevDeps<C, P> {
     /// Manual seal command sink
     pub command_sink: Option<futures::channel::mpsc::Sender<EngineCommand<Hash>>>,
     /// Channels for dev rpcs
-    pub dev_rpc_data: Option<flume::Sender<Vec<u8>>>,
+    pub dev_rpc_data: (
+        Option<flume::Sender<Vec<u8>>>, //downward
+        Option<flume::Sender<Vec<u8>>>, //upward
+    ),
```
tmpolaczyk, line 115: Give proper names to the channels, or use this struct that already has names:

```suggestion
    pub dev_rpc_data: Option<DevRpc>,
```

Also I don't understand why use an Option here, can we remove it?

solo-chains/node/tanssi-relay-service/src/dev_service.rs
```diff
@@ -388,6 +402,12 @@ where
                                 &validation_code_hash,
                             );
                             let collator_signature = collator_pair.sign(&payload);
+
+                            let upward_messages = UpwardMessages::try_from(
+                                upward_messages_receiver.drain().collect::<Vec<_>>().clone(),
```
tmpolaczyk, line 407:
```suggestion
                                upward_messages_receiver.drain().collect::<Vec<_>>(),
```

solo-chains/node/tanssi-relay-service/src/dev_service.rs
tmpolaczyk: We use a mock storage key when we want to store state, such as enable or disable some feature. Here we only want to send a single message for every time the rpc is called, right? Then we don't need storage, I believe passing the channel is ok.

tmpolaczyk (approved):

solo-chains/node/tanssi-relay-service/src/dev_service.rs
girazoki: yes we agree!

girazoki (approved):

girazoki:
At some point in the future we should inject messages that come from a specific para-id. Right now the only collatable para-id is 2000, so it's ok if we do it like this, but some food for thought in the future

PR #793, opened 2024-12-31
Starlight <> Ethereum: Slashes mapping

Agusrodri:
Introduces slashes to be sent to Ethereum from pallet-external-validator-slashes.

Up to now slashes were just recorded but nothing was done with them. This PR changes this by:

- scheduling those slashes to be sent externally at a certain era change.
- Send this slashes in batches of `QueuedSlashesProcessedPerBlock`. This batches are sent on-initialize of each block, with one batch at most per block
- Introduce the `Slashes` command which contains: The era in which the slashes are sent, the slashes, and a timestamp. Currently the timestamp is likely wrong, in the future it will contain a timestamp to identify the validators in the external chain.
- Tests validating the approach, both rust and typescript

An event called `SlashesMessageSent` was added to allow checking the consistency of the command sent, through unit and integration tests.


pallets/external-validator-slashes/src/lib.rs
```diff
@@ -485,6 +491,84 @@ impl<T: Config> OnEraStart for Pallet<T> {
...
+
+        let era_slashes = Slashes::<T>::get(era_index);
+        let unreported_slashes = UnreportedSlashes::<T>::get();
+
+        let free_slashing_space = SLASH_PAGE_SIZE.saturating_sub(unreported_slashes.len());
+
+        let mut slashes_to_send: Vec<_> = vec![];
+
+        for unreported_slash in unreported_slashes.iter() {
+            // TODO: check if validator.clone().encode() matches with the actual account bytes.
```
tmpolaczyk, line 508: Maybe not, use `validator.as_ref().to_vec()`

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -485,6 +491,84 @@ impl<T: Config> OnEraStart for Pallet<T> {
     }
 }
 
+impl<T: Config> tp_traits::OnEraEnd for Pallet<T> {
+    fn on_era_end(era_index: EraIndex) {
+        // Send up to SLASH_PAGE_SIZE slashes per era
```
tmpolaczyk, line 496: So the slashes get delayed for at least 1 era? I don't like this, why can't we send them in the next block?

girazoki: no I think we need to send them on-initialize if we were not able to send them all

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -481,24 +511,66 @@ impl<T: Config> OnEraStart for Pallet<T> {
...
+        UnreportedSlashesQueue::<T>::mutate(|queue| queue.append(&mut slashes));
+    }
+
+    fn process_slashes_queue_page() {
+        let mut slashes_to_send: Vec<_> = vec![];
+        let era_index = T::EraIndexProvider::active_era().index;
+
+        // prepare up to QueuedSlashesProcessedPerBlock slashes to be sent
+        for _ in 0..(T::QueuedSlashesProcessedPerBlock::get() as usize) {
+            let Some(slash) = UnreportedSlashesQueue::<T>::mutate(VecDeque::pop_front) else {
```
tmpolaczyk, line 531: (optimization): mutating storage inside a for loop sounds bad, let's try to read storage once and use drain to remove n elements @nanocryk:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a1269f106b287540fbb7fe2634f35b35

nanocryk: Issue with drain is that it panics if the range goes out of bound. What we can do is place the mutate around the for loop so that it we perform multiple pop_front in memory and write the result once the limit is reached or pop_front fails.

tmpolaczyk: Ok didn't know it panics, the alternative sounds good

Agusrodri:
Changes look good! It's an approval for me.

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -330,6 +346,16 @@ pub mod pallet {
             Ok(())
         }
     }
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+            log::info!("inside on-initialize");
```
tmpolaczyk, line 353:
```suggestion
```

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -481,24 +511,75 @@ impl<T: Config> OnEraStart for Pallet<T> {
...
-            );
-            for slash in era_slashes {
-                slash.confirmed = true;
+    fn add_era_slashes_to_queue(active_era: EraIndex) {
+        let mut slashes: VecDeque<_> = Slashes::<T>::get(&active_era).into();
+
+        UnreportedSlashesQueue::<T>::mutate(|queue| queue.append(&mut slashes));
+    }
+
+    fn process_slashes_queue(amount: u32) -> u32 {
```
tmpolaczyk, line 525:
```suggestion
    /// Process up to QueuedSlashesProcessedPerBlock slashes.
    /// Returns number of slashes that were sent to ethereum.
    fn process_slashes_queue(amount: u32) -> u32 {
```

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -78,6 +78,8 @@ pub mod pallet {
             fraction: Perbill,
             slash_era: EraIndex,
         },
+        /// The slashes message was sent correctly.
+        SlashesMessageSent { slashes_command: Command },
```
tmpolaczyk, line 82: This Command includes all the slashes, right? So this event will be quite big. I guess we don't care?

pallets/external-validator-slashes/src/mock.rs
```diff
@@ -197,11 +207,18 @@ impl DeferPeriodGetter {
...
+    SENT_ETHEREUM_MESSAGE_NONCE.with(|q| (*q.borrow()))
+}
+
 pub struct MockOkOutboundQueue;
 impl tp_bridge::DeliverMessage for MockOkOutboundQueue {
     type Ticket = ();
 
     fn deliver(_: Self::Ticket) -> Result<H256, SendError> {
+        // Every time we hit deliver, increment the nonce
+        SENT_ETHEREUM_MESSAGE_NONCE.with(|r| *r.borrow_mut() = r.take() + 1);
```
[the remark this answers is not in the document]
Agusrodri: That doesn't compile :(

```shell
error[E0368]: binary assignment operation `+=` cannot be applied to type `RefMut<'_, u64>`
    --> pallets/external-validator-slashes/src/mock.rs:220:46
     |
220  |         SENT_ETHEREUM_MESSAGE_NONCE.with(|r| r.borrow_mut() += 1);
     |                                              --------------^^^^^
     |                                              |
     |                                              cannot use `+=` on type `RefMut<'_, u64>`
     |
note: the foreign item type `RefMut<'_, u64>` doesn't implement `AddAssign<{integer}>`
    --> /root/.rustup/toolchains/1.81.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:1836:1
     |
1836 | pub struct RefMut<'b, T: ?Sized + 'b> {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not implement `AddAssign<{integer}>`
help: `+=` can be used on `u64` if you dereference the left-hand side
     |
220  |         SENT_ETHEREUM_MESSAGE_NONCE.with(|r| *r.borrow_mut() += 1);
     |                                              +

For more information about this error, try `rustc --explain E0368`.
error: could not compile `pallet-external-validator-slashes` (lib test) due to 1 previous error
```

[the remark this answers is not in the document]
Agusrodri: In my opinion I'd use the original approach, or do you have another optimization in mind?

pallets/external-validator-slashes/src/lib.rs
Agusrodri: The amount of slashes should not be very big as they should not happen very often, but even if we have a lot of them, I think we are fine anyways as this event is mainly to track the information sent inside tests.

pallets/external-validator-slashes/src/lib.rs
Agusrodri: _fixed_

pallets/external-validator-slashes/src/lib.rs
Agusrodri: _fixed_

pallets/external-validator-slashes/src/lib.rs
girazoki: Also, this command only contains up `QueuedSlashesProcessedPerBlock`. This is the limit we set to the vector that goes into the command. There are plenty of tests verifying this

pallets/external-validator-slashes/src/mock.rs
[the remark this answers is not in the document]
Agusrodri: _fixed_

tmpolaczyk (approved):

chains/orchestrator-relays/runtime/dancelight/src/tests/slashes.rs
chexware, line 1: I see we have `invulnerables_cannot_be_slashed` test for BABE but not one for GRANDPA. Is there any reason for that?

girazoki: possibly they hit the same line of code? I can add one

chexware: I was  just curious as we have 2 separate tests for `non_invulnerables_can_be_slashed_with_grandpa`( which uses `inject_grandpa_slash` helper function) and  `on_invulnerables_can_be_slashed_with_babe` (which uses `inject_babe_slash` helper function). However, `invulnerables_cannot_be_slashed` injects a BABE slash only

pallets/external-validator-slashes/src/lib.rs
```diff
@@ -481,24 +509,77 @@ impl<T: Config> OnEraStart for Pallet<T> {
...
-            );
-            for slash in era_slashes {
-                slash.confirmed = true;
+    fn add_era_slashes_to_queue(active_era: EraIndex) {
+        let mut slashes: VecDeque<_> = Slashes::<T>::get(&active_era).into();
+
+        UnreportedSlashesQueue::<T>::mutate(|queue| queue.append(&mut slashes));
+    }
+
+    /// Process up to QueuedSlashesProcessedPerBlock slashes.
```
nanocryk, line 523:
```suggestion
```
Since the function now takes the amount to process we should remove this line.

nanocryk (approved):
Looks good to me, just small nitpick in documentation.

chains/orchestrator-relays/runtime/dancelight/src/tests/slashes.rs
Agusrodri: Just added a `invulnerables_cannot_be_slashed_with_grandpa` test in cce0e2ece53466d5c9c308817cefc57711d8be3f.

pallets/external-validator-slashes/src/lib.rs
Agusrodri: _fixed_

chexware (approved):

PR #796, opened 2025-01-06
update polkadot api and other dependencies in typescript-api

Rihyx:
- Updated PolkadotJS to v15 
- Updated other dependencies


tmpolaczyk (approved):
The failing CI job (twiggy) is because of an integration that needs the PR author to have write access to the repo

girazoki:
I think we can merge without it

PR #801, opened 2025-01-09
Benchmark slashes queue processing

nanocryk:
Change the function signature of `process_slashes_queue` to take the limit as argument, which helps benchmarking it. It also now returns the number of slashes processed to return the actually used weight.

Following suggestion in #793, I only read and write from storage once by moving the `for` loop inside the `mutate` call.


girazoki (approved):
good job jeremy

PR #802, opened 2025-01-10
Changes in dancelight to enable solochain message relay

ParthDesai:
# Description
This PR contains small changes to enable solochain message relay. 

Changes are following:

1. Beefy `LeafExtra` is extended to include `CommitmentRoot`. Which is a message commitment root produced by outbound queue pallet.
2. New pallet called `outbound-message-commitment-recorder` is added which stores a commitment root and also allow other code to read it.
3. Adding `OnNewCommitment` hook implementation that takes new hook and stores it in the `outbound-message-commitment-recorder` pallet.
4. Putting bridge related pallets before pallet session in `construct_runtime`, as otherwise in case the reward message is sent on era end is lost because the outbound queue pallet kills message storage on initialization.
5. Changing outbound command enum's starting index from 32 to 12, since solidity does not like it when enum is out of bound.
6. Adding outbound queue runtime api in dancelight.


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1823,6 +1840,12 @@ construct_runtime! {
         ExternalValidatorSlashes: pallet_external_validator_slashes = 21,
         ExternalValidatorsRewards: pallet_external_validators_rewards = 22,
 
+        // Bridging stuff - 1
+        EthereumOutboundQueue: snowbridge_pallet_outbound_queue = 23,
+        EthereumInboundQueue: snowbridge_pallet_inbound_queue = 24,
+        EthereumSystem: snowbridge_pallet_system = 25,
+        OutboundMessageCommitmentRecorder: pallet_outbound_message_commitment_recorder = 26,
```
tmpolaczyk, line 1847: Can you add an integration test for this? Something that will break if someone changes the pallet order back to what it was before

primitives/bridge/src/lib.rs
```diff
@@ -86,8 +86,8 @@ impl Command {
     pub fn index(&self) -> u8 {
         match self {
             // Starting from 32 to keep compatibility with Snowbridge Command enum
-            Command::Test { .. } => 32,
-            Command::ReportRewards { .. } => 33,
+            Command::Test { .. } => 12,
+            Command::ReportRewards { .. } => 13,
```
tmpolaczyk, line 90: We talked about just adding dummy variants to the solidity enum, like

Variant29, Variant30, Variant31, etc

ParthDesai: Okay. I have reverted this change.

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
ParthDesai: Okay. will do.

ParthDesai: So, I tried to write integration test for this, but it turns out that will be covered as part of the rewards integration test.

girazoki (approved):
pallet looks good, let's add the scripts that make the e2e test pass

primitives/bridge/src/lib.rs
ParthDesai: *fixed*

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
ParthDesai: *disregard*

pallets/oubound-message-commitment-recorder/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Message commitment from last block.
+    /// This will be set only when there are messages to relay.
+    #[pallet::storage]
+    pub type RecordedCommitment<T: Config> = StorageValue<_, H256, OptionQuery>;
+
+    impl<T: Config> Pallet<T> {
+        pub fn take_commitment_root() -> Option<H256> {
+            let maybe_commitment = RecordedCommitment::<T>::take();
+            if let Some(commitment) = maybe_commitment {
+                Pallet::<T>::deposit_event(Event::<T>::CommitmentRootRead { commitment });
```
tmpolaczyk, line 60: Next time you can use Option::inspect, but you can ignore this now

https://doc.rust-lang.org/std/option/enum.Option.html#method.inspect

ParthDesai: I can open up separate PR for this later on.

tmpolaczyk (approved):

tmpolaczyk:
The coverage job is failing with a very strange error, I don't care and I don't plan to fix it, let's merge

```
Error: 8 [ERROR] Error parsing file profraws: Record containing invalid integer: 'FN record missing for function _RINvYNtNtNtCsjZ2CtgZzVoM_5serde2de11ignored_any10IgnoredAnyNtB7_7Visitor12visit_stringNtNtCs1tnj5dV4HOf_4toml2de5ErrorECs4sSS8dRY7SX_14wasmtime_cache'
```

https://github.com/moondance-labs/tanssi/actions/runs/12829246475/job/35774927601?pr=802

pallets/oubound-message-commitment-recorder/src/lib.rs
ParthDesai: *future*

PR #803, opened 2025-01-13
Add timeout to signAndSendAndInclude

tmpolaczyk:
Sometimes zombienet tests fail because we send a transaction that is never included in a block. This PR adds some logs for that case, so instead of a generic "test timed out" we will see "Transaction timed out".


girazoki (approved):

tmpolaczyk:
Waiting for CI to fail to see the new error message, now obviously it's not failing

PR #807, opened 2025-01-13
Add more tests for `ExternalValidatorsRewards` pallet

Agusrodri:
### What does it do?

Adds more tests to `external-validators-rewards` pallet, more specifically covering (at both unit and integration level) that `OnEraEnd` hook behaves as expected, also checking that we are sending the correct information inside the `ReportRewards` command.

Note: a `RewardsMessageSent` event was added to the pallet for us to be able to track the information that is being sent.


girazoki (approved):
looks good agustin!

PR #808, opened 2025-01-14
Mint inflation due to external validators rewards to Ethereum Sovereign Account

nanocryk:
TODOs:

- [x] Pick appropriate Ethereum chain ID for Starlight. To which Ethereum chain it will connect to?


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1384,6 +1385,14 @@ impl Get<u64> for TimestampProvider {
     }
 }
 
+parameter_types! {
+    // Chain ID of Holesky.
+    pub RewardsEthereumSovereignAccount: AccountId =
+        tp_bridge::EthereumLocationsConverterFor::<AccountId>::convert_location(
+            &Location::new(1, [GlobalConsensus(Ethereum { chain_id: 17_000 } )])
```
tmpolaczyk, line 1392: I believe we already have a chain_id somewhere in `chains/orchestrator-relays/runtime/dancelight/constants/src/lib.rs`

Can we use that?

chains/orchestrator-relays/runtime/dancelight/src/tests/external_validators_tests.rs
```diff
@@ -1160,3 +1164,61 @@ fn external_validators_rewards_test_command_integrity() {
...
+                b"Channels",
+            ));
+            full_storage_key.extend_from_slice(&combined_channel_id_key);
+
+            let channel = Channel {
+                agent_id: H256::default(),
+                para_id: 1000u32.into(),
+            };
+
+            frame_support::storage::unhashed::put(&full_storage_key, &channel);
```
Agusrodri, line 1201: Try removing these lines, they should not be needed since we already have the genesis config for snowbridge in our `mod.rs` file, which takes care of setting up all the required storage:

```rust
        snowbridge_pallet_system::GenesisConfig::<Runtime> {
            // This is irrelevant, we can put any number here
            // as long as it is a non-used para id
            para_id: 1000u32.into(),
            asset_hub_para_id: 1001u32.into(),
            ..Default::default()
        }
        .assimilate_storage(&mut t)
        .unwrap();
```

chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
```diff
@@ -69,6 +69,9 @@ pub type LocationConverter = (
     AccountId32Aliases<ThisNetwork, AccountId>,
     // Foreign locations alias into accounts according to a hash of their standard description.
     HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>,
+    // Ethereum contract sovereign account.
+    // (Used to get convert ethereum contract locations to sovereign account)
```
Agusrodri, line 73:
```suggestion
    // (Used to convert ethereum contract locations to sovereign account)
```

chains/orchestrator-relays/runtime/dancelight/src/tests/external_validators_tests.rs
```diff
@@ -1160,3 +1164,38 @@ fn external_validators_rewards_test_command_integrity() {
             );
         });
 }
+
+#[test]
+fn external_validators_rewards_are_minted_in_sovereign_account() {
```
tmpolaczyk, line 1169: In this test Alice and Bob are whitelisted validators, which don't get rewards. However the inflation is still minted. Why is that?

nanocryk: I think that's a know issue of current design. Here I just mint the same amount as what is sent to Ethereum.

Agusrodri: Just an idea: would it worth to create a relation between tokens and points (and insert that into the formula)? This would help us to mint a consistent number into the eth's sovereign account at the end of each era.

For instance:

- Alice and Bob get 20 points each for creating one block.
- Suppose we have 10 blocks in one era. Alice creates the first five blocks and Bob the last five.
- We have 100 points for Alice, and 100 points for Bob. 200 points in total.
- There is a `RewardPointFactor` for instance of 0.5, so we mint (200 * 0.5) = 100 tokens on eth's sovereign account.

What do you think?

girazoki:
Can I ask for typescript integration tests too ;)? I think you would need to print the address and hardcode it somewhere in ts

primitives/bridge/src/lib.rs
```diff
@@ -320,3 +323,30 @@ impl ConvertMessage for DoNothingConvertMessage {
         Err(ConvertMessageError::UnsupportedVersion)
     }
 }
+
+pub struct EthereumLocationsConverterFor<AccountId>(PhantomData<AccountId>);
```
tmpolaczyk, line 327: I see a slightly different implementation upstream, can you add a comment to why we don't use that one?

https://github.com/paritytech/polkadot-sdk/blob/711e6ff33373bc08b026446ce19b73920bfe068c/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L467

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -245,11 +255,20 @@ pub mod pallet {
     impl<T: Config> tp_traits::OnEraEnd for Pallet<T> {
         fn on_era_end(era_index: EraIndex) {
             if let Some(utils) = Self::generate_era_rewards_utils(era_index, None) {
+                let tokens_inflated = T::EraInflationProvider::get();
+
+                let ethereum_sovereign_account = T::RewardsEthereumSovereignAccount::get();
+                if let Err(err) =
+                    T::Currency::mint_into(&ethereum_sovereign_account, tokens_inflated.into())
+                {
+                    log::error!(target: "ext_validators_rewards", "Failed to mint inflation into Ethereum Soverein Account: {err:?}");
```
tmpolaczyk, line 264: In case of error (which will never happen) maybe set `tokens_inflated` in the message to 0?

girazoki: yes

primitives/bridge/src/lib.rs
girazoki: because that uses parents 1, and we are itself a solo-chain (not backed by any parent)

girazoki:
SHould we add the ethereumSov account into the genesis with the minimum balance? I think it might be a good idea

tmpolaczyk (approved):

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1406,6 +1407,18 @@ impl Get<u64> for TimestampProvider {
     }
 }
 
+parameter_types! {
+    // Chain ID of Holesky.
+    pub RewardsEthereumSovereignAccount: AccountId =
```
ParthDesai, line 1412: Design wise, We should have one sovereign account for both rewards and token transfer to ethereum.

From what I understand, having one sovereign account for ethereum make more sense as having different sovereign accounts for rewards and token transfer will not work once we implement token transfer from ethereum to starlight.

nanocryk: For now we don't know so let's keep it like that. If we take the decision to have the same account we'll change the name and use it in multiple places.

pallets/external-validators-rewards/src/lib.rs
nanocryk: Changed. However couldn't this be a source of bugs on the Ethereum side if the amount is zero? There must be checks in case the reward amount is zero to not distribute rewards (which would perform a division by zero)

tmpolaczyk: Yes but the alternative is rewarding with tokens that don't exist. Let's make sure we have a test for zero amount in ethereum side.

Or don't send a rewards message at all, but not sure if that's better.

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
ParthDesai: What I mean is it will need to be one account if we are ever going to have token transfer. 

Otherwise there can be accounting mismatch.

girazoki: I agree that the naming can change to something more generic like `EthereumSovereignAccount`. not critical since we can change it later

pallets/external-validators-rewards/src/lib.rs
nanocryk: I feel like not sending a message would be cleaner, but I don't know if the Ethereum side depends on receiving a message every session.

tmpolaczyk: It shouldn't break in that case, as there are other error paths in this PR that already don't send a message. So you can change this to not send a message.

nanocryk: Other path being the message not being validated? Makes sense

chains/orchestrator-relays/runtime/dancelight/src/genesis_config_presets.rs
```diff
@@ -239,7 +239,9 @@ fn dancelight_testnet_genesis(
     invulnerables: Vec<String>,
     host_configuration: HostConfiguration,
 ) -> serde_json::Value {
-    let endowed_accounts: Vec<AccountId> = endowed_accounts.unwrap_or_else(testnet_accounts);
+    let mut endowed_accounts: Vec<AccountId> = endowed_accounts.unwrap_or_else(testnet_accounts);
+    endowed_accounts.push(crate::EthereumSovereignAccount::get());
```
tmpolaczyk, line 243: Put it below, in the existing

```
let accounts_with_ed =
```

tmpolaczyk (approved):

PR #809, opened 2025-01-14
Add test that tip is not charged when there are enough collators

tmpolaczyk:
Add a test to check that: when a chain sets a tip for collator assignment, it needs to have enough balance to cover the tip, even if the tip is not charged.


girazoki (approved):
thank you @tmpolaczyk add a short description please

tmpolaczyk:
@girazoki added another test that should fail now, it's being skipped.

PR #812, opened 2025-01-17
Add ethereum bridge testing support

dimartiro:
# Description

Add support for testing ethereum snowbridge in dancelight using xcm-emulator bridge features

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
@@ -137,7 +137,7 @@ impl BridgeMessageHandler for DancelightEthMockBridgeHandler {
             sent_msgs.borrow_mut().extend(msgs.clone());
         });
 
-        // TODO: return the right Vec<BridgeMessage> here
+        // TODO: We don't check the dispatches messages from the bridge so it's fine to return default value here
         Default::default()
     }
 


chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
```diff
@@ -104,17 +113,77 @@ decl_test_parachains! {
...
             FrontierTemplateDancelight,
             SimpleTemplateDancelight,
         ],
-        bridge = ()
+        bridge = DancelightEthMockBridge
     }
 }
 
+pub fn force_process_bridge() {
+    type DancelightRelayMessageQueue = <DancelightRelay as DancelightRelayPallet>::MessageQueue;
```
girazoki, line 173: I would make this more generic. Maybe we can pass a mocknet and adjust the rest of the values? maybe not, just asking for reference

girazoki:
Now that we know how this works, maybe we can add the dispatcher in the other direction? we should Call `EthereumInboundQueue` with the extrinsic `submit`  and the message. We can use the same pattern, i.e., we create a `send_inbound_message` function that overrides a thread variable and we implement the other direction of the bridge

chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocks.rs
```diff
@@ -0,0 +1,46 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+
+use snowbridge_beacon_primitives::{
+    types::deneb, BeaconHeader, ExecutionProof, VersionedExecutionPayloadHeader,
+};
+
+pub fn mock_execution_proof() -> ExecutionProof {
```
girazoki, line 21: we have this duplicated if I am not mistaken (although in another place it's called dummy_prood. I would put it in test/common/mod.rs and re-use it in both places

chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
```diff
@@ -104,17 +117,80 @@ decl_test_parachains! {
     }
 }
 
+thread_local! {
```
girazoki, line 120: add comments explaining what this does

girazoki (approved):

chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
```diff
@@ -104,17 +117,81 @@ decl_test_parachains! {
...
+            snowbridge_pallet_outbound_queue::Messages::<<DancelightRelay as Chain>::Runtime>::get()
+        });
+
+        // Store messages in our static mock buffer
+        ETH_BRIDGE_SENT_MSGS.with(|sent_msgs| {
+            sent_msgs.borrow_mut().extend(msgs.clone());
+        });
+
+        // TODO: return the right Vec<BridgeMessage> here
+        Default::default()
```
tmpolaczyk, line 141: So this is not needed?

chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/bridge_to_eth.rs
```diff
@@ -0,0 +1,151 @@
...
+        let mut new_validators =
+            <Dancelight as DancelightRelayPallet>::ExternalValidators::validators()
+                .into_iter()
+                .filter(|v| !whitelisted.contains(v))
+                .collect::<Vec<_>>();
+
+        let mut expected_validators = payload_validators.clone();
+
+        new_validators.sort();
+        expected_validators.sort();
```
tmpolaczyk, line 148: The pallet doesn't change the order, it stores validators in the same order as the received message, so maybe this sort is not needed

chains/orchestrator-relays/runtime/dancelight/src/tests/common/xcm/mocknets.rs
tmpolaczyk: I mean the TODO

tmpolaczyk: Then add it to the TODO: We don't check the dispatches messages from the bridge so it's fine to return Default::default here. Maybe explain why we can't do it now.

Agusrodri (approved):

tmpolaczyk (approved):

PR #815, opened 2025-01-22
change to use ubuntu 24.04 to support GLIBC_2.39

quangtuyen88:
The old image does not support  GLIBC_2.39. So since version v0.10.1 we are getting error : 
```
/tanssi/tanssi-node: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /tanssi/tanssi-node)
/tanssi/tanssi-node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /tanssi/tanssi-node)
/tanssi/tanssi-node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /tanssi/tanssi-node)
```

### DIFF
--- a/docker/container-chain-evm-template.Dockerfile
+++ b/docker/container-chain-evm-template.Dockerfile
@@ -6,11 +6,11 @@ FROM docker.io/library/ubuntu:20.04 AS builder
 
 RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificates
 
-FROM debian:bookworm-slim
+FROM ubuntu:24.04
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
Thanks for the PR, we are currently discussing whether we still want to support ubuntu 22.04 or not.

tmpolaczyk:
Released 0.10.2 to fix this issue, will work with ubuntu 22.04 so this change is not needed

PR #825, opened 2025-01-27
add first version of document about message handling

ParthDesai:


docs/bridge/messages.md
```diff
@@ -0,0 +1,129 @@
...
+## Payload
+```solidity
+return bytes.concat(
+    bytes4(0x70150038), // Magic bytes
+    bytes1(uint8(Message.V0)), // Message versioning bytes
+    bytes1(uint8(OutboundCommandV1.ReceiveValidators)), // Enum byte
+    ScaleCodec.encodeCompactU32(operatorsCount), // Operator count
+    operatorsFlattened, // Operators array
+    ScaleCodec.encodeU64(uint64(epoch)) // timestamp
+);
```
tmpolaczyk, line 24: Add link to github where this is defined, so when this document becomes outdated we can still find the new definition. Same for other code snippets

ParthDesai: Sure

ParthDesai:
Will reopen later.

PR #826, opened 2025-01-27
Send token id with rewards

girazoki:
This PR sends the token-id to be minted in the external chain in the rewards message. The goal here is dual:

- first, we wont send any rewards message (nor mint anything) until we have registered the token
- second, we can change the reward token whenever we want without affecting the SCs

The `register_token` benchmark has been written with a para-id module in mind, so I have made different setup for benchmarking to be able to test. A more programatic approach would be that we enable to decide what token we register upstream

I also had to create a benchmarkhelper in `externalValidatorRewards` to properly benchmark it


pallets/external-validators-rewards/src/lib.rs
```diff
@@ -120,6 +121,12 @@ pub mod pallet {
         /// Ethereum Sovereign Account where rewards will be minted
         type RewardsEthereumSovereignAccount: Get<Self::AccountId>;
 
+        // Token Location
```
Agusrodri, line 124:
```suggestion
        /// Token Location from Ethereum's point of view.
```

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -120,6 +121,12 @@ pub mod pallet {
...
+        type TokenLocationReanchored: Get<Location>;
+
+        // Location to Foreign Id
```
Agusrodri, line 127:
```suggestion
        /// How to convert from a given Location to a specific TokenId.
```

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -120,14 +121,33 @@ pub mod pallet {
...
+        type BenchmarkHelper: BenchmarkHelperTrait;
     }
 
     #[pallet::pallet]
     #[pallet::storage_version(STORAGE_VERSION)]
     pub struct Pallet<T>(_);
 
+    #[cfg(feature = "runtime-benchmarks")]
+    /// Helper trait to benchmark the on_era_end
+    pub trait BenchmarkHelperTrait {
```
Agusrodri, line 143: It would be nice to move this trait inside a common place like `primitives/traits` as it will also be needed for the benchmarks of the `ethereum-token-transfers` pallet.

0xvgg: and put it under benchmark macro if it's only used in benchmarks

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -40,15 +40,16 @@ use {
     parity_scale_codec::Encode,
     polkadot_primitives::ValidatorIndex,
     runtime_parachains::session_info,
-    snowbridge_core::ChannelId,
```
0xvgg, line 43: is it possible to decouple the snowbridge later through Trait?

pallets/external-validators-rewards/src/tests.rs
```diff
@@ -113,3 +116,41 @@ fn test_on_era_end() {
...
+        Mock::set_location(Location::parent());
+        ExternalValidatorsRewards::reward_by_ids([(1, 10), (3, 30), (5, 50)]);
+        ExternalValidatorsRewards::on_era_end(1);
+
+        let rewards_utils = ExternalValidatorsRewards::generate_era_rewards_utils(1, None);
+        let expected_command = Command::ReportRewards {
+            timestamp: 31000u64,
+            era_index: 1u32,
+            total_points: 90u128,
+            tokens_inflated: 42u128, // test inflation value used in mock
```
0xvgg, line 139: what is 42? can we calculate hte expected tokens somehow here in test?

same for 90u128, which is probably item[1] for item in [(1, 10), (3, 30), (5, 50)]

0xvgg (changes requested):

pallets/external-validators-rewards/src/lib.rs
girazoki: yes I can

pallets/external-validators-rewards/src/lib.rs
girazoki: ok

pallets/external-validators-rewards/src/lib.rs
girazoki: Later we can pass the tokenId through trait, but I dont see much  value doing that now tbf. We will now later (once this pallet becomes more generic) how people are going to use and what needs to be generalized

pallets/external-validators-rewards/src/lib.rs
girazoki: done @Agusrodri

pallets/external-validators-rewards/src/tests.rs
girazoki: will fix it

girazoki: resolved

0xvgg (approved):

Agusrodri (approved):
Looks good!

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -254,62 +264,70 @@ pub mod pallet {
...
-                }
+            let token_location = T::TokenLocationReanchored::get();
+            let token_id = T::TokenIdFromLocation::convert_back(&token_location);
+
+            if let Some(token_id) = token_id {
+                if let Some(utils) = Self::generate_era_rewards_utils(era_index, None) {
+                    let tokens_inflated = T::EraInflationProvider::get();
+
+                    let ethereum_sovereign_account = T::RewardsEthereumSovereignAccount::get();
+                    if let Err(err) =
```
ParthDesai, line 275: Non blocking question: What are the scenario in which this can fail?

girazoki: mm I guess if you try to mint an amount that is less than the ED for insance

ParthDesai (approved):

PR #835, opened 2025-01-29
`EthereumTokenTransfers` pallet

Agusrodri:
## What does it do?

This PR introduces the `pallet-ethereum-token-transfers` and adds it to Dancelight runtime, alongside:

- Benchmarking and weights
- Rust unit and integration tests
- Typescript integration tests

## Pallet's structure

The purpose of this pallet is to send the native Starlight token from Starlight to Ethereum. It does this by sending a `MintForeignToken` command to Ethereum through a specific `channel_id` (which is also stored in this pallet).

### Config types

- `Currency`: type to manage the internal transfers of native balance.
- `OutboundQueue`: implementation that takes care of validating and send the message to Ethereum.
- `EthereumSystemHandler`: Hook that manages the insertion of channel information inside _EthereumSystem_ pallet.
- `EthereumSovereignAccount`: Sovereign account of Ethereum in Starlight chain. This is the account in which we store and track how much tokens are transferred from Starlight to Ethereum as a whole.
- `FeesAccount`: the account in which we deposit fees (calculated at the moment of delivering the message).
- `TokenLocationReanchored`: the Starlight token location seen from the external chain's (Ethereum) point of view. This is the location that should be stored in _EthereumSystem_ pallet prior to transfer the token using the _EthereumTokenTransfers_ pallet.
- `TokenIdFromLocation`: this is how we convert a Location object to a TokenId, used inside the _MintForeignToken_ command. If the previous _TokenLocationReanchored_ is not found in _EthereumSystem_, this conversion fails. 

### Extrinsics

`set_token_transfer_channel`: only callable by root. Used to specify which channel_id will be used to send the tokens through. It also receives the para_id and agent_id params corresponding to the channel specified.

`transfer_native_token`: used to perform the actual sending of the tokens, it requires to specify an amount and a recipient, which is a H160 account on the Ethereum side.

### Storage

Inside the pallet we group and store the channel information using the `ChannelInfo` struct:

```rust
    pub struct ChannelInfo {
        pub channel_id: ChannelId,
        pub para_id: ParaId,
        pub agent_id: AgentId,
    }
```

```rust
    #[pallet::storage]
    pub type CurrentChannelInfo<T: Config> = StorageValue<_, ChannelInfo, OptionQuery>;
```

### Events 

- `ChannelInfoSet`: information for the channel was set properly.
- `NativeTokenTransferred`: some native token was successfully transferred to Ethereum.

### Errors

- `ChannelInfoNotSet`: the channel's information has not been set on this pallet yet.
- `UnknownLocationForToken`: conversion from Location to TokenId failed.
- `InvalidMessage`: the outbound message is invalid prior to send.
- `TransferMessageNotSent`: the outbound message could not be sent.


primitives/bridge/src/lib.rs
```diff
@@ -195,12 +195,20 @@ impl TicketInfo for () {
...
 impl<T: snowbridge_pallet_outbound_queue::Config> TicketInfo for Ticket<T> {
     fn message_id(&self) -> H256 {
         self.message_id
     }
 }
 
+#[cfg(feature = "runtime-benchmarks")]
+impl<T: snowbridge_pallet_outbound_queue::Config> TicketInfo for Ticket<T> {
+    fn message_id(&self) -> H256 {
+        H256::default()
```
tmpolaczyk, line 208: Why?

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -180,10 +184,43 @@ impl snowbridge_pallet_system::Config for Runtime {
     type WeightInfo = crate::weights::snowbridge_pallet_system::SubstrateWeight<Runtime>;
 }
 
+pub struct EthereumSystemHandler;
```
girazoki, line 187: can we move this to primitives? and we build it in general for any runtime that has `snowbridge_pallet_system`::Config implemented

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -0,0 +1,208 @@
...
+    frame_support::assert_ok,
+    snowbridge_core::{AgentId, ChannelId, ParaId},
+    sp_core::{H160, H256},
+    sp_runtime::traits::MaybeEquivalence,
+    sp_std::vec,
+    xcm::{latest::Location, VersionedLocation},
+};
+
+#[test]
+fn test_set_token_transfer_channel_reflects_changes_in_ethereum_system() {
```
girazoki, line 33: A few ideas on what to test here:

- Sending through a channel that is not yet being initialized yields error
- Sending an amount bigger than what you have errors
- Sending without the token registered yields error

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+
+            let channel_info = ChannelInfo {
+                channel_id,
+                para_id,
+                agent_id,
+            };
+
+            CurrentChannelInfo::<T>::put(channel_info.clone());
+
+            T::EthereumSystemHandler::create_channel(
```
girazoki, line 221: I dont recall what did we decide to do here. Does this create the channel directly?

girazoki: Ah it injects it, but does not send the message. Cool so we can set it to whatever

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        #[pallet::call_index(1)]
+        #[pallet::weight(T::WeightInfo::transfer_native_token())]
+        pub fn transfer_native_token(
+            origin: OriginFor<T>,
+            amount: u128,
+            recipient: H160,
+        ) -> DispatchResult {
+            let source = ensure_signed(origin)?;
+
```
girazoki, line 240: CurrentChannelInfo::<T>::get().ok_or(Error::<T>::ChannelInfoNotSet.into())

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+            if let Some(channel_info) = CurrentChannelInfo::<T>::get() {
+                let token_location = T::TokenLocationReanchored::get();
+                let token_id = T::TokenIdFromLocation::convert_back(&token_location);
```
girazoki, line 243:
```suggestion
                let token_id = T::TokenIdFromLocation::convert_back(&token_location).ok_or(Error::::UnknownLocationForToken.into())?
```

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+                    T::Currency::transfer(
+                        &source,
+                        &T::EthereumSovereignAccount::get(),
+                        amount.into(),
+                        Preservation::Preserve,
+                    )?;
+
+                    let message_id = ticket.message_id();
+
+                    log::error!("MESSAGE ID: {:?}", message_id);
```
girazoki, line 279:
```suggestion
```

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+            amount: u128,
+            recipient: H160,
+        ) -> DispatchResult {
+            let source = ensure_signed(origin)?;
+
+            if let Some(channel_info) = CurrentChannelInfo::<T>::get() {
+                let token_location = T::TokenLocationReanchored::get();
+                let token_id = T::TokenIdFromLocation::convert_back(&token_location);
+
+                if let Some(token_id) = token_id {
```
nanocryk, line 245: Prefer using `ok_or` in simple cases, or use
```rust
let Some(x) = pattern else {
    // possible log
    return Err(...);
};
```
In general avoid nesting if the else branch diverges.

Agusrodri: fixed in 8795a1ddff4a25ff3db24c24cd0ac8b97a95fc78.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: fixed in 8795a1ddff4a25ff3db24c24cd0ac8b97a95fc78.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: fixed in 8795a1ddff4a25ff3db24c24cd0ac8b97a95fc78.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: fixed in 8795a1ddff4a25ff3db24c24cd0ac8b97a95fc78.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: Yes exactly, it doesn't send any message to Ethereum. It only reflects the channel creation in `EthereumSystem` pallet.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: Done in 39c534df3c556f83d21d0a3d9d2bf4c4497af881.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -228,6 +265,20 @@ mod benchmark_helper {
             Ok(())
         }
     }
+
+    pub struct EthereumTokenTransfersBenchHelper;
+    impl tp_bridge::TokenChannelSetterBenchmarkHelperTrait for EthereumTokenTransfersBenchHelper {
```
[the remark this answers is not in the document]
Agusrodri: Done in 39c534df3c556f83d21d0a3d9d2bf4c4497af881.

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
Agusrodri: Added more tests in both c5e12f1411f286d7e1adff28d8df89f042419186 and c38fb12960b21180c340a2c4c86d05887094f12a. 

Let me know if they look good to you :)

girazoki (approved):
For me it's an approval, but we should create an issue to verify this through e2e tests. Can you create that @Agusrodri ?

Agusrodri:
> For me it's an approval, but we should create an issue to verify this through e2e tests. Can you create that @Agusrodri ?

Sure, I'll create the ticket.

primitives/bridge/src/lib.rs
Agusrodri: Added a small comment for this in 6b8572ba9dfd314617192e709e499d778bcd7cdc.

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+        type EthereumSystemHandler: EthereumSystemChannelManager;
+
+        /// Ethereum sovereign account, where native transfers will go to.
+        type EthereumSovereignAccount: Get<Self::AccountId>;
+
+        /// Account in which fees will be minted.
+        type FeesAccount: Get<Self::AccountId>;
+
+        /// Token Location from the external chain's point of view.
+        type TokenLocationReanchored: Get<Location>;
```
ParthDesai, line 128: Regarding token id, we might need to have root endpoint like channel where we check if the token id created from the parameter exists or not.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: But weren't the token-transfers messages supposed to be sent through the secondary governance channel? I thought that the goal of this extrinsic was to only have the possibility of "choosing" between the channels that are already created in the Ethereum side as well.

Is there something I'm missing? If that's the case let me know, I'm happy to implement the message sending part if needed :)

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: Do you mean that we should somehow check inside the pallet if the token id exists on the Ethereum side? 

If that's the case, from what I understand (correct me if I'm wrong), it's not the responsibility of this pallet to check if the token id we receive via the config parameter exists or not. We should directly use it to send the message.

But again, I'm happy to implement whatever we decide best!

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: Mmm I see. As we discussed initially, I thought that the goal of this extrinsic was to only reflect the channel creation inside `EthereumSystem` pallet without sending any message. I guess because we were going to manually create the channel on the Ethereum side?

If that's the case, there is one part of the `EthereumSystemHandler` implementation that is wrong, and it's the fact that it shouldn't return an error if the channel is already created, it should just be a no-op, because otherwise we would not be able to insert channels in the `EthereumTokenTransfers` pallet that are already created in the Ethereum side (and therefore in the `EthereumSystem` pallet, like the secondary governance channel for instance).

Anyway you are right, so I see two options: either we decide to send a message to Ethereum each time a channel is registered on `EthereumTokenTransfers` pallet or we manually create the channel on the Ethereum side before calling `set_token_transfer_channel` extrinsic.

What do you guys think? Let me know if I'm missing something :)

pallets/ethereum-token-transfers/src/lib.rs
ParthDesai: Problem with this is, if you send token id which is not registered on other side, the messages will be submitted but they will not be processed.

Since, we do not receive acknowledgement here, starlight state will imply we have sent it but ethereum side will not receive anything. This can cause user's assets locked and may require manual intervention to unlock via sudo.

pallets/ethereum-token-transfers/src/lib.rs
ParthDesai: Second option make more sense (we manually create the channel on the Ethereum side before calling set_token_transfer_channel extrinsic.) As channel creation is administrative operation and not something that will happen multiple times so no point in automating it via first option.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: And how would be the approach to check that inside the pallet (or Starlight runtime itself)? Do you have something in mind? Not sure how we can fetch that information from Ethereum.

pallets/ethereum-token-transfers/src/lib.rs
Agusrodri: Regarding the `EthereumSystemHandler` refactor, I pushed some changes in 120a5dfedcafc0166981bc4363a48f1db7afcecf to make the `set_token_transfer_channel` call infallible. 

This will allow us to use existing channels (and also new ones) for token-transferring like the governance ones if we want to.

pallets/ethereum-token-transfers/src/lib.rs
ParthDesai: Okay. This is checked in runtime by implementation of `TokenIdFromLocation`. So resolving this issue.

ParthDesai (approved):

PR #836, opened 2025-01-29
External Index handling as sent by external chain

girazoki:
Adds handling of the epoch sent externally as part of the validators. this is used later to pinpoint reward and slashes to a certain point in time. As part of this PR we do:

- store the external index (**which can be an epoch/timestamp/set_id or in general whatever you want that acts as an indetifier for the validators sent at a given point in time)** per era. Everytime we receive one we put it in storage item called `ExternalIndex`. Before the era start, we put this in `PendingExternalIndex`. At the era start, we put it in `CurrentExternalIndex`.

- `on_era_start` now sends this timestamp over. The goal is that pallet-external-validator-slashes stores this information in `BondedEras`. As this is the information later used to pinpoint slahes, we use it to pinpoint a particular slash to a external index.

- externalValidatorsRewards now sends `CurrentExternalndex` as part of the rewards message.

- Includes a migration to migrate currently bonded eras

To verify it works with try-runtime: `try-runtime   --runtime target/release/wbuild/dancelight-runtime/dancelight_runtime.compact.compressed.wasm   on-runtime-upgrade live   --uri wss://stagelight.tanssi-dev.network:443   --pallet ExternalValidatorSlashes --pallet CollatorConfiguration   --pallet Balances   --pallet Migrations  --pallet Configuration
`


pallets/external-validators/src/lib.rs
```diff
@@ -351,19 +363,23 @@ pub mod pallet {
         pub fn set_external_validators(
             origin: OriginFor<T>,
             validators: Vec<T::ValidatorId>,
+            timestamp: u64,
```
Agusrodri, line 366: We should rename it here as well right?

girazoki: yes, good catch

primitives/bridge/src/symbiotic_message_processor.rs
```diff
@@ -86,14 +86,16 @@ where
         match message {
             Message::V1(InboundCommand::ReceiveValidators {
                 validators,
-                timestamp: _timestamp,
+                timestamp,
```
tmpolaczyk, line 89: And here

pallets/external-validators/src/mock.rs
```diff
@@ -228,8 +228,14 @@ pub mod mock_data {
 
 #[derive(Clone, Encode, Decode, PartialEq, sp_core::RuntimeDebug, scale_info::TypeInfo)]
 pub enum HookCall {
-    OnEraStart { era: u32, session: u32 },
-    OnEraEnd { era: u32 },
+    OnEraStart {
+        era: u32,
+        session: u32,
+        timestamp: u64,
```
Agusrodri, line 234:
```suggestion
        external_idx: u64,
```

tmpolaczyk (approved):

chains/orchestrator-paras/runtime/common/src/migrations.rs
```diff
@@ -837,6 +841,96 @@ where
     }
 }
 
+pub struct MigrateMMRLeafPallet<T>(pub PhantomData<T>);
```
Agusrodri, line 844: Can we remove this migration?

girazoki: I tend to not like removing the migration until at least 3 runtime versions later. But we have removed it from the migration list

Agusrodri (approved):

ParthDesai (approved):

PR #838, opened 2025-01-31
Slash + rewards e2e test version 1

ParthDesai:
# Description
This PR implements slash + rewards e2e tests version 1. 

The version 1 only checks if the message is indicated to be successfully processed by gateway. But, it does not verify if it is processed by trying to claim the reward and actually checking if slash is applied. This will be implemented by version 2.

Other changes:
- We increased `max_dispatch_gas` for `ReportRewards` and `ReportSlashes` to 1 million from 60k gas.
- We modified the test to check operator block production to check if operator is part of session validators.
- We now start secondary governance channel relayer for token registration.

### DIFF
--- a/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
+++ b/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
@@ -5,6 +5,7 @@ import { signAndSendAndInclude, waitSessions } from "../../util/block.ts";
 import { ethers } from "ethers";
 import { decodeAddress } from "@polkadot/util-crypto";
 import { u8aToHex } from "@polkadot/util";
+import { MultiLocation } from "@polkadot/types/interfaces/xcm/types";
 
 // Change this if we change the storage parameter in runtime
 const GATEWAY_STORAGE_KEY = "0xaed97c7854d601808b98ae43079dafb3";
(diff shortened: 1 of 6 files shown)


girazoki (approved):

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -295,6 +325,86 @@ describeSuite({
...
+                        rewardMessageSuccess = success;
+                    } else if (slashMessageId == messageID) {
+                        slashMessageReceived = true;
+                        slashMessageSuccess = success;
+                    }
+                });
+
+                while (!rewardMessageReceived || !slashMessageReceived) {
+                    await sleep(1000);
+                }
```
tmpolaczyk, line 401: This is an infinite loop or I don't know javascript

girazoki: until the test times out right?

girazoki: which right now is set to 10 minutes since it starts

girazoki: maybe we should increase the timeout to 15, I think we are soon going to be on the limit

girazoki: image
Actually we should have timed out, anybody knows why we did not?

tmpolaczyk: I believe it's 10 minutes per test case yes, because zombie_data_preservers took 40 minutes to timeout.

girazoki: ok then I think it's ok

tmpolaczyk (approved):

PR #844, opened 2025-02-04
refactor: ♻️ Update repo to use biome

timbrinded:
> [!WARNING]  
> There is a very high number of files changes as a result of the change of linting. The github preview is slow as a result

## High Level Changes

- Update the tanssi repo to use [biome.js](https://biomejs.dev/) to replace both eslint and prettier. 
  - Not only is this faster to run, but is easier to maintain and more modern/better supported code. This reduces some of our tech debt and finds some common linting gotchas
- Fix the 1000+ linting issues on existing test files (typescript tech debt)
- Re-index test case IDs since we have many duplicates
- Add types to helpers (not exhaustive though)
- Update Package Versions


typescript-api/src/dancebox/interfaces/augment-api-consts.ts
```diff
@@ -24,324 +24,443 @@ export type __AugmentedConst<ApiType extends ApiTypes> = AugmentedConst<ApiType>
...
             [key: string]: Codec;
         };
         balances: {
             /**
              * The minimum amount required to keep an account open. MUST BE GREATER THAN ZERO!
              *
-             * If you _really_ need it to be zero, you can enable the feature `insecure_zero_ed` for this pallet. However, you
-             * do so at your own risk: this will open up a major DoS vector. In case you have multiple sources of provider
-             * references, you may also get unexpected behaviour if you set this to zero.
+             * If you *really* need it to be zero, you can enable the feature `insecure_zero_ed` for
```
tmpolaczyk, line 48: Why did this change? Because of a polkadot js update? If so, then a separate PR with only the upgrade will make this one smaller in size eventually

timbrinded: This was caused by biome formatting our comments.

girazoki (approved):

PR #845, opened 2025-02-04
Fix BABE smoke tests

Agusrodri:
### What does it do?

Fixes two smoke tests related to BABE, that were failing within a certain period of time in which session keys for validators were changed.

This PR aims to adapt the tests for future situations and make them consistent even if there is a change in session keys.


test/suites/smoke-test-dancelight/test-babe.ts
```diff
@@ -154,12 +161,9 @@ describeSuite({
                         // Verify seal signature
                         const message = hexToU8a(preHash);
                         const signature = hexToU8a(sealLog.toHex());
-                        const authorKeys = accountKeys.get(expectedAuthor);
-                        expect(
-                            authorKeys && authorKeys.babe,
-                            `Missing babe key for block author: ${expectedAuthor}`
-                        ).toBeTruthy();
-                        const pubKey = hexToU8a(authorKeys.babe);
+                        const authorBabe = accountsWithBabeKeys.find((acc) => acc[0] == expectedAuthor);
```
girazoki, line 164: I think will not be covering some cases, for instance, what happens if we are adding or removing one validator?

Here the more important problem is that we are doing the following:
- in the first test, we are going to the block where the session started, we are getting the validators that are outputed by the digest log and compare them against the current set of babe validaotrs: https://github.com/moondance-labs/tanssi/blob/b7b18d2d86d30a825d55a6c3b9a29470f32ec19f/test/suites/smoke-test-dancelight/test-babe.ts#L99

This is wrong, as the digest in session N is expressing the babe validators for session N+1, and we are comparing it against those in session N

What we should do instead is go to the session N-1 start block, get the digest, and compare it against those in session N

girazoki: I was mistaken as [0] already takes the previous epoch start block (and not current). So good job!

girazoki (approved):

PR #846, opened 2025-02-05
Fix and run benches

girazoki:
Fixes and runs benchmarks for all runtimes


chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
```diff
@@ -268,7 +268,7 @@ impl pallet_xcm::Config for Runtime {
     type ExecuteXcmOrigin = xcm_builder::EnsureXcmOrigin<RuntimeOrigin, LocalOriginToLocation>;
     type XcmExecuteFilter = Everything;
     type XcmExecutor = XcmExecutor<XcmConfig>;
-    type XcmTeleportFilter = Everything;
+    type XcmTeleportFilter = Nothing;
```
tmpolaczyk, line 271: This should change for benchmarks only? Or do we want to change it for real runtime as well?

girazoki: for everything. We should not allow teleports

tmpolaczyk (approved):

nanocryk (approved):

PR #847, opened 2025-02-05
Don't send message to Ethereum with zero rewards

nanocryk:
- Early return if token inflation or points are zero.
- Refactor the code to favor if-guard style (if => early return, avoid nesting)
- Update tests that assumed message was always emitted

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/tests/external_validators_tests.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/tests/external_validators_tests.rs
@@ -834,7 +834,10 @@ fn external_validators_rewards_merkle_proofs() {
                 vec![AccountId::from(CHARLIE), AccountId::from(DAVE)]
             );
 
-            // Validators are automatically rewarded.
+            // Reward all validators in era 1
+            crate::RewardValidators::reward_backing(vec![ValidatorIndex(0)]);
+            crate::RewardValidators::reward_backing(vec![ValidatorIndex(1)]);
+            
             assert_eq!(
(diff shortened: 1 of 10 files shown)


chains/orchestrator-relays/runtime/dancelight/src/tests/external_validators_tests.rs
```diff
@@ -849,6 +849,8 @@ fn external_validators_rewards_merkle_proofs() {
                     .unwrap();
             assert!(era_rewards.total > 0);
 
+            println!("era_rewards: {era_rewards:?}");
```
tmpolaczyk, line 852: remove

pallets/external-validators-rewards/src/lib.rs
```diff
@@ -271,70 +271,82 @@ pub mod pallet {
...
-                    );
+            let tokens_inflated = T::EraInflationProvider::get();
+
+            if tokens_inflated.is_zero() {
+                log::error!(target: "ext_validators_rewards", "Not sending message because tokens_inflated is 0");
+                return;
+            }
+
+            if utils.total_points.is_zero() {
+                log::error!(target: "ext_validators_rewards", "Not sending message because total_points is 0");
```
tmpolaczyk, line 298: Can we have this documented somewhere? Not sure if at the top of this function or where, but basically "Sends a message to ethereum, but only if there are rewards to distribute".

Or maybe in the Command::ReportRewards fields? Say that the rewards will never be 0?

tmpolaczyk (approved):

PR #858, opened 2025-02-07
Add pallet-multiblock-migrations to all runtimes

tmpolaczyk:
This pallet is needed for the stable2412 upgrade, because of a migration of pallet_identity. Currently there are no migrations.


tools/dev-benchmark-pallet.sh
```diff
@@ -68,28 +68,29 @@ fi
 
 # For each runtime, only update the weights if they already existed before
 # This is because the different runtimes have different pallets
-if [ -f "chains/orchestrator-paras/dancebox/src/weights/$PALLET.rs" ]; then
+# When adding a new pallet, you can set env variable FORCE_COPY=1
```
nanocryk, line 71: Add information about this in `docs/benchmarking.md`.

nanocryk (approved):

tools/dev-benchmark-pallet.sh
tmpolaczyk: Done, now I'm thinking that this will break if you add a pallet to some runtimes but not all of them, but that should be easy to fix by modifying the script when it happens...

Agusrodri (approved):

PR #859, opened 2025-02-07
Check if integrity_test hook is being tested

tmpolaczyk:


tmpolaczyk:
It is, as part of cargo test

PR #867, opened 2025-02-10
refactor: 🔧 Update dancelight smoketests

timbrinded:
Updates to the dancelight smoke tests to make more robust.

- Importing the generated interfaces so tests are type safe
- Reworked some state consistency checks so they don't halt on first failure
- Fixed type annotations to remove `any`, so future inconsistencies get flagged in IDE

image


test/suites/smoke-test-dancelight/test-authority-consistency-assignation.ts
```diff
@@ -17,22 +17,32 @@ describeSuite({
...
             title: "Collator assignation and authority assignation should match with observed mapping in containers",
             test: async () => {
-                const assignmentCollatorAccount = (
-                    await api.query.tanssiCollatorAssignment.collatorContainerChain()
-                ).toJSON();
+                const assignmentCollatorAccount = await api.query.tanssiCollatorAssignment.collatorContainerChain();
                 const sessionIndex = (await api.query.session.currentIndex()).toNumber();
                 const assignmentCollatorKey = (
                     await api.query.tanssiAuthorityAssignment.collatorContainerChain(sessionIndex)
-                ).toJSON();
```
tmpolaczyk, line 26: We abuse toJSON because it makes it easier to see the diff when an expect fails. Can you verify that expect(assignmentCollatorKey).toBeNull() shows a nice error message with the assignmentCollatorKey?

timbrinded: Will check. Problem about using toJSON() is that it casts it to `any`. What we currently have is that it will show you an error without even having to run the tests

girazoki:
@timbrinded can you merge master back to see the actual changes?

test/suites/smoke-test-dancelight/test-authority-consistency-assignation.ts
tmpolaczyk: @timbrinded please check this expect thing

timbrinded: When it fails it'll still show that blah does not contain "5DFi8fa9gABiFpWsSosS5xVBwGVy28crWmh1J…"

The added benefit here is that if it cant find it due to test error (i.e. the shape of the struct changes) then it's more obvious then what we get right now which is "cannot find .toHex() on undefined"

tmpolaczyk: When I do this:

```
expect(assignmentCollatorKey).toBe(null);
```

I see this, with about 10000 lines of type stuff:

imagen

... many lines more in between ....

imagen



On the other hand, if I use `.toJSON()`:

```
expect(assignmentCollatorKey.toJSON()).toBe(null);
```

I see this:

imagen

Much nicer

tmpolaczyk: Same applies to console log, removing toJSON makes debugging harder

timbrinded: I agree that toJSON is great for printing to console (and you can still do that!)

It makes for brittle and opaque test cases though, which in my experience with moonbeam you want to optimize to remove. Test case maintainance is a pain because you end up having to debug features you worked on years ago and forgot how it work. In that scenario a magic blob of JSON is really unhelpful because you dont know what the correct shape is.

Using the proper PolkadotJS types means you can do things like:
`expect(assignmentCollatorKey.isEmpty).toBe(true);`

to replicate your example.

timbrinded: Furthermore when you change the shape of the underlying structures you can pick up those errors at compile time instead of run time

timbrinded: @tmpolaczyk I've logged already on most test cases:
image

But i'll go through them all now and see if theres any other places you'd want to see it.

timbrinded: ok we should have best of both worlds now:
image

girazoki (approved):

PR #869, opened 2025-02-11
ci: :green_heart: Reduce self-hosted runner usage

timbrinded:
### Changes

- Moved clippy as dependent on build, so that it can use the cached results from build step
- ~~Split `rust-test` into generating the test files and then execution step.~~ Our runners have too slow network speed so do not benefit from sharding
- Made zombienet upgrade task a GHA runner step, since it only spawns 3 nodes

### Context 

The queue to use `self-hosted` runners can sometimes be over an hour long. 

This has introduced a blockages into the critical path, which can be ameliorated by lowering the usage of self-hosted for heavy tasks only.


.github/workflows/release.yml
```diff
@@ -394,8 +397,12 @@ jobs:
                   curl -L --retry 10 --silent --show-error https://github.com/rui314/mold/releases/download/v2.30.0/mold-2.30.0-$(uname -m)-linux.tar.gz | tar -C $(realpath mold) --strip-components=1 -xzf -
             - name: Setup Rust toolchain
               run: rustup show
+            - name: Install Protoc
+              uses: arduino/setup-protoc@v3
+            - name: Install nextest
+              uses: taiki-e/install-action@nextest
             - name: Unit tests
-              run: cargo test --release --all
+              run: cargo nextest run
```
tmpolaczyk, line 405: This can't work, our tests fail if not run with `--release`. Perhaps this is not running all tests? Or it is using release mode by default?

timbrinded:
I'm very close, let me cook

.github/workflows/release.yml
```diff
@@ -111,34 +112,39 @@ jobs:
...
+              uses: arduino/setup-protoc@v3
+            - name: Run sccache-cache
+              uses: mozilla-actions/sccache-action@v0.0.7
             - name: Check for duplicate git dependencies
               run: ./.github/scripts/check_duplicate_git_dependencies.sh
-
             - name: Find toml files with lints key not set
               run: ./.github/scripts/check_toml_lints.sh
-
-            # Do NOT add SKIP_WASM_BUILD=1 environment variable here as it skips wasm binary build which leads to some compilation bugs escaping clippy
```
tmpolaczyk, line 141: Don't remove this comment

.github/workflows/release.yml
```diff
@@ -223,33 +223,29 @@ jobs:
...
     build:
         runs-on: self-hosted
         needs: ["set-tags"]
         env:
             TMP_TARGET: "/tmp/target"
             CARGO_TARGET_DIR: "target"
+            RUSTC_WRAPPER: "sccache"
+            CARGO_INCREMENTAL: "0"
+            SCCACHE_CACHE_SIZE: "100GB"
+            SCCACHE_GHA_ENABLED: "true"
```
tmpolaczyk, line 235: We didn't have `SCCACHE_GHA_ENABLED` before, what does that do?

.github/workflows/release.yml
```diff
@@ -283,121 +279,115 @@ jobs:
...
+        name: "Check WASM runtimes with Twiggy"
+        runs-on: ubuntu-latest
+        if: github.event_name == 'pull_request'
         needs: ["set-tags", "build"]
         env:
-            TMP_TARGET: "/tmp/target"
-            CARGO_TARGET_DIR: "target"
-            RUSTC_WRAPPER: "sccache"
-            CARGO_INCREMENTAL: "0"
-            SCCACHE_CACHE_SIZE: "100GB"
```
tmpolaczyk, line 378: Why remove these ones?

.github/workflows/release.yml
timbrinded: GHA native sccache, haven't fully explored options with this:

https://github.com/mozilla/sccache/blob/main/docs/GHA.md

.github/workflows/release.yml
timbrinded: restoed

.github/workflows/release.yml
timbrinded: restored

tmpolaczyk (approved):

PR #872, opened 2025-02-13
Fix tanssi-relay and container collator version to match tanssi-node

tmpolaczyk:
`tanssi-relay --version` used to return 1.16.2, the polkadot-sdk node version. We want it to return 0.12.0, to match with the tanssi release version.

Also in the container collator rpc, calling system.version used to return 0.1.0. This is now fixed to the same version as `tanssi --version`, ensured by a unit test.


chains/orchestrator-relays/client/cli/src/command.rs
```diff
@@ -51,8 +51,7 @@ impl SubstrateCli for Cli {
     }
 
     fn impl_version() -> String {
-        let commit_hash = env!("SUBSTRATE_CLI_COMMIT_HASH");
-        format!("{}-{commit_hash}", NODE_VERSION)
+        env!("SUBSTRATE_CLI_IMPL_VERSION").into()
```
tmpolaczyk, line 54: This is still wrong, tanssi-relay --version returns 1.16.2

tmpolaczyk: Nevermind I didn't compile :see_no_evil: 

```
$ ./target/release/tanssi-relay --version
tanssi 0.10.0-b0cb99f258e
```

tmpolaczyk: Fixed to 0.12.0 with latest commit

girazoki (approved):

PR #873, opened 2025-02-14
first test workflow merge when ready

girazoki:
Gets rid of jobs that run in the merge queue and instead creates a new workflow that runs jobs upon setting the label `A8-mergeoncegreen`


timbrinded (approved):
looks good and surprisingly simple now. we can optimize the build step later, even if it means specifying a cache directory manually

timbrinded:
You just need to remove `zombienet-tests-merge-queue-bridge-e2e` and `zombienet-tests-merge-queue` from required CIs in repo settings.

PR #874, opened 2025-02-15
MD867 - Enable/Disable Slashing

augustocollado:
This PR introduces changes to the pallet `external-validator-slashes`.
It adds a new sudo-only extrinsic to switch between three modes for slashing:

- Enabled (default mode) - Slashing events generate an event (log), offenders are slashed
- LogOnly - Slashing events generates an event (log), and offenders are not slashed
- Disabled - Slashing events don't generate events (log),  and offenders are not slashed

It also adds tests for the three modes using babe offences.

**Note:** Weights are not correct, they need to be properly benchmarked.


girazoki (approved):
Good job @augustocollado !

pallets/external-validator-slashes/src/weights.rs
```diff
@@ -51,12 +51,15 @@
 use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
 use sp_std::marker::PhantomData;
 
+use crate::SlashingModeOption;
```
Agusrodri, line 54: I guess this import is not needed?

pnpm-lock.yaml
Agusrodri, line 1: I believe we shouldn't need to modify this file according the changes of the PR. Could you please check if it was uploaded by mistake? Thanks :)

test/suites/dev-tanssi-relay/slashes/test_slashingmodes.ts
```diff
@@ -0,0 +1,185 @@
...
+            const keyringBabe = new Keyring({ type: "sr25519" });
+
+            alice = context.keyring.alice;
+            aliceStash = keyringSr25519.addFromUri("//Alice//stash");
+            aliceBabePair = keyringBabe.addFromUri("//Alice");
+        });
+
+        it({
+            id: "E01",
+            title: "Slashing mode LogOnly should generate an event but not trigger a slash",
```
Agusrodri, line 31: Nice tests!

Agusrodri (commented):
Changes look good! Only a few small comments.

Also, I see a lot of formatting changes that should disappear after running `cargo fmt`. Can you please check that as well? Thanks!

Other than that the PR looks good. Great job 🥇

pallets/external-validator-slashes/src/weights.rs
augustocollado: You're right!

augustocollado:
> Also, I see a lot of formatting changes that should disappear after running `cargo fmt`. Can you please check that as well? 

All those files changed as a result of running `cargo fmt`.  In fact, the CI rust fmt check is failing only because I manually undid one change (image attached) that caused the compilation to fail. (Tomasz mentioned that this is a know issue related to the rust compiler version)
image

girazoki:
> > Also, I see a lot of formatting changes that should disappear after running `cargo fmt`. Can you please check that as well?
> 
> All those files changed as a result of running `cargo fmt`. In fact, the CI rust fmt check is failing only because I manually undid one change (image attached) that caused the compilation to fail. (Tomasz mentioned that this is a know issue related to the rust compiler version) image

augustocollado:
> > > Also, I see a lot of formatting changes that should disappear after running `cargo fmt`. Can you please check that as well?
> > 
> > 
> > All those files changed as a result of running `cargo fmt`. In fact, the CI rust fmt check is failing only because I manually undid one change (image attached) that caused the compilation to fail. (Tomasz mentioned that this is a know issue related to the rust compiler version) image

I undid this change because it wouldn't compile in my env. Probably it will compile and pass the fmt check is I change my rust version. I'll try that later!

augustocollado:
> > > Also, I see a lot of formatting changes that should disappear after running `cargo fmt`. Can you please check that as well?
> > 
> > 
> > All those files changed as a result of running `cargo fmt`. In fact, the CI rust fmt check is failing only because I manually undid one change (image attached) that caused the compilation to fail. (Tomasz mentioned that this is a know issue related to the rust compiler version) image

I've set my rust version to 1.81.0 stable, ran fmt, and among the many fmt changes (already pushed in this PR), there's one that breaks compilation:
image
That one change I manually undid to make the build check pass, but then the fmt check fails.
Not sure how to make both things (build & fmt) work.

augustocollado:
There was something wrong with my environment, but now using 1.81.0 stable fmt & builds are fine. Thanks @Agusrodri for the support!

PR #878, opened 2025-02-18
More StreamPayment config

nanocryk:
Adds 2 new config items:
- `minimum_request_deadline_delay`: currently stream parties can create a mandatory change request with an immediate/very close deadline, which can allow for the payer to stop paying immediately. This new item allows to set a minimal delay (in the stream time unit) so that the other party have fair time to respond.
- `soft_minimum_deposit`: Allows to configure a minimum amount of deposit when the stream source change it manually. It also prevents the source to close the stream if not stalled. This allows to guarantee to the target that payment will still occur for at least `soft_minimum_deposit / rate` time, and prevents the stream to be closed/emptied by surprise. This can be useful for non-automated services that could need time to prepare and react to the end of the service.

## Breaking changes

This changes the `StreamConfig` struct stored in this pallet but also in DataPreservers pallets as a config can be stored in a profile if the preserver wishes to provide services in exchange for a specific stream payment.

## TODOs:
- [x] Write migration


pallets/stream-payment/src/migrations.rs
```diff
@@ -0,0 +1,145 @@
...
+            })
+        });
+
+        let db_weights = T::DbWeight::get();
+        db_weights.reads_writes(count, count)
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade(&self) -> Result<Vec<u8>, sp_runtime::DispatchError> {
+        Ok(vec![])
```
girazoki, line 138: I'd say let's write pre-upgrade and post upgrade hooks. I think they should be easily testable with try-runtime.

Essentially, read one (if any) of the streams before, return it encoded. Receive it in post-upgrade encoded, decode it, and construct the new struct with the expected new fields. Verify that the storage has been migrated accordingly

girazoki (approved):
Minor comment about try-runtime, I think it's a good excercise that we always write those hooks

pallets/stream-payment/Cargo.toml
```diff
@@ -46,6 +49,7 @@ std = [
 	"frame-system/std",
 	"log/std",
 	"pallet-balances/std",
+	"pallet-migrations?/std",
```
girazoki, line 52: The way we usually do it is the following:

- in the pallet, we just write the funciton that migrates
- in the runtime/common we write a custom struct that implements the trait from pallet-migrations. This allows us to not depend on pallet-migrations in the pallets

girazoki (changes requested):
IMportant commens of formatting and question whether we should include both in dancebox and flashbox

girazoki (approved):
looks good now @nanocryk

PR #881, opened 2025-02-19
feat: offline marking support for collators

chexware:
This PR adds offline marking extrinsics  in `pallet-inactivity-tracking`.

TO DO:
- [x] Add Rust integration tests for offline marking using `set_offline` and `set_online` extrinsics
- [x] Add Moonwall tests for offline marking 
- [x] Move offline marking extrinsics to `pallet_inactivity_tracking`

### DIFF
--- a/test/moonwall.config.json
+++ b/test/moonwall.config.json
@@ -1074,7 +1074,7 @@
         },
         {
             "name": "dancelight_smoke",
-            "testFileDir": ["suites/smoke-test-common-solo", "suites/smoke-test-common-all"],
+            "testFileDir": ["suites/smoke-test-common-solo", "suites/smoke-test-common-all", "suites/smoke-test-dancelight"],
             "foundation": {
                 "type": "read_only"
             },
--- a/test/suites/smoke-test-common-solo/test-era-changes.ts
+++ b/test/suites/smoke-test-common-solo/test-era-changes.ts
@@ -1,0 +1,142 @@
+import "@tanssi/api-augment/dancelight";
+
+import { beforeAll, describeSuite, expect } from "@moonwall/cli";
+import type { ApiPromise } from "@polkadot/api";
+import { getCurrentEraStartBlock, getPastEraStartBlock } from "utils/block";
+
+describeSuite({
+    id: "SMOK16",
+    title: "Era changes suit that only runs on Dancelight chains",
+    foundationMethods: "read_only",
+    testCases: ({ it, context, log }) => {
+        let api: ApiPromise;
+        let currentEraIndex: number;
+        let currentEraStartBlock: number;
+        let pastEraStartBlock: number;
+
+        beforeAll(async () => {
+            api = context.polkadotJs();
+            currentEraStartBlock = await getCurrentEraStartBlock(api);
+            currentEraIndex = (await api.query.externalValidators.activeEra()).unwrap().index.toNumber();
+            pastEraStartBlock = await getPastEraStartBlock(api, currentEraStartBlock - 1);
+        });
+
+        it({
+            id: "C01",
+            title: "Era changes are happening as expected",
+            test: async () => {
+                const sessionsPerEra = api.consts.externalValidators.sessionsPerEra.toNumber();
+                const currentBlockNumber = await api.query.system.number();
+                const currentEraIndex = (await api.query.externalValidators.activeEra()).unwrap().index.toNumber();
+                const currentSessionIndex = (await api.query.session.currentIndex()).toNumber();
+
+                const apiAtCurrentEraStart = await api.at(await api.rpc.chain.getBlockHash(currentEraStartBlock));
+                const currentEraStartSessionIndex = (
+                    await apiAtCurrentEraStart.query.session.currentIndex()
+                ).toNumber();
+                const newEraEvents = (await apiAtCurrentEraStart.query.system.events()).filter(
+                    (eventRecord) => eventRecord.event.method === "NewEra"
+                );
+                const apiAtPreviousEraStart = await api.at(await api.rpc.chain.getBlockHash(pastEraStartBlock));
+                const previousEraStartSessionIndex = (
+                    await apiAtPreviousEraStart.query.session.currentIndex()
+                ).toNumber();
+
+                expect(newEraEvents.length).to.be.greaterThan(
+                    0,
+                    `No NewEra event found at the start of era ${currentEraIndex}`
+                );
+                expect(newEraEvents[0].event.data[0].toHuman()).to.be.equal(
+                    currentEraIndex.toString(),
+                    `NewEra event data does not match current era index ${currentEraIndex} at the start of the era.`
+                );
+                expect(previousEraStartSessionIndex + sessionsPerEra).to.be.equal(
+                    currentEraStartSessionIndex,
+                    `Error at block number ${currentBlockNumber}: Era change between era ${currentEraIndex - 1} and ${currentEraIndex} happened in ${currentEraStartSessionIndex - previousEraStartSessionIndex} sessions instead of ${sessionsPerEra} sessions.`
+                );
+                expect(currentSessionIndex).to.be.within(
+                    currentEraIndex * sessionsPerEra,
+                    (currentEraIndex + 1) * sessionsPerEra,
+                    `Error at block number ${currentBlockNumber}: Current session index ${currentSessionIndex} is not within the expected range for era ${currentEraIndex}.`
+                );
+            },
+        });
+
+        it({
+            id: "C02",
+            title: "Era slashes records are pruned as expected",
+            test: async () => {
+                const bondingDuration = api.consts.externalValidatorSlashes.bondingDuration.toNumber();
+                const apiAtCurrentEraStart = await api.at(await api.rpc.chain.getBlockHash(currentEraStartBlock));
+
+                // Verify that BoundedEras records are pruned correctly
+                const boundedEras = await apiAtCurrentEraStart.query.externalValidatorSlashes.bondedEras();
+                const boundedErasErrorRecords = [];
+                for (let i = 0; i < boundedEras.length; i++) {
+                    const eraIndex: number = boundedEras[i][0].toNumber();
+                    if (eraIndex < currentEraIndex - bondingDuration) {
+                        boundedErasErrorRecords.push(boundedEras[i]);
+                    }
+                }
+                expect(boundedErasErrorRecords.length).to.be.equal(
+                    0,
+                    `Found BoundedEras records outside of bonding duration: ${boundedErasErrorRecords.join("\n")}`
+                );
+
+                // Verify that ValidatorSlashInEra are pruned correctly
+                const validatorSlashInEra =
+                    await apiAtCurrentEraStart.query.externalValidatorSlashes.validatorSlashInEra.keys();
+                const validatorSlashInEraErrorRecords = [];
+                for (let i = 0; i < validatorSlashInEra.length; i++) {
+                    const validatorSlashEra = validatorSlashInEra[i].args[0].toNumber();
+                    if (validatorSlashEra < currentEraIndex - bondingDuration) {
+                        validatorSlashInEraErrorRecords.push(validatorSlashInEra[i]);
+                    }
+                }
+                expect(validatorSlashInEraErrorRecords.length).to.be.equal(
+                    0,
+                    `Found ValidatorSlashInEra records outside of bonding duration: ${validatorSlashInEraErrorRecords.join("\n")}`
+                );
+
+                // Verify that Slashes are pruned correctly
+                const slashes = await apiAtCurrentEraStart.query.externalValidatorSlashes.slashes.keys();
+                const slashesErrorRecords: number[] = [];
+                for (let i = 0; i < slashes.length; i++) {
+                    const slashesEra = slashes[i].args[0].toNumber();
+                    if (slashesEra < currentEraIndex - bondingDuration) {
+                        slashesErrorRecords.push(slashesEra);
+                    }
+                }
+                expect(slashesErrorRecords.length).to.be.equal(
+                    0,
+                    `Found Slashes records outside of bonding duration: ${slashesErrorRecords.join("\n")}`
+                );
+            },
+        });
+
+        it({
+            id: "C03",
+            title: "Era rewards records are pruned as expected",
+            test: async () => {
+                const rewardsHistoryDepth = api.consts.externalValidatorsRewards.historyDepth.toNumber();
+                const apiAtCurrentEraStart = await api.at(await api.rpc.chain.getBlockHash(currentEraStartBlock));
+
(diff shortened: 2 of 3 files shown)


nanocryk (changes requested):
In the PooledStaking pallet the added code messes with sensitive storages managed by the rest of the pallet, and doesn't take into account the frequent updates of `SortedEligibleCandidates` which will re-insert the candidate, being both in the eligible list and "offline". (see comments in specific parts of the code).

I would recommend adopting a different strategy: keep track of offline collators in the inactivity tracking pallet, and provide a type to use in `EligibleCandidatesFilter` that considers a candidate not eligible if it is offline (in case they are online it should fallback to the filter we already have right now). This also removes the added complexity from the PooledStaking pallet (no changes needed in it), which IMO is about tracking delegations and distributing rewards only. `EligibleCandidatesFilter` was designed to allow customizing who can be selected as a collator.

When marking a candidate offline, set it the inactivity tracking pallet and then call `Calls::<T>::update_candidate_position(&[offline_collator])`, which will properly remove the candidate from the sorted eligible list. When marking online, call the same function and it will put it back in the list (**with up to date stake**) if it is still eligible according to the other filters and if it still ranks in the top N candidates.

pallets/pooled-staking/src/lib.rs
```diff
@@ -671,6 +747,54 @@ pub mod pallet {
...
+                .into_iter()
+                .find(|c| c.candidate == collator.clone())
+                .ok_or(Error::<T>::CollatorDoesNotExist)?;
+
+            let _ = <OfflineCollators<T>>::try_mutate(|offline_collators| -> DispatchResult {
+                offline_collators.retain(|c| c.candidate != collator.clone());
+                Ok(())
+            })?;
+
+            let _ = <SortedEligibleCandidates<T>>::try_mutate(|eligible_candidates| {
```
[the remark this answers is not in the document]
chexware: `OfflineCollators ` storage updated to `StorageMap` in https://github.com/moondance-labs/tanssi/pull/881/commits/66d3f916552331bb62d6ab5cbfdbd1d220bfb770 and moved to `pallet-inactivity-tracking` in https://github.com/moondance-labs/tanssi/pull/881/commits/bd0c8fcf32347f7b63f73431c69562eb14df36be

pallets/pooled-staking/src/lib.rs
```diff
@@ -671,6 +747,54 @@ pub mod pallet {
...
+            );
+
+            let collator_info = <SortedEligibleCandidates<T>>::get()
+                .into_iter()
+                .find(|c| c.candidate == collator.clone())
+                .ok_or(Error::<T>::CollatorDoesNotExist)?;
+
+            let _ = <SortedEligibleCandidates<T>>::try_mutate(
+                |eligible_candidates| -> DispatchResult {
+                    eligible_candidates.retain(|c| c.candidate != collator.clone());
```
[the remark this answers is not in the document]
chexware: OfflineCollators  storage updated to StorageMap in https://github.com/moondance-labs/tanssi/commit/66d3f916552331bb62d6ab5cbfdbd1d220bfb770 and moved to pallet-inactivity-tracking in https://github.com/moondance-labs/tanssi/commit/bd0c8fcf32347f7b63f73431c69562eb14df36be

pallets/pooled-staking/src/lib.rs
```diff
@@ -374,6 +383,21 @@ pub mod pallet {
         ValueQuery,
     >;
 
+    /// Switch to enable/disable marking offline feature.
+    #[pallet::storage]
+    pub type EnableMarkingOffline<T: Config> = StorageValue<_, bool, ValueQuery>;
+
+    /// A list of offline collators
+    #[pallet::storage]
+    pub type OfflineCollators<T: Config> = StorageValue<
```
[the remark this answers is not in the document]
chexware: `OfflineCollators` storage updated to `StorageMap` in https://github.com/moondance-labs/tanssi/commit/66d3f916552331bb62d6ab5cbfdbd1d220bfb770 and moved to `pallet-inactivity-tracking` in https://github.com/moondance-labs/tanssi/commit/bd0c8fcf32347f7b63f73431c69562eb14df36be

chexware:
@nanocryk updated the code  so the majority of the offline marking logic is handled inside `pallet-inactivity-tracking` now besides the extrinsic calls. Do you reckon we should move them to `pallet-inactivity-tracking` as well?

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1880,6 +1880,7 @@ impl pallet_pooled_staking::Config for Runtime {
     type LeavingRequestTimer = SessionTimer<StakingSessionDelay>;
     type EligibleCandidatesBufferSize = ConstU32<100>;
     type EligibleCandidatesFilter = CandidateHasRegisteredKeys;
```
nanocryk, line 1882: `EligibleCandidatesFilter` still doesn't filter out inactive candidates, so calls to `update_candidate_position` will not remove them. `update_candidate_position` is the one using `EligibleCandidatesFilter` to know if it should appear in the list.

Since `CandidateHasRegisteredKeys` is defined in this file, you can modify it (with rename) to also check if the candidate is offline or not.

nanocryk (changes requested):
Indeed, you should now be able to have the extrinsics in `inactivity-tracking` pallet, and call pooled staking pallet update function (probably through a trait like you did with `NodeActivityTrackingHelper`. Something like `NotifyInactivityChange` with a `notify_inactivity_change(AccountId)` function, that will call `update_candidate_position`.

IMO you shouldn't need any change in pooled staking pallet :)

pallets/pooled-staking/src/lib.rs
```diff
@@ -754,4 +757,30 @@ pub mod pallet {
             pools::distribute_rewards::<T>(&candidate, rewards)
         }
     }
+    impl<T: Config> tp_traits::NotifyCollatorOnlineStatusChange<Candidate<T>> for Pallet<T> {
+        fn is_collator_in_sorted_eligible_candidates(collator: &Candidate<T>) -> bool {
+            <SortedEligibleCandidates<T>>::get()
+                .into_iter()
+                .any(|c| c.candidate == collator.clone())
```
nanocryk, line 764: Can it be `&c.candidate == collator` to avoid cloning?

primitives/traits/src/lib.rs
```diff
@@ -615,12 +615,45 @@ pub trait ExternalIndexProvider {
...
+// A trait to help remove offline collators from the pending assignment
+pub trait PendingCollatorAssignmentHelper<AccountId> {
+    /// Remove a collator from the current pending collator assignment.
+    fn remove_offline_collator_from_pending_assignment(collator: &AccountId);
+}
+
+// A trait to update the collators rewards when a collator's online status changes.
+pub trait NotifyCollatorOnlineStatusChange<AccountId> {
+    /// Check if the collator is in SortedEligibleCandidates list.
+    fn is_collator_in_sorted_eligible_candidates(collator: &AccountId) -> bool;
```
nanocryk, line 653: Make the function names a bit more agnostic? Something like `is_candidate_selected` and `on_online_status_change` ? In case other implementations needs it, maybe include the online status as a bool here ? (even if we don't need it for pooled staking)

pallets/pooled-staking/src/lib.rs
```diff
@@ -754,4 +757,30 @@ pub mod pallet {
...
+            let minimum_stake = T::MinimumSelfDelegation::get();
+            T::Currency::set_balance(collator, minimum_stake + minimum_stake);
+            T::EligibleCandidatesFilter::make_candidate_eligible(collator, true);
+            Calls::<T>::request_delegate(
+                collator.clone(),
+                collator.clone(),
+                ActivePoolKind::AutoCompounding,
+                minimum_stake,
+            );
+            T::JoiningRequestTimer::skip_to_elapsed();
```
nanocryk, line 783: Once the timer is elapsed you need to execute the pending operation.

nanocryk (commented):
Small nitpick + small issue in benchmark hook. Otherwise it looks pretty good !

pallets/pooled-staking/src/lib.rs
chexware: Yup updated it in https://github.com/moondance-labs/tanssi/pull/881/commits/8f345e6068d5edd50e2809f61c32e6b0d3732c7b

pallets/pooled-staking/src/lib.rs
```diff
@@ -98,6 +100,8 @@ pub mod pallet {
         fungible::Credit<<T as frame_system::Config>::AccountId, <T as Config>::Currency>;
     pub type Delegator<T> = <T as frame_system::Config>::AccountId;
 
+    pub type SessionIndex = u32;
```
nanocryk, line 103: No longer needed?

pallets/pooled-staking/src/lib.rs
```diff
@@ -754,4 +757,40 @@ pub mod pallet {
...
+            Calls::<T>::update_candidate_position(&[candidate.clone()])
+        }
+
+        #[cfg(feature = "runtime-benchmarks")]
+        fn make_collator_eligible_candidate(collator: &Candidate<T>) {
+            use sp_std::vec;
+            let minimum_stake = T::MinimumSelfDelegation::get();
+            T::Currency::set_balance(collator, minimum_stake + minimum_stake);
+            T::EligibleCandidatesFilter::make_candidate_eligible(collator, true);
+            Calls::<T>::request_delegate(
```
nanocryk, line 779: Make sure fallible calls succeed using `.expect()`, otherwise if it breaks for some reason it'll be harder to find the cause.

pallets/pooled-staking/src/lib.rs
chexware: Yes removed in https://github.com/moondance-labs/tanssi/pull/881/commits/461c1c672e4b12c8752138d29eb326f7d38fe3e9

pallets/pooled-staking/src/lib.rs
chexware: Good point! Updated them in https://github.com/moondance-labs/tanssi/pull/881/commits/3a116ecd6275c2d5b05cb220118b2cf019298151

pallets/collator-assignment/src/lib.rs
```diff
@@ -670,6 +671,35 @@ pub mod pallet {
             }
         }
     }
+
+    impl<T: Config> PendingCollatorAssignmentHelper<T::AccountId> for Pallet<T> {
+        fn remove_offline_collator_from_pending_assignment(collator: &T::AccountId) {
+            // If there is a  pending assignment, we remove the collator from it
```
girazoki, line 677: @tmpolaczyk I would say we cannot do that. we cannot remove them from pending assignment nor current. that would disturb current assignments and therefore harm existing chains. what we have to do is simply mark it as non-online in pallet-staking, making it non-eligible anymore in the next sessions

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -222,6 +253,42 @@ pub mod pallet {
...
+        pub fn enable_offline_marking(origin: OriginFor<T>, value: bool) -> DispatchResult {
+            ensure_root(origin)?;
+            <EnableMarkingOffline<T>>::set(value);
+            Ok(())
+        }
+
+        #[pallet::call_index(2)]
+        #[pallet::weight(T::WeightInfo::set_offline())]
+        pub fn set_offline(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+            let collator = ensure_signed(origin)?;
```
girazoki, line 268: what is this for? marking one himself offline or offline?

pallets/pooled-staking/src/lib.rs
```diff
@@ -754,4 +755,42 @@ pub mod pallet {
...
+        fn is_candidate_selected(candidate: &Candidate<T>) -> bool {
+            <SortedEligibleCandidates<T>>::get()
+                .into_iter()
+                .any(|c| &c.candidate == candidate)
+        }
+        fn on_online_status_change(
+            candidate: &Candidate<T>,
+            _is_online: bool,
+        ) -> DispatchResultWithPostInfo {
+            Calls::<T>::update_candidate_position(&[candidate.clone()])
```
girazoki, line 768: in my opinion this is where you should mark collators as offline or not.  you need to make sure that pallet-staking filters out non-eligible collators (because they are offline). So I think it makes sense to have a fucntion here called sorted_eligible_candidates where you filter out those offline, hence the storage item needs to be in this pallet (the extrinsic though, can be in another pallet). Or you create a struct here that applies the filder from inactivity pallet.

pallets/inactivity-tracking/src/lib.rs
chexware: Yes it allows the collator to mark itself offline

tmpolaczyk: Add a doc comment, so it also shows in polkadot js :)

chexware: Added doc comments for the extrinsics in https://github.com/moondance-labs/tanssi/pull/881/commits/391e86b311e5d48a376a397f47bed177cf250c8f

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -418,11 +488,52 @@ pub mod pallet {
...
+            Ok(().into())
+        }
+
+        pub fn mark_collator_online(collator: &Collator<T>) -> DispatchResultWithPostInfo {
+            ensure!(
+                <OfflineCollators<T>>::get(collator),
+                Error::<T>::CollatorNotOffline
+            );
+            <OfflineCollators<T>>::insert(collator.clone(), false);
+            T::CollatorStakeHelper::on_online_status_change(collator, true)?;
```
tmpolaczyk, line 525: Add a comment here that this removes the elegible staking candidates list, and it needs to be called after updating the `OfflineCollators` map.

chexware: Comments added in https://github.com/moondance-labs/tanssi/pull/881/commits/2e9c7eb6fa96ebe337f569d5e70853c0d53886c1

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -418,11 +488,52 @@ pub mod pallet {
...
+            });
+            Ok(().into())
+        }
+
+        pub fn mark_collator_online(collator: &Collator<T>) -> DispatchResultWithPostInfo {
+            ensure!(
+                <OfflineCollators<T>>::get(collator),
+                Error::<T>::CollatorNotOffline
+            );
+            <OfflineCollators<T>>::insert(collator.clone(), false);
```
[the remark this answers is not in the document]
chexware: Updated in https://github.com/moondance-labs/tanssi/pull/881/commits/2e9c7eb6fa96ebe337f569d5e70853c0d53886c1

PR #882, opened 2025-02-22
WIP: test: 🧪 New RuntimeAPI for test Helpers

timbrinded:
> [!IMPORTANT]  
> *Still Work In Progress* 🏗️
> Need to add tests that actually use this runtime api properly, as well as ensuring it's part of `api-augment` package. Plus adding it to all runtimes

This adds a new RuntimeAPI for where we want to place helper functions. Currently this only contains a single function: to get the `Session::Period` for the runtime, which is useful in knowing whether a connected chain is using a fast runtime or slow. Down the line, we can add new methods to this runtime to aid in testing and debugging by providing more visibility of privileged chain constants.


test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -11,37 +11,50 @@ describeSuite({
...
 
         beforeAll(async () => {
-            api = context.polkadotJs();
+            api = context.pjsApi;
 
-            const rtBefore = api.consts.system.version.specVersion.toNumber();
-            const sessionBefore = api.query.session.currentIndex();
-            log("About to upgrade to runtime at:");
-            log((await MoonwallContext.getContext()).rtUpgradePath);
+            console.dir((await api.call.tanssiUtilApi.sessionPeriod()).toNumber(), { depth: 1 });
```
girazoki, line 16: we probably cannot call this before the runtime is up is that right?

timbrinded: that's correct

girazoki: can we just do an if runtime > 1100 then call thsi?

PR #884, opened 2025-02-24
AuthorNoting should only deal with parachains with assigned collators (instead of all)

nanocryk:
Currently the runtime checks that the inherents contain the relay proof of all registered parachains to notice new blocks.

If a chain doesn't have assigned collators, we know there can't be new blocks, so it's safe to skip them in the runtime, and thus not insert them in the inherents. This PR thus updates both runtime and client sides to only check/insert inherents related to parachains with assigned collators.


chains/orchestrator-paras/node/src/service.rs
```diff
@@ -566,9 +568,20 @@ fn start_consensus_orchestrator(
             let relay_chain_interface = relay_chain_interace_for_cidp.clone();
             let client_set_aside_for_cidp = client_set_aside_for_cidp.clone();
             async move {
-                let para_ids = client_set_aside_for_cidp
+                // We added a new runtime api that allows to know which parachains have
+                // some collators assigned to them. We'll now only include those. For older
+                // runtimes we continue to write all of them.
```
tmpolaczyk, line 573: Can someone explain what does this mean? (and add to PR description). Do we include the headers for all parachains even if the block number has not changed? Or is this only an optimization to not have to check all the headers, but the created inherent data will be the same with and without the runtime api?

nanocryk: To know that the block number didn't change you have to take a look at the relay, which is what the proof allows. We know that if the parachain don't have assigned collators, new blocks won't be produced, so we can skip that and include a smaller amount of proofs.

This PR update both the runtime part that verifies all the chains have their proof (now only the ones with assigned collators), and the client part that will now only insert the required proofs.

tmpolaczyk: Ok thanks! I was missing the fact that even if the block number does not change, we still need to include the header in the inherent to prove that it didn't change.

pallets/author-noting/src/lib.rs
```diff
@@ -136,7 +140,7 @@ pub mod pallet {
     #[pallet::call]
     impl<T: Config> Pallet<T> {
         #[pallet::call_index(0)]
-        #[pallet::weight((T::WeightInfo::set_latest_author_data(<T::ContainerChains as GetCurrentContainerChains>::MaxContainerChains::get()), DispatchClass::Mandatory))]
+        #[pallet::weight((T::WeightInfo::set_latest_author_data(T::MaxContainerChains::get()), DispatchClass::Mandatory))]
```
girazoki, line 143: why did we change this? do you think this is better than having the max amount as part of the trait?

girazoki (approved):
Changes overall look good! small comments

girazoki:
Actually I just realized there is one change we need to do:

registered_paras() currently tries to detect whether the next block is going to change the session. If so, we use the scheduled registered paras for the next session. See https://github.com/moondance-labs/tanssi/blob/56a39f51ea9951a498b361e900b38fc45b33fc4f/chains/orchestrator-paras/runtime/flashbox/src/lib.rs#L2055

The reason behind this is that runtime-apis are called with the state on-finalize of the previous block. And if the previous block is a session-boundary, we will have the wrong things injected here.

MY suggestion is that you keep the new `parachains_with_some_collators`. Meaning if it's a session boundary block, we take the nextPendingCollators or whatever the name is

pallets/author-noting/src/lib.rs
```diff
@@ -148,7 +152,11 @@ pub mod pallet {
                 "DidSetContainerAuthorData must be updated only once in a block",
             );
 
-            let registered_para_ids = T::ContainerChains::current_container_chains();
+            let registered_para_ids: Vec<_> =
```
girazoki, line 155:
```suggestion
            let container_chains_to_check: Vec<_> =
```

tmpolaczyk:
@nanocryk dev_tanssi tests are failing because the author noting weight changed. Could you update the test to use this syntax? So that we see the new value the next time the test fails:

```js
const min = (CLEAR_ORIGIN_WEIGHT * 90n) / 100n;
const max = (CLEAR_ORIGIN_WEIGHT * 110n) / 100n;

expect(
    weight,
    `weightMessage returned ${weight} but expected a value between ${min} and ${max}`
).to.satisfy((val) => val >= min && val <= max);
```

girazoki:
> @nanocryk dev_tanssi tests are failing because the author noting weight changed. Could you update the test to use this syntax? So that we see the new value the next time the test fails:
> 
> ```js
> const min = (CLEAR_ORIGIN_WEIGHT * 90n) / 100n;
> const max = (CLEAR_ORIGIN_WEIGHT * 110n) / 100n;
> 
> expect(
>     weight,
>     `weightMessage returned ${weight} but expected a value between ${min} and ${max}`
> ).to.satisfy((val) => val >= min && val <= max);
> ```

this is actually a good sign, I do believe that we have 2 registered chains but only 1 gets collators. If that is the case the weight reduction would make sense. Can you check that?

test/suites/dev-tanssi/weights/test_set_latest_author_data_weight.ts
```diff
@@ -14,8 +14,12 @@ describeSuite({
...
             test: async () => {
                 // TODO: is it expected that this test breaks, just copy the new weights
-                const expectedRefTime = new BN(912947136);
-                const expectedProofSize = new BN(8046);
+                const expectedRefTime = { avg: new BN(786288068) };
+                expectedRefTime.min = expectedRefTime.avg.divn(1.1);
+                expectedRefTime.max = expectedRefTime.avg.muln(1.1);
+                const expectedProofSize = { avg: new BN(8046) };
+                expectedProofSize.min = expectedProofSize.avg.divn(1.1);
+                expectedProofSize.max = expectedProofSize.avg.muln(1.1);
```
tmpolaczyk, line 22: Thanks, love it. Now we see proper errors:

```
AssertionError: proofSize is 5507 but expected a value between 7314 and 8850: expected { Object (negative, words, ...) } to satisfy [Function]
```

tmpolaczyk (approved):

girazoki (approved):

PR #885, opened 2025-02-24
Minor doc update and Node version compatibility

evgeny-s:
**Description**
1. In the case of local test execution, the arc64 platform is not supported, because it pulls x64 by default.
The Polkadot binaries should be built from sources and copied manually.

2. Websocket is stable starting from Node v22. So it should force switch the Node version using `.nvmrc`.


girazoki (approved):

timbrinded:
We could go further by detecting arch type Darwin (e.g. using the node os built in) in the download polkadot script and skipping (maybe exiting if no polkadot bins found).

evgeny-s:
> We could go further by detecting arch type Darwin (e.g. using the node os built in) in the download polkadot script and skipping (maybe exiting if no polkadot bins found).

or throw the exception that binary for the current platform is not found. 
I checked how moonwall is working, we need either extend the repo binaries in this repo, or even extend it in the test configuration.

timbrinded:
@evgeny-s no need for moonwall change, just change this:
https://github.com/moondance-labs/tanssi/blob/master/test/scripts/downloadPolkadot.ts

Here is an example of arch filtering:
https://github.com/Moonsong-Labs/storage-hub/blob/main/test/scripts/crossBuildMac.ts

PR #891, opened 2025-02-27
fix smoke test regarding external validators rewards

girazoki:
There was an error in a smoke test regarding how we check whether rewards expire. And this was because polkadotjs by default returns a default value there, so we cannot check this properly. Instead I changed it to retrieve all keys, and then check whether the key exists

### DIFF
--- a/test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
+++ b/test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
@@ -99,6 +99,8 @@ describeSuite({
             test: async () => {
                 const historyDepth = api.consts.externalValidatorsRewards.historyDepth;
 
+                console.log(historyDepth);
+
                 // Checkpoint A: current era index - historyDepth
                 const eraIndexCheckpointA =
                     (await api.query.externalValidators.activeEra()).unwrap().index.toNumber() -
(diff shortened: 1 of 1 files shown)


test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
```diff
@@ -99,21 +99,24 @@ describeSuite({
             test: async () => {
                 const historyDepth = api.consts.externalValidatorsRewards.historyDepth;
 
+                console.log(historyDepth);
```
timbrinded: is this needed?

test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
```diff
@@ -99,21 +99,24 @@ describeSuite({
...
-                const validatorRewardCheckpointB =
-                    await api.query.externalValidatorsRewards.rewardPointsForEra(eraIndexCheckpointB);
-
-                expect(validatorRewardCheckpointA.isEmpty).toBe(true);
-                expect(validatorRewardCheckpointB.isEmpty).toBe(false);
+                // The mapping only contains the keys that are in `externalValidatorsRewards`
+                const rewardMappingKeys = (await api.query.externalValidatorsRewards.rewardPointsForEra.keys()).map(
+                    (key) => {
+                        // Convert "1,020" into 1020
+                        return Number.parseInt(key.toHuman().toString().replace(",", ""));
```
timbrinded, line 113: Why not just convert key to number instead of human to string and replace the commas:

`(key) => key.toNumber()`

evgeny-s: @timbrinded , there is no `toNumber` helper for key.

evgeny-s (approved):

test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
girazoki: jsut copied it from another place, will apply this change

timbrinded (approved):
LGTM :+1:

PR #892, opened 2025-02-27
Removed runtime build with fast-runtime feature

evgeny-s:
[Ticket](https://opslayer.atlassian.net/browse/MD-917)

**Description**
After the runtime build is removed for fast-runtime, it affected several more files that was fixed.
Please make sure we don't need those as well.


chains/orchestrator-relays/node/tanssi-relay-service/src/chain_spec.rs
```diff
@@ -213,46 +211,6 @@ pub fn dancelight_development_config(
...
-        })
-        .chain(
-            mock_container_chains
-                .iter()
-                .map(|x| (*x, mock_container_chain_genesis_data(*x), vec![])),
-        )
-        .collect();
-
-    Ok(DancelightChainSpec::builder(
-        dancelight::fast_runtime_binary::WASM_BINARY
```
tmpolaczyk, line 241: Try to keep this function but use the non fast runtime binary here. For zombienet tests we compile with --features=fast-runtime, so if this is implemented correctly, it will be a fast runtime even if the name doesn't say it.

PR #893, opened 2025-02-27
CI: ⚙️ Improve E2E Setup

timbrinded:
Changes our e2e setup scripts to make it easier to maintain and quicker to run.

- We no longer checkout and build `lodestar` or `geth`: We instead download pre-built binaries from GH releases and dockerhub respectively
- We don't modifiy lodestar chain config, we pass it as a runtime arg via: `--params.SECONDS_PER_SLOT 1`
- Modified the cleardown and build scripts to improve performance on sucessive local runs (wont try to checkout and build everything from scratch everytime when testing locally)
- Cached build files in the CI to speed up runs (foundry, symbiotic, snowbridge contracts). - 
  - This is keyed by the contents of `set-env.sh`, due to it containing the version constants used for checking out and building these repos


.github/workflow-templates/bridge-e2e/action.yml
```diff
@@ -67,6 +114,11 @@ runs:
       id: check_yarn
       run: |
         yarn --version
+    - name: Use Node.js 22.x
```
evgeny-s, line 117: Probably we need to fix it in the other places, but since we added the `.nvmrc` recently, probably it's a good idea to read the version from there, in this case, we can update all the CI in 1 place. The example:
```
      - name: Read .nvmrc
        id: nvmrc
        run: echo "##[set-output name=NODE_VERSION;]$(cat .nvmrc)"
      
      - name: 'Set up node'
        uses: actions/setup-node@v2
        with:
          node-version: "${{ steps.nvmrc.outputs.NODE_VERSION }}"
```

timbrinded: parameterized this now and given it a default val. the nvmrc step doesnt exist in the `merge-when-ready.yml` so that will need to be added down the road.

.github/workflows/merge-when-ready.yml
```diff
@@ -96,7 +96,7 @@ jobs:
                   chmod uog+x target/release/container-chain-simple-node
                   chmod uog+x target/release/container-chain-frontier-node
 
-            - name: Run Zombienet Test ${{ matrix.test_name }}
```
girazoki, line 99: why removing the matrix name?

test/scripts/bridge/cleanup.sh
```diff
@@ -19,9 +19,9 @@ elif [ $1 = "ole" ]; then
   rm -rf $logs_dir
   rm -rf $output_dir
   rm -rf $ethereum_data_dir
-elif [ $1 = "olep" ]; then
+elif [ $1 = "lep" ]; then
   rm -rf $logs_dir
-  rm -rf $output_dir
```
girazoki, line 24: we cannot remove the output dir now?

girazoki: I think if anything, I would add the option for that at least.

.github/workflows/merge-when-ready.yml
timbrinded: because it's not a matrix

girazoki: wait how is it not? https://github.com/moondance-labs/tanssi/blob/3d9025a271799ff917455670ff4657a3ac9179bb/.github/workflows/merge-when-ready.yml#L47

test/scripts/bridge/cleanup.sh
timbrinded: removing the bins is unnessecary step and adds time to subsequent local runs (not to mention breaks caching). Have changed it so we remove everything but the bins

timbrinded: the bin versions are determined by constants `set-env.sh`. 
- CI: It will always download the version set here or restore from cache (faster) if it's been downloaded previously 
- Local: It will always use what's already downloaded and print warning if it doesn't match what is set in `set-env.sh`. This allows you to test against different versions locally of the eth nodes

test/scripts/bridge/build-ethereum-node.sh
```diff
@@ -7,54 +7,54 @@ scripts_path="$(realpath ./scripts/bridge)"
...
 else
-  git clone https://github.com/ChainSafe/lodestar $lodestar_dir
-  pushd $lodestar_dir
-  git fetch && git checkout $LODESTAR_TAG
-  set_slot_time 1
-  popd
+    echo "Downloading lodestar from GitHub releases (version $LODESTAR_TAG)"
+    
+    # assuming linux since the rest of the project requires it
+    DOWNLOAD_URL="https://github.com/ChainSafe/lodestar/releases/download/$LODESTAR_TAG/lodestar-$LODESTAR_TAG-linux-amd64.tar.gz"
```
ParthDesai, line 21: I do not think there is a requirement for this. If possible, let's do this:
if it is linux, we can directly download it. 
otherwise, we can build it.

timbrinded: done

girazoki (approved):

evgeny-s (approved):

PR #901, opened 2025-03-05
New smoke test for slahes

girazoki:
Adds a new smoke test that

- ensures information about bonded eras is not stored for more than maxBOndedEras
- ensures that if there was a slash, a message was sent yielding an increase of the nonce


test/suites/smoke-test-dancelight/test-external-validator-slashes.ts
```diff
@@ -0,0 +1,91 @@
...
+        beforeAll(async () => {
+            api = context.polkadotJs();
+        });
+
+        it({
+            id: "C01",
+            title: "Number of bonded eras does not exceed bonding duration",
+            test: async () => {
+                const bondedEras = await api.query.externalValidatorSlashes.bondedEras();
+                const bondedDuration = await api.consts.externalValidatorSlashes.bondingDuration;
```
evgeny-s, line 23: I don't think we need await here, because `consts` is not a Promise.

test/suites/smoke-test-dancelight/test-external-validator-slashes.ts
```diff
@@ -0,0 +1,91 @@
...
+                }
+            },
+        });
+
+        it({
+            id: "C02",
+            title: "Slash message should increase nonce",
+            timeout: 600000,
+            test: async () => {
+                const bondedDuration = await api.consts.externalValidatorSlashes.bondingDuration;
```
evgeny-s, line 39: same here, no await needed

test/utils/block.ts
```diff
@@ -479,8 +479,9 @@ export async function isEventEmittedInTheNextBlocks(
     return false;
 }
 
+// we should always use active era
 export const getCurrentEraStartBlock = async (api: ApiPromise): Promise<number> => {
-    const currentEra = await api.query.externalValidators.currentEra();
```
evgeny-s, line 483: What is the diff between currentEra and activeEra. I see it's the same, but in different format:
```
419
```
```
{
  index: 419
  start: 1,741,172,094,000
}
```

test/suites/smoke-test-dancelight/test-external-validator-slashes.ts
girazoki: will check

test/utils/block.ts
girazoki: there is one session at which is different. but we should always look at active era. Current era = active era+1 only if the next session has an era change

test/utils/block.ts
```diff
@@ -489,10 +490,37 @@ export const getCurrentEraStartBlock = async (api: ApiPromise): Promise<number>
...
+    ) {
         epochStartBlock = (await apiAtPreviousEpochBlock.query.babe.epochStart()).toJSON()[1];
         apiAtPreviousEpochBlock = await api.at(await api.rpc.chain.getBlockHash(epochStartBlock - 1));
     }
 
     return epochStartBlock;
 };
+
+// getBlockHash is not available unless we use the current api, hence the reason to use 2
+export const getPastEraStartBlock = async (currentApi: ApiPromise, block: Number): Promise<number> => {
```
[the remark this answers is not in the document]
girazoki: why not? I have been able to succesfully run the test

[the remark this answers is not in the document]
evgeny-s: Because `number` is primitive, but `Number` is object type.
I don't think we do TS check before run check, so probably even if you put `boolean` here, it still would work.

test/utils/block.ts
```diff
@@ -489,10 +490,37 @@ export const getCurrentEraStartBlock = async (api: ApiPromise): Promise<number>
...
+        (await apiAtPreviousEpochBlock.query.externalValidators.activeEra()).unwrap().index.toNumber()
+    ) {
         epochStartBlock = (await apiAtPreviousEpochBlock.query.babe.epochStart()).toJSON()[1];
         apiAtPreviousEpochBlock = await api.at(await api.rpc.chain.getBlockHash(epochStartBlock - 1));
     }
 
     return epochStartBlock;
 };
+
+// getBlockHash is not available unless we use the current api, hence the reason to use 2
```
evgeny-s, line 504: Maybe rename the block argument or add a description that the provided block is the block in the era we are looking at the first block, to make it a bit more clear.

test/utils/block.ts
Agusrodri: Can you add a small comment for that in the function please @girazoki? It will be useful to have it there.

girazoki: yes

girazoki: fixed

test/utils/block.ts
girazoki: fixed

test/utils/block.ts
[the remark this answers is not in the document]
girazoki: fixed

test/utils/block.ts
tmpolaczyk: Rename the variable as well to avoid confusion

test/utils/block.ts
```diff
@@ -489,10 +493,39 @@ export const getCurrentEraStartBlock = async (api: ApiPromise): Promise<number>
...
     return epochStartBlock;
 };
+
+// Same as getCurrentEraStartBlock, but using and api at a certain block height
+// the block that you pass is the block from which we will get the active era
+// and from which we will get the era start block number
+export const getPastEraStartBlock = async (currentApi: ApiPromise, block: number): Promise<number> => {
+    const apiAtCheckpointForEra = await currentApi.at(await currentApi.rpc.chain.getBlockHash(block));
+
+    const currentEra = await apiAtCheckpointForEra.query.externalValidators.activeEra();
```
tmpolaczyk, line 513: The rest of the function is identical, why not just

return getCurrentEraStartBlock(apiAtCheckpointForEra)

tmpolaczyk (approved):

test/utils/block.ts
girazoki: you cant, because you call api.rpc.chain.getBlockHash, and that is only available in the current api

PR #905, opened 2025-03-06
Add StreamPayment to Dancelight and wire it to DataPreservers

nanocryk:
- Adds StreamPayment pallet to dancelight, which was only present in dancebox/flashbox
- Wire it to DataPreservers which allows data preservers to be paid using stream payments.
- Add a new integration test to check different modes of payment cannot be mixed between the profile request and the assignment.
- Copy integration tests related to StreamPayment and DataPreservers to dancelight
- Fix typos replacing "Assignement" with "Assignment"


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1638,6 +1638,157 @@ impl pallet_services_payment::Config for Runtime {
...
+        // only one asset id
+        let StreamPaymentAssetId::Native = asset_id;
+
+        Balances::set_balance(account, amount);
+    }
+}
+
+#[apply(derive_storage_traits)]
+#[derive(Copy, Serialize, Deserialize, MaxEncodedLen)]
+pub enum TimeUnit {
```
girazoki, line 1740: I am starting to think we should put this in some common folder, at least now that everybody has the same values. WDYT jeremy?

nanocryk: As per TODO, do we want in the future to support relay block number? In which case it will not work for dancelight, unless we make it `ParentBlockNumber` to refer to Ethereum (but that will need a diverging implementation)

girazoki (approved):

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
evgeny-s: We already have `client/node-common`. 
Probably a good place.

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1639,6 +1639,157 @@ impl pallet_services_payment::Config for Runtime {
...
+#[derive(Copy, Serialize, Deserialize, MaxEncodedLen)]
+pub enum StreamPaymentAssetId {
+    Native,
+}
+
+pub struct StreamPaymentAssets;
+impl pallet_stream_payment::Assets<AccountId, StreamPaymentAssetId, Balance>
+    for StreamPaymentAssets
+{
+    fn transfer_deposit(
```
evgeny-s, line 1652: I see the same in the dancebox, flashbox, we might want to consider re-using the the logic and put it to `client/node-common` probably.

chains/orchestrator-relays/runtime/dancelight/src/tests/integration_test.rs
```diff
@@ -356,6 +358,197 @@ fn test_container_deregister_unassign_data_preserver() {
         });
 }
 
+#[test]
+fn stream_payment_works() {
```
evgeny-s, line 362: Same for tests, we need to find a way reusing tests for different nodes, probably in the another PR.

evgeny-s (approved):
Several minor comments for future

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
nanocryk: Created a ticket for common code (if possible)

chains/orchestrator-relays/runtime/dancelight/src/tests/integration_test.rs
nanocryk: Created a ticket for common code (if possible)

PR #911, opened 2025-03-11
[MD-275] Command struct refactor across all nodes

evgeny-s:
**Description**
The Goal of the PR is to remove copy-pasted code across all nodes, where possible, to minimize code duplication towards re-using from the node-common crate.

**What's done**
- Extracted `BuildSpecCmd` struct and proper implementation
- Extracted `RelayChainCli` struct and proper implementation
- Extracted `Subcommand` struct and proper implementation
- Extracted `Extensions` struct


nanocryk (approved):
Good strategy!

client/node-common/src/command.rs
```diff
@@ -44,3 +51,179 @@ pub fn generate_genesis_block<Block: BlockT>(
...
+        {} <parachain-args> -- <relay-chain-args>",
+                n.borrow(),
+                RelayChainCli::executable_name()
+            )
+        })
+    }
+}
+
+thread_local! {
+    static NODE_NAME: RefCell<String> = RefCell::new("Default".to_string());
```
tmpolaczyk, line 224: This will only work if `NODE_NAME.with()` is called from the same thread that run `set_node_name`. Maybe try using some of the OnceLock / OnceCell stuff from std instead

nanocryk: Not a fan of globals like that. It would be cleaner for RelayChainCli to be generic over such constant, such as a Get<&'static str>.

tmpolaczyk: Yeah, since it's known at compile time it should be easy to turn into a generic, but if it's hard to do then I don't mind the global

evgeny-s: Will check guys. 
I was trying to find a way to override trait function implementation for the specific chain, but not sure if it's possible easily without overriding the whole trait which doesn't make sense. 
Will try to replace it with generic.

tmpolaczyk (approved):

client/node-common/src/command.rs
evgeny-s: Fixed

evgeny-s: @nanocryk, please also take a look. Thanks!

nanocryk: Way better!

nanocryk (approved):

PR #912, opened 2025-03-11
Add e2e test for token transfers

Agusrodri:
### What does it do?

Adds an end-to-end test to check the integrity of bidirectional token transfers between Starlight and Ethereum.

For the test environment to be successfully setup, the following steps were needed:

- Configure two new relayers (Starlight -> Ethereum and Ethereum -> Starlight) for the AssetHub channel.
- Use the proper values for `ASSET_HUB_PARAID` and `ASSET_HUB_AGENT_ID` according to the current Dancelight live configuration and the behavior of the Gateway contract implementation.
- Inside the test, we first wait a few sessions for the Starlight token to be properly registered on Ethereum.
- Set the AssetHub channel as the default one for token transfers via `ethereumTokenTransfers.setTokenTransferChannel`.
- Send the token from Starlight via `ethereumTokenTransfers.transferNativeToken`.
- Ensure the token was received on Ethereum by checking the `InboundMessageDispatched` event and the owner's (recipient) balance.
- Before sending the token back to Starlight we first need to approve the Gateway contract as a spender of the owner's address, via `tokenContract.approve`.
- Send the token back to Starlight through `Gateway.sendToken` function.
- Wait a few sessions for the message to be relayed.
- Check balances consistency.


girazoki (approved):
good job agus!

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -118,7 +129,7 @@ describeSuite({
 
             console.log("Contracts deployed");
 
-            const ethInfo = JSON.parse(<string>(await execCommand("./scripts/bridge/generate-eth-info.sh")).stdout);
+            ethInfo = JSON.parse(<string>(await execCommand("./scripts/bridge/generate-eth-info.sh")).stdout);
```
evgeny-s, line 132: Ideally, add the return type for the `execCommand` to avoid type-casting here:
```diff
- function execCommand(command: string, options?) {
+ function execCommand(command: string, options?): Promise<{stdout: string, stderr: string}> {
function execCommand(command: string, options?): Promise<{stdout: string, stderr: string}> {
    return new Promise((resolve, reject) => {
        exec(command, options, (error: unknown, stdout: string, stderr: string) => {
            if (error) {
                reject({ error, stdout, stderr });
            } else {
                resolve({ stdout, stderr });
            }
        });
    });
}
```

In this case, we can remove `<string>` from here and other places.

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -410,6 +421,166 @@ describeSuite({
...
+                const tokenAddress = await gatewayContract.tokenAddressOf(tokenTransferTokenId);
+                const tokenIsRegistered = await gatewayContract.isTokenRegistered(tokenAddress);
+                expect(tokenIsRegistered).to.be.true;
+
+                let tokenTransferReceived = false;
+                let tokenTransferSuccess = false;
+
+                console.log("Waiting for InboundMessageDispatched event...");
+
+                gatewayContract.on("InboundMessageDispatched", (_channelID, _nonce, messageID, success) => {
```
evgeny-s, line 514: `gatewayContract.on` returns promise, probably we want to wait here until it is finished: 
```diff
- gatewayContract.on("InboundMessageDispatched", (_channelID, _nonce, messageID, success) => {
+ await gatewayContract.on("InboundMessageDispatched", (_channelID, _nonce, messageID, success) => {

evgeny-s (approved):

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
Agusrodri: We are currently waiting for the event checking the `tokenTransferReceived` variable, which we modify once we receive the event with the exact data we expect, so I don't think await is fully necessary here, but I can add it of course.

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
Agusrodri: Done in 72378525e722659b2b907f37ae69ff5194a26200.

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
Agusrodri: Done in 72378525e722659b2b907f37ae69ff5194a26200.

PR #913, opened 2025-03-11
Use dedicated fees account for snowbridge

dimartiro:
# Description

Use a new dedicated fees account to collect fees and reward relayers instead of using the treasury account.

This account is used as part of `pallet_ethereum_token_transfers` to collect fees [when transfering native tokens](https://github.com/moondance-labs/tanssi/blob/6fe5cb287e352ae3a2671060d26abf699492668e/pallets/ethereum-token-transfers/src/lib.rs#L228-L234) and used by `RewardProcessor` to [reward the relayer](https://github.com/moondance-labs/tanssi/blob/6fe5cb287e352ae3a2671060d26abf699492668e/chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs#L406) that process the incoming messages from snowbridge

### Changes

- [x] Create new `SnowbridgeFeesAccount`
    - [x] Use it as `FeesAccount` config in `pallet_ethereum_token_transfers`
    - [x] Use it as part of `RewardProcessor` to reward relayers
- [x] Rename old `RewardThroughTreasury` to `RewardThroughFeesAccount` to reflect the above change
- [x] Test updated to check using the new account

### DIFF
--- a/test/suites/dev-tanssi-relay/balances/test_balances.ts
+++ b/test/suites/dev-tanssi-relay/balances/test_balances.ts
@@ -20,7 +20,7 @@ describeSuite({
             title: "Checking total issuance is correct on genesis",
             test: async () => {
                 const totalIssuance = (await polkadotJs.query.balances.totalIssuance()).toBigInt();
-                expect(totalIssuance).toBe(12_000_000_000_166_666_665n);
+                expect(totalIssuance).toBe(12_000_000_000_199_999_998n);
             },
         });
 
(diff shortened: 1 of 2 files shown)


girazoki (approved):
I like it!

evgeny-s (approved):

tmpolaczyk:
Did you add this to genesis? So it has existential deposit. For existing runtimes we will need to manually fund it at first

girazoki:
yes that is correct

nanocryk (approved):
LGTM, please add to genesis for ED.

tmpolaczyk (approved):

tmpolaczyk:
Yes

Agusrodri:
> @Agusrodri I had to update some tests, one made by you, could you take a look please? I think there might be a rounding issue but I want to double-check with you

Yes it's a rounding thing. In my opinion it's safe to ignore it.

PR #914, opened 2025-03-12
Update StreamPayment storage deposit

nanocryk:
Updates the stream payment deposit. Computed by making a stream with a pending request (worst-case size) and calling `Encode::encoded_size` on it.


tmpolaczyk:
Could you add a test with the worst case for reference? So we document what's the worst case, and also we have an example on how to calculate the deposit size. Don't need to add it to every runtime, starlight is enough.

evgeny-s (approved):

girazoki (commented):
No test breaks? If so we are not testing this properly, can we add one?

girazoki:
Ok I see some typescript tests break, we should add one for dancelight too!

tmpolaczyk (approved):

PR #916, opened 2025-03-12
Use middleware v2 in e2e test

ParthDesai:
# Description
This PR updates tanssi symbiotic repo commit and e2e test to work with middleware v2.

### DIFF
--- a/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
+++ b/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
@@ -58,6 +58,7 @@ describeSuite({
         let middlewareContract: ethers.Contract;
         let gatewayContract: ethers.Contract;
         let gatewayProxyAddress: string;
+        let middlewareAddress: string;
         let middlewareDetails: any;
 
         let operatorAccount: KeyringPair;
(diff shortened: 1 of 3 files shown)


girazoki (approved):

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -126,8 +127,9 @@ describeSuite({
             console.log("Gateway contract proxy address is:", ethInfo.snowbridge_info.contracts.GatewayProxy.address);
             gatewayProxyAddress = ethInfo.snowbridge_info.contracts.GatewayProxy.address;
 
-            console.log("Symbiotic middleware address is: ", ethInfo.symbiotic_info.contracts.Middleware.address);
+            console.log("Symbiotic middleware address is: ", ethInfo.symbiotic_info.contracts.ERC1967Proxy.address);
```
NindoK, line 130: All looks fine. Just for future reference, we are adding multiple proxies in our repo for upgradeability purposes. I'm just thinking that probably we should create a named proxy such that there could be no margin of error? What you think? Or since it's an e2e test we can "oversee" this?

NindoK: Ok will make a PR later on where I upgrade a bunch of stuff for middleware v2 and add this

PR #917, opened 2025-03-13
[MD-952] The storage deposit for container-chain registration is insufficient 

evgeny-s:
[Ticket](https://opslayer.atlassian.net/browse/MD-952) 

**Description**:
Instead of fixed deposit amount for the container chain register operation, calculate it dynamically based on the genesis data size

**What is done**:
- Updated the hold deposit calculation, based on the genesis size instead of fixed.
- Fixed tests according to new math
- The price is [1 CENT per byte](https://github.com/moondance-labs/tanssi/blob/3b945213eb2353a0389fa721da4d2f66e3b68ff7/chains/orchestrator-relays/runtime/dancelight/src/lib.rs#L610).
- Removed previously used const for the hold: `DepositAmount`
- Updated benchmarks for the registrar pallet, because old parameter used there, re-generated weights

**TODO**: 

- [ ] Once the [dancekit PR](https://github.com/moondance-labs/dancekit/pull/37) is approved, update toml with new commit hash


chains/orchestrator-paras/runtime/dancebox/src/lib.rs
```diff
@@ -286,6 +286,7 @@ pub const DAYS: BlockNumber = HOURS * 24;
 
 // Unit = the base number of indivisible units for balances
 pub const UNIT: Balance = 1_000_000_000_000;
+pub const CENTS: Balance = UNIT / 30_000;
```
nanocryk, line 289: What is this unit? Divided by `30_000`? Can we stay metric-ish in our units? :p

girazoki: haha it comes from rococo, it's ok we can leave it like this or if we modify it we should do it as a separqate pr

girazoki (approved):
LGTM

Agusrodri (approved):

PR #921, opened 2025-03-14
Fix external validators smoke test SMOK05

ParthDesai:
# Description
This PR fixes external validator smoke test SMOK05 by checking first if there were any external validators for `externalValidatorsAtCheckpointB` and only if there were it will do the assertion that there should be some rewards.


test/utils/block.ts
```diff
@@ -529,3 +529,59 @@ export const getPastEraStartBlock = async (currentApi: ApiPromise, block: number
...
     return epochStartBlock;
 };
+
+export const getEraIndexForBlock = async (api, blockNumber) => {
+    const apiAtBlock = await api.at(await api.rpc.chain.getBlockHash(blockNumber));
+    const eraAtBlock = await apiAtBlock.query.externalValidators.activeEra();
+    return eraAtBlock.toJSON().index;
+};
+
+export const findEraBlockUsingBinarySearch = async (api, eraIndex) => {
```
evgeny-s, line 539:
```suggestion
export const findEraBlockUsingBinarySearch = async (api: ApiPromise, eraIndex: number): Promise<{found: boolean, blockHash: string}> => {
```

ParthDesai: *Applied*

test/suites/smoke-test-dancelight/test-external-validators-rewards.ts
```diff
@@ -3,13 +3,18 @@ import "@tanssi/api-augment/dancelight";
...
+    getAccountBalance,
+    getCurrentEraStartBlock,
     HOLESKY_SOVEREIGN_ACCOUNT_ADDRESS,
     PRIMARY_GOVERNANCE_CHANNEL_ID,
     SEPOLIA_SOVEREIGN_ACCOUNT_ADDRESS,
-    getAccountBalance,
-    getCurrentEraStartBlock,
 } from "utils";
 
+async function calculateBlockHashAtStartOfAnEra(api, blocksPerSession, eraIndex) {
```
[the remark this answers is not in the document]
ParthDesai: We are doing that now.

PR #899, opened 2025-03-04
feat: inactivity tracking pallet (chain inactivity tracking)

chexware:
This PR adds chain inactivity tracking to `pallet-inactivity-tracking`

TO DO:
- [x] Fix and refactor unit tests
- [x] Add manual disabling container chains from block production RPC call to Dancelight  in dev mode
- [x] Prevent collators assigned to parathreads to be marked as inactive
- [x] Add unit tests covering parathread-related logic
- [x] Add integration tests covering parathread-related edge cases

PR #922, opened 2025-03-14
feat: inactivity tracking pallet (collators tracking)

chexware:
This PR adds `pallet-inactivity-tracking` for tracking inactive nodes and container chains:

TO DO:

- [x] Tracking inactive collators 
- [x] Trait that allows other pallets to verfiy if node is inactive
- [x] Tests

Dependent on:
- support for tracking inactive chains (https://github.com/moondance-labs/tanssi/pull/899)


pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+            {
+                total_weight += Self::on_author_noted(orchestrator_chain_author);
+            }
+
+            // Update inactive collator records only after a session has ended
+            let current_session = T::CurrentSessionIndex::session_index();
+            let current_unprocessed_session = <LastUnprocessedSession<T>>::get();
+            if current_unprocessed_session < current_session {
+                total_weight +=
+                    Self::process_ended_session(current_unprocessed_session, current_session);
```
girazoki, line 155: We said this could be done through pallet-intiliazer right?

girazoki: and indeed this does not work in my opinion, you should see it in the tests. Here you will enter when the new session starts, which means that your `ActiveCollatorsForCurrentSession` will be of those when the session changes, and not before. you need to do this BEFORE the next session starts. so either you predict that is going to change or do it through pallet-initializer

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+                let _ = <crate::pallet::ActiveCollators<T>>::remove(
+                    active_session_id - minimum_sessions_required,
+                );
+            }
+            <LastUnprocessedSession<T>>::put(active_session_id);
+            total_weight
+        }
+        pub fn on_author_noted(author: T::CollatorId) -> Weight {
+            let mut total_weight = T::DbWeight::get().reads_writes(1, 0);
+            let _ = <ActiveCollatorsForCurrentSession<T>>::try_mutate(
```
girazoki, line 187: maybe can be inserted ordered so that the lookup takes less time? @ParthDesai and @tmpolaczyk

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+                },
+            );
+            total_weight
+        }
+    }
+}
+
+impl<T: Config> NodeActivityTrackingHelper<T::CollatorId> for Pallet<T> {
+    fn is_node_inactive(node: &T::CollatorId) -> bool {
+        if !<EnableInactivityTracking<T>>::get() {
```
girazoki, line 205: comment why

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+            return false;
+        }
+
+        let current_session = T::CurrentSessionIndex::session_index();
+
+        let minimum_sessions_required = T::MaxInactiveSessions::get();
+        if current_session < minimum_sessions_required {
```
girazoki, line 212: nice, test for this

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+            return false;
+        }
+
+        for session_index in current_session.saturating_sub(T::MaxInactiveSessions::get())
+            ..current_session.saturating_sub(1u32)
+        {
+            if <ActiveCollators<T>>::get(session_index).contains(node) {
+                return false;
```
girazoki, line 220: is there a more "RUST" way to do this @tmpolaczyk @nanocryk ?

typescript-api/scripts/runtime-upgrade.ts
```diff
@@ -36,7 +36,7 @@ src/dancelight/interfaces/types-lookup.ts
 function hackXcmV5Support() {
     // For dancebox, replace "Lookup89" with "StagingXcmV5Junction"
     const danceboxFilePath = "src/dancebox/interfaces/types-lookup.ts";
-    hackTypeReplacement(danceboxFilePath, "Lookup89", "StagingXcmV5Junction", 8);
+    hackTypeReplacement(danceboxFilePath, "Lookup90", "StagingXcmV5Junction", 8);
```
girazoki, line 39: why is this needed?

girazoki (commented):
Overall PR looks good, obviously there is a lot of testing still pending but the structure is good. minor comments

typescript-api/scripts/runtime-upgrade.ts
chexware: Otherwise typescript types generation returns an error

pallets/inactivity-tracking/src/lib.rs
chexware: Yup plan to refactor this one

typescript-api/scripts/runtime-upgrade.ts
tmpolaczyk: Apparently "Lookup89" means this is the 89th type, but this PR adds a new type that happens to be before that, so now the xcm junction is the 90th type...

pallets/inactivity-tracking/src/lib.rs
tmpolaczyk: Using BoundedBTreeSet instead of BoundedVec will automatically ensure it's ordered and speed up lookups

pallets/inactivity-tracking/src/lib.rs
tmpolaczyk: Like this but it's ugly, for loop is fine

```rust
(current_session.saturating_sub(T::MaxInactiveSessions::get())..current_session.saturating_sub(1u32))
    .all(|session_index| !ActiveCollators::<T>::get(session_index).contains(node))
```

pallets/inactivity-tracking/src/lib.rs
chexware: Test for this added in [24ce3ee](https://github.com/moondance-labs/tanssi/pull/922/commits/24ce3ee253e57b05a8f7825232b430723d24c5bd)

pallets/inactivity-tracking/src/lib.rs
chexware: Comment added in [378edb3](https://github.com/moondance-labs/tanssi/pull/922/commits/378edb3d7c2b0c76e2de4991de338a4beb899ebe)

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    impl<T: Config> Pallet<T> {
+        fn process_ended_session(
+            unprocessed_session_id: SessionIndex,
+            active_session_id: SessionIndex,
+        ) -> Weight {
+            let mut total_weight = T::DbWeight::get().reads_writes(1, 3);
+            ActiveCollators::<T>::insert(
+                unprocessed_session_id,
+                <ActiveCollatorsForCurrentSession<T>>::get(),
```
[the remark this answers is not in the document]
chexware: Comment added in [378edb3](https://github.com/moondance-labs/tanssi/pull/922/commits/378edb3d7c2b0c76e2de4991de338a4beb899ebe)

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        let current_session = T::CurrentSessionIndex::session_index();
+
+        let minimum_sessions_required = T::MaxInactiveSessions::get();
+        if current_session < minimum_sessions_required {
+            return false;
+        }
+
+        for session_index in current_session.saturating_sub(T::MaxInactiveSessions::get())
```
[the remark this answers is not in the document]
chexware: Refactored the loop in https://github.com/moondance-labs/tanssi/pull/922/commits/7a20d2233322359f9d884c934fda6222c0eaa06f

pallets/inactivity-tracking/src/lib.rs
nanocryk: Loop is fine

pallets/inactivity-tracking/src/lib.rs
chexware: Good point! Updated it in https://github.com/moondance-labs/tanssi/pull/922/commits/86f1ce1ef64412ec230fdb25c0ce1fa4e260578d

pallets/inactivity-tracking/src/lib.rs
chexware: Refactored the function in https://github.com/moondance-labs/tanssi/pull/922/commits/366077535ba21025f2eab141774aa71fcbd648b6 and added it to `apply_new_session` in https://github.com/moondance-labs/tanssi/pull/922/commits/213e28b88a5dccb8988077aa160c6f0dc57fb578

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    #[pallet::error]
+    pub enum Error<T> {
+        MaxCollatorsPerSessionReached,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::DbWeight::get().reads_writes(0, 1))]
+        pub fn set_inactivity_tracking_status(
```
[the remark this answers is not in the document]
chexware: `set_inactivity_tracking_status` clears `ActiveCollatorsForCurrentSession` if disabled in https://github.com/moondance-labs/tanssi/pull/922/commits/aaa5bda3f28766c816db56f4981280328d474e0d and prevents adding new activity records in https://github.com/moondance-labs/tanssi/pull/922/commits/e29fc84f9630839af152aec19d0e1376a56e39aa . Also added suspension period when enabling/disabling activity tracking  in https://github.com/moondance-labs/tanssi/pull/922/commits/c4d331409f2bf034b697a685d2856940c3a6a264 and delay of enabling activity record until the next full session in https://github.com/moondance-labs/tanssi/pull/922/commits/5f398f77f179f8b65be577912ed4a3e202872626 to prevent switching between states and messing up activity records for a session.

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+        /// The weight information of this pallet.
+        type WeightInfo: weights::WeightInfo;
+    }
+
+    /// Switch to enable/disable inactivity tracking
+    #[pallet::storage]
+    pub type CurrentActivityTrackingStatus<T: Config> =
+        StorageValue<_, ActivityTrackingStatus, ValueQuery>;
+
+    /// A list of double map of inactive collators for a session
```
Agusrodri, line 126: Here the comment says "inactive" while the actual storage name is `ActiveCollators`. I believe we should unify concepts to avoid confusions (or expand a bit more in the comment).

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+    #[pallet::storage]
+    pub type ActiveCollators<T: Config> = StorageMap<
+        _,
+        Twox64Concat,
+        SessionIndex,
+        BoundedBTreeSet<T::CollatorId, T::MaxCollatorsPerSession>,
+        ValueQuery,
+    >;
+
+    /// A list of inactive collators for a session. Repopulated at the start of every session
```
Agusrodri, line 136: Same here.

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+    #[pallet::storage]
+    pub type ActiveCollatorsForCurrentSession<T: Config> =
+        StorageValue<_, BoundedBTreeSet<T::CollatorId, T::MaxCollatorsPerSession>, ValueQuery>;
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        ActivityTrackingStatusSet { status: ActivityTrackingStatus },
```
Agusrodri, line 144: I would suggest adding a short comment line explaining the event.

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        MaxCollatorsPerSessionReached,
```
Agusrodri, line 149: Also a small explanation line for each error :)

pallets/inactivity-tracking/src/lib.rs
chexware: Added comment in https://github.com/moondance-labs/tanssi/pull/922/commits/11e8de05240124c45ebb91bc99a958d3441fe140

pallets/inactivity-tracking/src/lib.rs
chexware: Good point! Updated in https://github.com/moondance-labs/tanssi/pull/922/commits/11e8de05240124c45ebb91bc99a958d3441fe140

pallets/inactivity-tracking/src/lib.rs
chexware: Updated in https://github.com/moondance-labs/tanssi/pull/922/commits/11e8de05240124c45ebb91bc99a958d3441fe140

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+// (at your option) any later version.
+
+// Tanssi is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
+#![cfg_attr(not(feature = "std"), no_std)]
```
girazoki, line 16: Small doc about what the pallet does

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                ActivityTrackingStatus::Enabled { start: _, end } => end,
+                ActivityTrackingStatus::Disabled { end } => end,
+            };
+            let current_session_index = T::CurrentSessionIndex::session_index();
+            ensure!(
+                current_session_index > current_status_end_session_index,
+                Error::<T>::ActivityStatusUpdateSuspended
+            );
+            let new_status_end_session_index =
+                current_session_index + T::MaxInactiveSessions::get();
```
girazoki, line 175: use saturating add or checked add

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::set_inactivity_tracking_status())]
+        pub fn set_inactivity_tracking_status(
+            origin: OriginFor<T>,
+            is_enabled: bool,
+        ) -> DispatchResult {
+            ensure_root(origin)?;
+            let current_status_end_session_index = match <CurrentActivityTrackingStatus<T>>::get() {
+                ActivityTrackingStatus::Enabled { start: _, end } => end,
+                ActivityTrackingStatus::Disabled { end } => end,
```
girazoki, line 167: I am guessing if this is disabled, there is no reason to re-disable it again no? so maybe we should early return in that case?

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            <CurrentActivityTrackingStatus<T>>::put(new_status.clone());
+            Self::deposit_event(Event::<T>::ActivityTrackingStatusSet { status: new_status });
+            Ok(())
+        }
+    }
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: BlockNumberFor<T>) -> Weight {
+            let mut total_weight = T::DbWeight::get().reads_writes(1, 0);
```
girazoki, line 196: I count 3 reads at least (getSelfChain author, activity ttracking status, and session index

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                current_session_index.saturating_sub(1),
+                <ActiveCollatorsForCurrentSession<T>>::get(),
+            );
+
+            <ActiveCollatorsForCurrentSession<T>>::put(BoundedBTreeSet::new());
+
+            // Cleanup active collator info for sessions that are older than the maximum allowed
+            if current_session_index > T::MaxInactiveSessions::get() {
+                <crate::pallet::ActiveCollators<T>>::remove(
+                    current_session_index - T::MaxInactiveSessions::get() - 1,
```
girazoki, line 230: saturating, just in case

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                );
+            }
+        }
+        pub fn on_author_noted(author: T::CollatorId) -> Weight {
+            let mut total_weight = T::DbWeight::get().reads_writes(1, 0);
+            let _ = <ActiveCollatorsForCurrentSession<T>>::try_mutate(
+                |active_collators| -> DispatchResult {
+                    if active_collators
+                        .try_insert(author)
```
girazoki, line 239: @tmpolaczyk why does try_insert work? according to docs is a nightly feature only: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.try_insert

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                        total_weight += Self::on_author_noted(orchestrator_chain_author);
+                    }
+                }
+            }
+            total_weight
+        }
+    }
+
+    impl<T: Config> Pallet<T> {
+        pub fn process_ended_session() {
```
girazoki, line 214: I still think we should separate this into:

- things that should execute when session ends through the trait (store the current active collators).
- things that should execute when the session starts (put an empty btree map)

This would avoid having that saturating.sub which is rather ugly. But we can leave this for another PR/issue. can you create one for this?¿

chains/orchestrator-paras/runtime/dancebox/src/lib.rs
```diff
@@ -1729,6 +1737,27 @@ impl pallet_multisig::Config for Runtime {
...
+
+impl pallet_inactivity_tracking::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type CollatorId = CollatorId;
+    type MaxInactiveSessions = ConstU32<5>;
+    type MaxCollatorsPerSession = ConstU32<100>;
+    #[cfg(not(feature = "runtime-benchmarks"))]
+    type CurrentSessionIndex = CurrentSessionIndexGetter;
+    #[cfg(feature = "runtime-benchmarks")]
+    type CurrentSessionIndex = MockCurrentSessionGetter;
```
girazoki, line 1756: this should not be needed for benchmarks I think. Why did we need a mockCurrentSessionGetter?

girazoki: in theory, everything should work without the mocker.

chains/orchestrator-paras/runtime/dancebox/src/tests/inactivity_tracking.rs
```diff
@@ -0,0 +1,209 @@
...
+        .with_collators(vec![
+            (AccountId::from(ALICE), 100_000),
+            (AccountId::from(BOB), 100_000),
+            (AccountId::from(CHARLIE), 100_000),
+            (AccountId::from(DAVE), 100_000),
+        ])
+        .build()
+        .execute_with(|| {
+            run_block();
+            note_block_authors(vec![(CHARLIE.into(), 3001.into())]);
```
girazoki, line 61: I would separate this into 2 tests:

- one where we test the author noting for the orchestrator chain. that one should be rather easy, as every run_block() we simply inject an author log

- one where we test the author noting for the container-chains. In this case, we need to use something similar to https://github.com/moondance-labs/tanssi/blob/04b18155ddb9dd85b396196a72299dfcbedfe1df/chains/orchestrator-paras/runtime/dancebox/src/tests/integration_test.rs#L4838 to inject the log. This are supposed to be integration tests, ergo we should try to go through the real path instead of directly calling it like you are doing in `note_block_authors`. we can discuss these further if you need help

pallets/inactivity-tracking/src/lib.rs
tmpolaczyk: https://paritytech.github.io/polkadot-sdk/master/sp_core/bounded_btree_map/struct.BoundedBTreeMap.html#method.try_insert

girazoki: thansk!

pallets/inactivity-tracking/src/lib.rs
chexware: Yeah fair point. I extended `pallet-initializer`'s SessionHandler trait in https://github.com/moondance-labs/tanssi/pull/899 to also execute `on_before_session_ending`. Perhaps can updating the `ActiveCollators` for the session in that PR too?

pallets/inactivity-tracking/src/lib.rs
chexware: Would it make sense to increment the  `total_weight ` before each if instead?

pallets/inactivity-tracking/src/lib.rs
chexware: Added in https://github.com/moondance-labs/tanssi/pull/922/commits/0b111609fd8b3db997d6b05f416c833732c95de8

pallets/inactivity-tracking/src/lib.rs
chexware: Very good point! Also re-enabling after being enabled will stop recording activity for the current session. Added check that prevents that from happening + test in https://github.com/moondance-labs/tanssi/pull/922/commits/dfbcf46af9922f42291558f55de584f1419cc270

chains/orchestrator-paras/runtime/dancebox/src/lib.rs
chexware: Removed the mocker in https://github.com/moondance-labs/tanssi/pull/922/commits/2d880c174d44e0c3a0fab082e79e73d7bda818fe

pallets/inactivity-tracking/src/lib.rs
girazoki: yes makes sense

pallets/inactivity-tracking/src/lib.rs
girazoki: let's leave it for another PR, just create an issue to move it there, I dont want to make this bigger

chains/orchestrator-paras/runtime/dancebox/src/tests/inactivity_tracking.rs
chexware: I split the dancebox  inactivity tracking test into 2 tests in https://github.com/moondance-labs/tanssi/pull/922/commits/47da736d18bcabdf3f15647d89dc5ad7bda6d36d and replaced `note_block_authors `with `ParaHeaderSproofBuilder` in https://github.com/moondance-labs/tanssi/pull/922/commits/2a829cf9b759dcecccb15f98c65ff08b41eeb7e0

test/suites/dev-tanssi/inactivity-tracking/test-inactivity-tracking.ts
```diff
@@ -0,0 +1,56 @@
...
+                    await polkadotJs.query.inactivityTracking.activeCollators(2);
+                expect(activeCollatorsRecordBeforeActivityWindow.isEmpty).to.be.true;
+                await jumpToSession(context, 3);
+                const activeCollatorsRecordWithinActivityWindow =
+                    await polkadotJs.query.inactivityTracking.activeCollators(2);
+                expect(activeCollatorsRecordWithinActivityWindow.size).to.be.equal(3);
+
+                await jumpToSession(context, maxInactiveSessions + 3);
+                const activeCollatorsRecordAfterActivityWindow =
+                    await polkadotJs.query.inactivityTracking.activeCollators(2);
```
girazoki, line 51: do it based on maxINactiveSesssions, something like currentSession - maxInactiveSessions -1

girazoki: also put comments in the test about what you are testing. something like //activity should have been removed for history farther than `maxInactiveSesssions`

girazoki (approved):
request for additional tests, but overall looks good. good job @chexware !

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+                );
+            }
+        }
+        pub fn on_author_noted(author: T::CollatorId) -> Weight {
+            let mut total_weight = T::DbWeight::get().reads_writes(1, 0);
+            let _ = <ActiveCollatorsForCurrentSession<T>>::try_mutate(
+                |active_collators| -> DispatchResult {
+                    if active_collators
+                        .try_insert(author)
+                        .map_err(|_| Error::<T>::MaxCollatorsPerSessionReached)?
```
tmpolaczyk, line 261: This error is being ignored by the `let _` above, add some logging here so we at least see the log if it happens. Maybe rewrite `let _ =` as `if let Err(e) = ` and log there

tmpolaczyk: Actually this is important, if the storage is full it means that we cannot mark any more collators as active, so there will be some collators that produce blocks and will not be included, and will be considered inactive.

Since this should never happen, how about disabling activity tracking in that case? My point is that if the storage is full it means there is a bug in the pallet, so it should be disabled.

Any other ideas @girazoki @chexware ?

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+                    Ok(())
+                },
+            );
+            total_weight
+        }
+    }
+}
+
+impl<T: Config> NodeActivityTrackingHelper<T::CollatorId> for Pallet<T> {
+    fn is_node_inactive(node: &T::CollatorId) -> bool {
```
tmpolaczyk, line 274: Depending on how this function is called it may return invalid results. For example, if a new collator joins, it has not produced any blocks in the past sessions, so this function will say it is inactive.

So I guess the only way for this to work is to keep track of which collators have been collators for more than 5 sessions, and only allow checking activity for those. Am I right or am I missing something?

pallets/inactivity-tracking/src/lib.rs
girazoki: yes I think that is a good idea. as you said this should never happen, but in the odd event that it does, it's better to disable it directly yes. 

Add a test for it!

pallets/inactivity-tracking/src/lib.rs
girazoki: No there is something we can do, which is count non-activity instead of activity. that way you can clearly distinguish between a collator not joining and one being inactive

girazoki: @chexware revisit tomasz suggestion, and if it's true, the let's chain the code to count whether a collator was `offline` for a given session (I think for that you just need to change the `process_ended_session` function)

chexware: @chexware Agree that should prevent marking newly  joined collators as inactive.

pallets/inactivity-tracking/src/lib.rs
chexware: Added overflow handling and test for it in https://github.com/moondance-labs/tanssi/pull/922/commits/82458129f6464a042fcc1a93b8eeac7c805d8435

test/suites/dev-tanssi-relay/inactivity-tracking/test-inactivity-tracking.ts
```diff
@@ -0,0 +1,112 @@
+import "@tanssi/api-augment";
+import { type DevModeContext, beforeAll, describeSuite, expect } from "@moonwall/cli";
+import type { KeyringPair } from "@moonwall/util";
+import type { ApiPromise } from "@polkadot/api";
+import type { Digest, DigestItem, HeadData, Header, ParaId, Slot } from "@polkadot/types/interfaces";
+import { jumpToSession } from "utils";
+import { stringToHex } from "@polkadot/util";
+
+async function mockAndInsertHeadData(
```
Agusrodri, line 9: This function is already present in `test/utils/block.ts`, you can use it directly from there.

chains/orchestrator-paras/runtime/flashbox/src/lib.rs
```diff
@@ -568,6 +568,7 @@ impl pallet_initializer::ApplyNewSession<Runtime> for OwnApplySession {
             &assignments.next_assignment,
         );
     }
+    fn on_before_session_ending() {}
```
Agusrodri, line 571: I guess we are not adding the inactivity tracking pallet to flashbox on purpose right?

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        pub fn on_before_session_ending() {
+            ActiveCollators::<T>::insert(
+                T::CurrentSessionIndex::session_index(),
+                <ActiveCollatorsForCurrentSession<T>>::get(),
+            );
+        }
+
+        pub fn on_author_noted(author: T::CollatorId) -> Weight {
```
Agusrodri, line 261: A couple lines of docs for these helper functions would be great :)

Agusrodri:
Very nice job @chexware!! Could you please add a more extensive description of the PR when you have some time? 

It would be great to cover in a more detailed way some edge cases on the explanation, as well as the main purpose of the pallet, its hooks, behaviors, etc.

This way we can have this (huge) functionality documented for the future in the original PR.

Thanks! :)

chains/orchestrator-paras/runtime/flashbox/src/lib.rs
chexware: Yes. Since we don't have `pallet-pooled-staking`  enabled for flashbox only part of the flow will be available without it.

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+        ActivityTrackingStatusAlreadySet,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight(T::WeightInfo::set_inactivity_tracking_status())]
+        pub fn set_inactivity_tracking_status(
+            origin: OriginFor<T>,
+            is_enabled: bool,
```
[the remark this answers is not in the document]
chexware: Good point! Updated in https://github.com/moondance-labs/tanssi/pull/922/commits/2eab1dff0fc47085f4bab39982cdb469331adbdb

pallets/inactivity-tracking/src/lib.rs
chexware: Added docs in https://github.com/moondance-labs/tanssi/pull/922/commits/2eab1dff0fc47085f4bab39982cdb469331adbdb

pallets/inactivity-tracking/src/lib.rs
chexware: Using inactivity tracking from https://github.com/moondance-labs/tanssi/pull/922/commits/dc58e94a1504539385be07e11cb57d075cef1955 . However as a result the integration tests are braking

chexware: Tests related issues fixed in https://github.com/moondance-labs/tanssi/pull/922/commits/22d672ccf9180a94d92e11cadf8a9965675ce75e and https://github.com/moondance-labs/tanssi/pull/922/commits/7c91fc87fd8f6c64b508d37ee5bca2e61fbad624

test/suites/dev-tanssi/inactivity-tracking/test-inactivity-tracking.ts
```diff
@@ -0,0 +1,56 @@
...
+        it({
+            id: "E01",
+            title: "Pallet should correctly update collators' activity records",
+            test: async () => {
+                const maxInactiveSessions = polkadotJs.consts.inactivityTracking.maxInactiveSessions.toNumber();
+                jumpToSession(context, 2);
+                const activeCollatorsForSessionBeforeNoting =
+                    await polkadotJs.query.inactivityTracking.activeCollatorsForCurrentSession();
+                expect(activeCollatorsForSessionBeforeNoting.size).to.be.equal(1);
+
```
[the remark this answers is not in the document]
chexware: @girazoki Added changes that successfully add collator to 2001 in https://github.com/moondance-labs/tanssi/pull/922/commits/a9d5676bba674aaf53d73089ef3089c94c73f9ae. However, 2001 seems to produces blocks.  Tomash and I found  way to make only one collator per chain producing block by modifying the `slot_number ` from https://github.com/moondance-labs/tanssi/blob/14826b25d6d75eab29c75979610c1d01953750da/primitives/author-noting-inherent/src/mock.rs#L148 to` 2 * u64::from(self.slots_per_para_block) * u64::from(self.current_para_block)` but that will break other tests.

pallets/inactivity-tracking/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+        RuntimeDebug,
+        MaxEncodedLen,
+    )]
+    pub enum ActivityTrackingStatus {
+        Enabled {
+            start: SessionIndex,
+            end: SessionIndex,
+        },
+        Disabled {
+            end: SessionIndex,
```
tmpolaczyk, line 88: I don't understand what are start and end, can you add doc comments?

pallets/inactivity-tracking/src/tests.rs
```diff
@@ -0,0 +1,573 @@
...
+            ActiveCollatorsForCurrentSession::<Test>::get(),
+            current_session_active_collator_record
+        );
+        assert_eq!(InactiveCollators::<Test>::get(0), empty_vec);
+
+        roll_to(SESSION_BLOCK_LENGTH - 1);
+        Pallet::<Test>::on_before_session_ending();
+        roll_to(SESSION_BLOCK_LENGTH);
+        Pallet::<Test>::process_ended_session();
+
```
tmpolaczyk, line 390: Having to call `on_before_session_ending` and `process_ended_session` manually in each test is very error prone. Why don't we integrate it into `roll_to`?

I see another pallet uses

```
impl pallet_session::SessionHandler<u64> for TestSessionHandler {
```

not sure what's the difference between that and what you have here, if you can use that it's probably better

pallets/inactivity-tracking/src/lib.rs
chexware: Added doc comments explaining each  status parameter value means in https://github.com/moondance-labs/tanssi/pull/922/commits/19d28a1b6304312438cbbe172f00917c85f8e01e

pallets/inactivity-tracking/src/tests.rs
chexware: For sure. Added the setup in https://github.com/moondance-labs/tanssi/pull/922/commits/00b45e565658b359c373ebc659269eb8f88c8ecf

tmpolaczyk (approved):
Looks good!

I didn't review the tests but I wrote my own and didn't find any bugs. We will probably be looking at this pallet again in the following PRs, so any improvements can be done later.

test/suites/dev-tanssi/inactivity-tracking/test-inactivity-tracking.ts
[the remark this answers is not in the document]
chexware: Added manual disabling of chains from block production in https://github.com/moondance-labs/tanssi/pull/922/commits/3694cc71b5f28110a2b133df890fa786ee45d466, https://github.com/moondance-labs/tanssi/pull/922/commits/d885c3ee03010b8c0c7e91984cd758c08e7fcba6, https://github.com/moondance-labs/tanssi/pull/922/commits/bad497ecc40befed5aa82fd66218179d1652fded, and https://github.com/moondance-labs/tanssi/pull/922/commits/3bda034b482f43a91a0acd01df2f3c12f1669fd8

chains/orchestrator-paras/node/src/rpc.rs
```diff
@@ -58,6 +61,8 @@ pub struct FullDeps<C, P> {
     pub xcm_senders: Option<(flume::Sender<Vec<u8>>, flume::Sender<(ParaId, Vec<u8>)>)>,
     /// Channels for manually activating the randomness
     pub randomness_sender: Option<flume::Sender<(bool, Option<[u8; 32]>)>>,
+    /// Channels for manually excluding container chains from producing blocks
+    pub container_chain_exclusion_sender: Option<flume::Sender<Vec<ParaId>>>,
```
girazoki, line 65: let's create an issue for adding this also for dancelight

girazoki (approved):
looks good! good job on the manual seal stuff!

chains/orchestrator-paras/node/src/rpc.rs
chexware: Yeah I can add it in the chain inactivity PR too!

PR #925, opened 2025-03-17
Test native token transfer fee pricing parameters

dimartiro:
# Description

The `snowbridge_pallet_system::PricingParameters` are use to calculate the fee that the user will pay when sending native tokens to etheruem.

This test is to check that the native token transfer fee is modified when the pricing parameters are updated.

plus: Update e2e test to check fee changes using the new fees account

### DIFF
--- a/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
+++ b/test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
@@ -724,8 +724,10 @@ describeSuite({
 
                 // Reward is reduced from fees account
                 const feesAccountBalanceAfterReceiving = (await relayApi.query.system.account(SNOWBRIDGE_FEES_ACCOUNT))
-                .data.free;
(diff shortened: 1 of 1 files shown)


chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -719,6 +719,115 @@ fn can_process_message_returns_false_for_wrong_message_type() {
...
+                ) = &r.event
+                {
+                    second_fee_found = *fee;
+                }
+            });
+
+            // Check the relation between two fees is the pricing parameters multiplier
+            assert!(
+                second_fee_found.div_ceil(first_fee_found) as f64
+                    == pricing_parameters.multiplier.to_float()
```
Agusrodri, line 821: Mmm not sure if casting to float here is the best. Can't we do these operations with `FixedU128` type?

tmpolaczyk: +1, also clippy doesn't like when you compare floats using ==

girazoki (approved):

Agusrodri (approved):

PR #928, opened 2025-03-18
Disable block history download after warp sync in container collator

tmpolaczyk:
See also commit in moonkit: https://github.com/moondance-labs/moonkit/commit/4742f8fb1a5efc5d165f7cb4137ee2cbdcb2969b

Disables the block history download after warp sync for container chain collators. This was not needed but polkadot forces it as part of warp sync. Old behavior can be restored using a CLI flag `--download-block-history` in case this breaks something. Anything other than a container chain collator (relay, full nodes, data preservers, etc) will still download the block history if using warp sync, as polkadot intented.

And add zombienet test that ensures that warp sync does not download the block history


test/suites/zombie_tanssi_warp_sync/test_zombie_tanssi_warp_sync.ts
```diff
@@ -284,6 +284,35 @@ describeSuite({
...
+TRACE tokio-runtime-worker sync: [Container-2000] 7 blocks ready for import
+DEBUG tokio-runtime-worker sync: [Container-2000] Drained 7 gap blocks from 0
+TRACE tokio-runtime-worker sync: [Container-2000] Accepted 7 blocks (0xf634a6692aef17af72ccc9ae716d991a3bd8c8fa2c52218eae441d3097c8a503) with origin NetworkInitialSync
+TRACE tokio-runtime-worker sync::import-queue: [Container-2000] Scheduling 7 blocks for import
+                 */
+                // Can't check the other logs because we also see them during normal sync:
+                // When the collator warp syncs to block 8, in the meantime the latest block is 11,
+                // so it syncs blocks 9-11 printing almost the same logs as the gap sync.
+                await checkLogsNotExist(logFilePath, [
+                    "Starting gap sync",
```
tmpolaczyk, line 312: Here if we enable trace logs we could check if we import blocks 1, 2, 3, etc, but that may be too much. I didn't find a way to enable only this one log that is printed on each imported block

test/configs/zombie_tanssi_warp_sync.json
```diff
@@ -72,7 +72,9 @@
                 {
                     "name": "Collator1000-03",
                     "command": "../target/release/tanssi-node",
-                    "args": ["--no-hardware-benchmarks", "--database=paritydb", "--wasmtime-precompiled=wasm"]
+                    "args": ["--no-hardware-benchmarks", "--database=paritydb", "--wasmtime-precompiled=wasm", "-ldebug,sync=trace",
+                        "-- --rpc-port 9952 --rpc-cors all --unsafe-rpc-external --rpc-methods unsafe"
```
girazoki, line 76: why this change? I mean here you are nowhere indicating that you want to skip the full download is that right?

tmpolaczyk: This is to enable testing:
* `-lsync=trace` is to check the gap sync log
* rpc is to enable collator rpc, which would otherwise be on a random port

girazoki: ok cool

girazoki (approved):

test/suites/zombie_data_preservers/test_zombie_data_preservers.ts
```diff
@@ -208,12 +214,25 @@ describeSuite({
                 console.log((await customHttpProvider.getNetwork()).chainId);
 
                 const signer = new ethers.Wallet(BALTATHAR_PRIVATE_KEY, customHttpProvider);
-                const tx = await signer.sendTransaction({
-                    to: CHARLETH_ADDRESS,
-                    value: parseUnits("0.001", "ether"),
-                });
 
-                await customHttpProvider.waitForTransaction(tx.hash);
+                for (let i = 0; i <= 5; i++) {
```
evgeny-s, line 218: Why don't we just increase a timeout to (60_000 x 5), instead of waiting 5 times and then failing? 
Something like:
```
try {
  const nonce = await customHttpProvider.getTransactionCount(BALTATHAR_ADDRESS);
  const tx = await signer.sendTransaction({
    to: CHARLETH_ADDRESS,
    value: parseUnits("0.001", "ether"),
    nonce,
  });

  await customHttpProvider.waitForTransaction(tx.hash, 1, 60_000 * 5);
} catch (e) {
  console.log("tx inclusion failed: ", e);
  expect.fail("failed to send tx");
}


```

evgeny-s (approved):

test/suites/zombie_data_preservers/test_zombie_data_preservers.ts
tmpolaczyk: We need to send the transaction again, if not it will never get included. I will add a comment to explain this as it's not obvious

PR #930, opened 2025-03-19
Improve fetching information in Pooled Staking

nanocryk:
- [x] Easily get the list of a delegator's delegations
- [x] Easily get the amount of delegators of a candidate
- [X] ~~Migrations: compute summaries by iterating the pools and insert them~~
- [x] Write a multi-block migration due to the big amount of delegations present in existing network

These changes adds 2 read+write to each extrinsic that modifies pool values. Since they shouldn't be called that often by users I think this is an acceptable cost given the benefit for fetching data.

### Delegator's delegations

To query the list of delegations of a candidate, use
```ts
const keys = await polkadotJs.query.pooledStaking.delegatorCandidateSummaries.keys(delegatorAddress);
const delegationCandidates = keys.map(({ args: [_delegator, candidate] }) => candidate);
```
`delegationCandidates` will be an array of all the candidates this delegator has delegated to.

To avoid fetching unnecessary pool data, the value of this storage is a bitfield encoding in which pools of this candidate the delegator is into. 
```ts
const poolsBits = await polkadotJs.query.pooledStaking.delegatorCandidateSummaries(delegatorAddress, candidateAddress);
```
1 = Joining / 2 = auto compounding / 4 = manual rewards / 8 = leaving
Having a bit set guarantees that the amount of shares for that pool is non-zero, and dApp should request the amount of shares, total amount of shares and total staked to compute the amount of stake represented by those shares.  

### Candidate stats

A candidate stats are summarized in a storage that can be queries with
```ts
const summary = await polkadotJs.query.pooledStaking.candidateSummaries(candidateAddress);
```
The returned summary contains the following fields:
- `delegators`: number of delegators, without duplicates (if a delegator is in multiple pools it counts for 1 only)
- `joiningDelegators`: number of delegators in joining pool (multiple joining requests from same delegator count for 1 only)
- `autoCompoundingDelegators`: number of delegators in auto compounding pool
- `manualRewardsDelegators`! number of delegators in manual rewards pool
- `leavingDelegators`: number of delegators in leaving pool (multiple leaving requests from same delegator count for 1 only)


pallets/pooled-staking/src/migrations.rs
```diff
@@ -0,0 +1,80 @@
...
+        write_count += 1;
+        DelegatorCandidateSummaries::<T>::insert(delegator.clone(), candidate.clone(), summary);
+    }
+
+    for (candidate, summary) in candidate_summaries.into_iter() {
+        write_count += 1;
+        CandidateSummaries::<T>::insert(candidate, summary);
+    }
+
+    T::DbWeight::get().reads_writes(read_count, write_count)
```
tmpolaczyk, line 79: What would be the easiest way to do this migration if it uses more than one block of weight? Any ideas?

nanocryk: Iterate over `Pools` entries, accumulate the numbers and when we run out of available weight we store as migration state the output of `iterator.last_raw_key()`, which can be used to resume the iteration in the next block. 

I just need to rewrite the migration code in a way that doesn't need to accumulate in a big BTreeMap, and instead probably directly read/write from storage. For a candidate delegators count (any pool without duplicate), I should be able to do it without the BTreeMap by increasing this number only if the delegator summary goes from empty to some value.

pallets/pooled-staking/src/lib.rs
```diff
@@ -374,6 +354,24 @@ pub mod pallet {
...
+        Candidate<T>,
+        DelegatorCandidateSummary,
+        ValueQuery,
+    >;
+
+    /// Summary of a candidate state.
+    #[pallet::storage]
+    pub type CandidateSummaries<T: Config> =
+        StorageMap<_, Blake2_128Concat, Candidate<T>, CandidateSummary, ValueQuery>;
+
```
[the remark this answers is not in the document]
nanocryk: Need to find a way to make it compatible with migration code, in such a way that it avoids duplicates. I can probably make the migration function generic over the storage types, such that it can fill either the real ones or test ones, and then iterate over them and ensure they match.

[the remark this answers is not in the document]
tmpolaczyk: Or just copy the function and change a few things, after the migration is done we will remove the duplicate one

pallets/pooled-staking/src/migrations.rs
tmpolaczyk: Ok sounds hard. Nothing will break if the pools are updated while the migration is running?

nanocryk: Extrinsics needs to be paused.
Rewarding can change amounts, but not change value from/to zero; which is what counts for the summaries, so we don't need to pause rewarding.

pallets/pooled-staking/src/calls.rs
```diff
@@ -113,9 +113,14 @@ impl<T: Config> Calls<T> {
     pub fn request_delegate(
         candidate: Candidate<T>,
         delegator: Delegator<T>,
-        pool: TargetPool,
+        pool: ActivePoolKind,
         stake: T::Balance,
     ) -> DispatchResultWithPostInfo {
+        ensure!(
+            !PausePoolsExtrinsics::<T>::get(),
+            Error::<T>::PoolsExtrinsicsArePaused
```
tmpolaczyk, line 121: I believe this storage item is not needed because ALL extrinsics for all pallets are paused during a multi block migration. But not sure, need to verify that

nanocryk: Let's keep it just in case. It can also be useful at some point if we notice an issue with the pallet and what to reduce damages quickly.

girazoki:
For me this is an approval but I want to test it just a little bit more. once I do that I will approve

tmpolaczyk (approved):

PR #931, opened 2025-03-19
Use 12 decimals for UNIT

dimartiro:
# Description

Currently we are using 18 decimales instead of 12 for the UNIT const.
This PR updates this to use 12 decimals like polkadot

### DIFF
--- a/chains/container-chains/runtime-templates/frontier/src/lib.rs
+++ b/chains/container-chains/runtime-templates/frontier/src/lib.rs
@@ -180,10 +180,10 @@ pub type Executive = frame_executive::Executive<
 pub mod currency {
     use super::Balance;
 
-    pub const MICROUNIT: Balance = 1_000_000_000_000;
-    pub const MILLIUNIT: Balance = 1_000_000_000_000_000;
-    pub const UNIT: Balance = 1_000_000_000_000_000_000;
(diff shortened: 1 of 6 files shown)


chains/container-chains/runtime-templates/frontier/src/lib.rs
```diff
@@ -180,10 +180,10 @@ pub type Executive = frame_executive::Executive<
 pub mod currency {
     use super::Balance;
 
-    pub const MICROUNIT: Balance = 1_000_000_000_000;
-    pub const MILLIUNIT: Balance = 1_000_000_000_000_000;
-    pub const UNIT: Balance = 1_000_000_000_000_000_000;
-    pub const KILOUNIT: Balance = 1_000_000_000_000_000_000_000;
+    pub const MICROUNIT: Balance = 1_000_000;
+    pub const MILLIUNIT: Balance = 1_000_000_000;
+    pub const UNIT: Balance = 1_000_000_000_000;
```
Agusrodri, line 185: We should not update these ones. In this case, this is a frontier (evm) template, so let's keep the 18 decimals here.

girazoki:
@dimartiro there seem to be some tests failing, can you chekc?

Agusrodri (approved):

PR #933, opened 2025-03-20
Fix some easy javascript lints

tmpolaczyk:


typescript-api/scripts/generate-types.ts
```diff
@@ -51,8 +51,8 @@ async function main() {
             process.removeListener("SIGINT", onProcessExit);
         });
 
-        await new Promise(async (resolve, reject) => {
-            const onData = async (data: any) => {
+        await new Promise((resolve, reject) => {
```
evgeny-s, line 54: We might need to update this implementation. There is no need to create a Promise here.

biome.jsonc
```diff
@@ -35,13 +35,9 @@
     "rules": {
       "recommended": true,
       "suspicious": {
-        "noAsyncPromiseExecutor": "off",
         "noExplicitAny": "off",
```
evgeny-s, line 38: It would be great to disable this one as well 😄 , but we will have a lot of places to fix it.

evgeny-s (approved):

typescript-api/scripts/generate-types.ts
tmpolaczyk: That's legacy code from moonbeam, better not to touch it. I also see a sleep(5000) that could be improved, but let's not touch code that works :joy:

PR #935, opened 2025-03-21
Add reward checking to ethereum tests and other fixes

girazoki:
This PR:

- adds reward checking for ethereum tests at the cost of almost no time as it is a test that does not imply waiting
- adds a better handling of the number of sessions to wait before we check balances transfers work
- removes "edited" from the github label to avoid rerunning the test when editting the PR (e.g., comments)


test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -423,6 +453,85 @@ describeSuite({
...
+                const currentEra = (await relayApi.query.externalValidators.activeEra()).unwrap().index;
+
+                let eraToAnalyze = currentEra.toNumber();
+                console.log(await operatorRewardContract.eraRoot(eraToAnalyze));
+
+                // Try to find latest reported era
+                // eraRoot returns a struct of 4 items, the 3rd of which is the era root
+                // we try to retrieve the first non-default one
+                while (
+                    (await operatorRewardContract.eraRoot(eraToAnalyze))[3] ===
```
evgeny-s, line 468: Does indexing start from 1?
You mentioned that you are looking for 3rd argument, so shouldn't it be with index 2?

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -423,6 +453,85 @@ describeSuite({
...
+                        "0x0000000000000000000000000000000000000000000000000000000000000000" &&
+                    eraToAnalyze >= 0
```
evgeny-s, line 470: It would be great if we keep named constants instead of values like this. So it will be clear what it is about. 
So instead of 
```
(await operatorRewardContract.eraRoot(eraToAnalyze))[3] === "0x0000000000000000000000000000000000000000000000000000000000000000"
```
we would have something like:
```
const ERA_ROOT_ADDRESS = "0x0000000000000000000000000000000000000000000000000000000000000000"; // or whatever
(await operatorRewardContract.eraRoot(eraToAnalyze))[3] === ERA_ROOT_ADDRESS

```

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
girazoki: yes sorry, 5 items and I am looking for 4th, will updaget

test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
girazoki: DEFAULT_ERA_ROOT_ADDRESS good?

evgeny-s (approved):

PR #940, opened 2025-03-24
[MD-884] Automatic typescript-api in PRs

evgeny-s:
**The use case:**
- It should be possible to generate TS interfaces automatically via PR without building the node locally in case the developer misses it.

**How it works:**
- When a new push comes to PR, the CI job checks if interfaces are outdated and sends a comment about it. 
- Only developers with permission levels: "write" and "admin" can execute the CI to update TS interfaces.
- The command that can trigger the job: `/cmd generate-ts-api`

**Additional permissions:**
- To be able to commit changes to the repo by CI job, we need to grant a "write" permission for the CI job here: `Settings -> Actions -> General -> Workflow permissions` - "Read and write permissions" instead of "Read repository contents and packages permissions"

**What's done:**
- Added CI job to react to the PR comment for updating TS interfaces
- Send the comment to PR once TS interfaces need to be synced

**Before merging:**

- [ ] Remove test trigger from CI job
- [ ] Enable comment body [check](https://github.com/moondance-labs/tanssi/pull/940/files#diff-228a31e246be7875a1fca18ab855d9ab4a340d767a670a611cf96316e03cf5deR24)


tmpolaczyk:
Does this work? I don't see the commit.

Also, for a bot command I would expect the following:

* Discoverability: how do I know which commands exist
* Debugging: when the command fails, how do I check why
* Security: who can run the commands? anyone from the team or anyone at all?

evgeny-s:
I am testing it now. I don't see it is triggered probably because it is not in the base branch. 
For your questions:
- Discoverability - for now we have only 1 command and you will see it in the chat, probably I can tag the PR owner, so you won't miss it. 
- Debugging - you will see it in the CI jobs, if it is failed or succeed.
- Security - this is a good question, I will see if we can restrict it to only core dev. 

Thanks for the good questions!

.github/workflows/cmd.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Commit and push changes
+        run: |
+          if [[ -n "$(git status --porcelain)" ]]; then
+            cd typescript-api
+            git add .
+            echo "About to commit changes"
+            git commit -m "Auto-generated TypeScript interfaces"
+            echo "About to push changes"
+            git push -u origin $(git branch --show-current)
```
tmpolaczyk, line 111: Not sure if this branch name works, did you test it? Because not sure if the script ever checks out the branch, or only the commit. But if it checks out the branch, then just "git push" should also work?

evgeny-s: No, this part has not been tested yet, because there is no permissions to push. If we check the commit instead of branch, we can read the branch from the PR, will check.

tmpolaczyk: I tested this in a private repo and it pushed to master :joy: 

So for the first test, instead of git push let's use echo "Will push to $(git branch --show-current)" and not push

tmpolaczyk (approved):

PR #943, opened 2025-03-26
Refactor typescript functions into utils + more lints

tmpolaczyk:
* Add hacky script to find functions with the same name in different files
* Refactor all the duplicate functions into utils
* For some reason moving functions into utils fixes the linter in my IDE, so fix some lints as well
* Fix missing await in waitForLogs function, was always returning true
* Tried to enable a linter rule for that, but its still in beta: https://next.biomejs.dev/linter/rules/no-floating-promises/
* Add noUnusedImports rule to remove unused imports


girazoki (approved):

test/scripts/duplicateFunctionFinder.ts
```diff
@@ -0,0 +1,95 @@
+// Crude duplicate function finder because we copy paste tests a lot.
+// Any duplicates should be refactored into "utils" package.
```
evgeny-s, line 2: 🔥

test/suites/zombie_data_preservers/test_zombie_data_preservers.ts
```diff
@@ -1,19 +1,21 @@
...
 import { u8aToHex } from "@polkadot/util";
 import { decodeAddress } from "@polkadot/util-crypto";
 import { WebSocketProvider, ethers, parseUnits } from "ethers";
-import fs from "node:fs/promises";
-import { getHeaderFromRelay, getTmpZombiePath, signAndSendAndInclude, sleep } from "utils";
+import { getHeaderFromRelay, getTmpZombiePath, signAndSendAndInclude, waitForLogs } from "utils";
+
+// Checks every second the log file to find the watcher best block notification until it is found or
+// timeout is reached. If timeout is reached, throws an error.
+export async function waitForLogs2(logFilePath: string, timeout: number, logs: string[]): Promise<void> {
```
evgeny-s, line 13: we need either include waitLogs to this function and call it waitLogs or come up with a better name, because I don't think it's good to have function like waitLogs, waitLogs2, etc

evgeny-s (approved):
I think in general we should avoid having any helper function in tests, but rather move all of them to utils, even if we use it once.

test/suites/zombie_data_preservers/test_zombie_data_preservers.ts
tmpolaczyk: Yeah this was a temporary name, will rename it to assertWaitForLogs or similar

PR #953, opened 2025-04-01
Test TS api generate command

evgeny-s:
**Description:**
- This PR fix the CI job that can be triggered from the PR comment.

**What's done:**
- Updated the checkout command
- Added rebase in case if any new commit appeared after job started

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -406,13 +406,8 @@ jobs:
                   pnpm create-local-interfaces
             - name: Check for changes
               run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -407,17 +407,12 @@ jobs:
             - name: Check for changes
               run: |
                   git diff --exit-code || (echo "There are uncommitted changes after running typegen script.\n Please make sure to run the command before pushing." && exit 1)
```
tmpolaczyk, line 409: add /cmd here as well

evgeny-s: Updated

tmpolaczyk (approved):

PR #957, opened 2025-04-01
Test `/cmd generate-ts-api` to trigger Typescript interfaces update

evgeny-s:
**Description:**
This PR is for testing the Typescript interfaces update job via the `/cmd generate-ts-api`


evgeny-s:
> did it work?

@girazoki, yes, as you can see above, the commit from the GitHub action bot was successfully received, but actions were not triggered. 
So the only downside is you need to push anything to re-run all the checks. 
Theoretically just updating the branch from "master" would work.

evgeny-s:
Closing it.

PR #967, opened 2025-04-04
Fix swapped inflation rates in dancelight

nanocryk:
Inflation rates are swapped between block and era. This PR fixes it and add tests ensuring they are used properly.


ParthDesai (approved):
Looks good. Just change the label to breaking.

girazoki (approved):

PR #969, opened 2025-04-07
fix: unsafe math operations

chexware:
This PR replaces unsafe math operations with the `saturating` or `checked` equivalent. Addresses https://github.com/moondance-labs/tanssi-sr-labs/issues/26


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -715,14 +716,14 @@ where
         // take the biggest period possible.
         let period = BlockHashCount::get()
             .checked_next_power_of_two()
-            .map(|c| c / 2)
+            .map(|c| c.saturating_div(2))
```
tmpolaczyk, line 719: This function is useless because it still panics when dividing by 0, I propose we ban it.

Can you try to do that? Using https://rust-lang.github.io/rust-clippy/master/#disallowed_methods

And here use either checked_div or nothing, I think nothing looks better

```suggestion
            .map(|c| c / 2)
```
```suggestion
            .map(|c| c.checked_div(2).expect("2 != 0; qed"))
```

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -715,14 +716,14 @@ where
...
             .unwrap_or(2) as u64;
 
         let current_block = System::block_number()
             .saturated_into::<u64>()
             // The `System::block_number` is initialized with `n+1`,
             // so the actual block number is `n`.
-            .saturating_sub(1);
+            .saturating_less_one();
```
tmpolaczyk, line 726: I hate that that this function exists, but whatever, there must be some reason for it

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
girazoki: agreed

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
chexware: Reverted in https://github.com/moondance-labs/tanssi/pull/969/commits/784b13e0eec61e060080f45b8ded9f3093f0b998

pallets/collator-assignment/src/assignment.rs
```diff
@@ -435,7 +438,9 @@ where
         let mut needed_new_collators = 0;
         for (para_id, num_collators) in chains.iter() {
             let cs = old_assigned.entry(*para_id).or_default();
-            needed_new_collators += (*num_collators as usize).saturating_sub(cs.len());
+            needed_new_collators = needed_new_collators
```
girazoki, line 441: this can be saturating_accrue and saturating_reduce no?

chexware: So  `saturating_accrue` and `saturating_reduce` and directly modifies `needed_new_collators`. So it needs to be applies either this way or  or  as 

```
needed_new_collator.saturating_accrue();  
needed_new_collator.saturating_reduce();
```

Otherwise it does not calculate the correct value and breaks the tests.

evgeny-s (approved):

pallets/collator-assignment/src/assignment.rs
girazoki: ok then leave it as it is, im fine with it!

tmpolaczyk (approved):

pallets/xcm-core-buyer/src/lib.rs
```diff
@@ -125,7 +126,7 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
         for info in info {
             let para_id = info.para_id;
             PendingBlocks::<T>::remove(para_id);
-            writes += 1;
+            writes.saturating_inc();
```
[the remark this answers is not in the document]
chexware: Good point! Updated in https://github.com/moondance-labs/tanssi/pull/969/commits/afd504b81abdf217f128874260537a05824d15f4

PR #970, opened 2025-04-07
unhash-key script

tmpolaczyk:
Given a raw storage key, try to find which pallet it belongs to, and which storage item

Usage:

```
# List all pallet prefixes for a selected chain
pnpm unhash-key
# Search for a key in all endpoints
pnpm unhash-key 0x94eadf0156a8ad5156507773d0471e4a49f6c9aa90c04982c05388649310f22f
# Search for a key in a specific endpoint
pnpm unhash-key --url 'wss://dancelight.tanssi-api.network' 0x94eadf0156a8ad5156507773d0471e4a49f6c9aa90c04982c05388649310f22f
```

Sample output:

```
✅ Found matching prefix: pallet ParaScheduler 0x94eadf0156a8ad5156507773d0471e4a
✅ Found key in network Polkadot:

0x94eadf0156a8ad5156507773d0471e4a49f6c9aa90c04982c05388649310f22f: ParaScheduler ClaimQueue
```

### DIFF
--- a/test/scripts/unhashKey.ts
+++ b/test/scripts/unhashKey.ts
@@ -1,7 +1,8 @@
 /**
 unhash-key
 
-Given a raw storage key, try to find which pallet it belongs to, and which storage item
+Given a raw storage key, try to find which pallet it belongs to, and which storage item.
+Only works for public storage items (exported with pub).
 
 Usage:
 # List all pallet prefixes for a selected chain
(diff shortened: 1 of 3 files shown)


test/scripts/unhashKey.ts
```diff
@@ -0,0 +1,179 @@
...
+                console.log("No rpc provided. Will try to check all the known endpoints. Use --url to specify one");
+                endpoints = DEFAULT_ENDPOINTS;
+            }
+
+            let found = false;
+            for (const endpoint of endpoints) {
+                console.log(`\nTrying network endpoint: ${endpoint}`);
+                let api: ApiPromise;
+                try {
+                    // Wrap connection attempt in a 10 second timeout.
```
evgeny-s, line 119: We might want to abstract the timeout implementation to the `getApiFor` function with an optional parameter `timeout`.

evgeny-s (approved):

test/scripts/unhashKey.ts
tmpolaczyk: I believe the default timeout is 60 seconds, or more, I will try to change that to 10 as it doesn't make sense to wait that much

PR #972, opened 2025-04-08
Add xcm versioning test to chopsticks upgrade

girazoki:
What this new test does (for non-flashbox chains) is:

- For "*light" chains, it creates a random query with para id 1. after the upgrade, it checks that the version of the query on-storage matches that advertised by the pallet

- For "*box" chains, it creates a query with the relay. THe reason is that mocking the connection to a parachain is much more difficult here. We also reset the query counter because some chains might have already a query to the relay, so we start fresh. The later verification is identical in both cases.

- For containers, same as for "box" chains

Unfortunately I did not get much help from the API for this

Companion: https://github.com/paritytech/polkadot-sdk/pull/8173

### DIFF
--- a/test/configs/dancebox.yml
+++ b/test/configs/dancebox.yml
@@ -12,3 +12,5 @@ import-storage:
                     free: "100000000000000000000000"
     Sudo:
         Key: "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
+    PolkadotXcm:
+        QueryCounter: 0
--- a/test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
+++ b/test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
@@ -62,7 +62,7 @@ describeSuite({
                     // we send queries
                     batchTx.push(api.tx.registrar.setCurrentHead(1, "0x11"));
                     batchTx.push(api.tx.xcmPallet.forceSubscribeVersionNotify(versionedLocation));
-                } else if (runtimeName !== "flasbox") {
+                } else if (runtimeName !== "flashbox") {
                     batchTx.push(api.tx.polkadotXcm.forceSubscribeVersionNotify(versionedLocation));
                 }
 
--- a/test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
+++ b/test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
@@ -2,6 +2,11 @@ import "@tanssi/api-augment/dancelight";
 import { MoonwallContext, beforeAll, describeSuite, expect } from "@moonwall/cli";
 import { generateKeyringPair } from "@moonwall/util";
 import { Keyring, type ApiPromise } from "@polkadot/api";
+import type { KeyringPair } from "@polkadot/keyring/types";
+import { type MultiLocation } from "utils";
+import { PalletXcmQueryStatus } from "@polkadot/types/lookup";
+import { VersionedMultiLocation } from "@polkadot/types/interfaces";
+import { xcm } from "@polkadot/types/interfaces/definitions";
 
 const MAX_BALANCE_TRANSFER_TRIES = 5;
 describeSuite({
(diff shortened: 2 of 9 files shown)


test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -11,17 +19,77 @@ describeSuite({
...
+                    // we send queries
+                    batchTx.push(api.tx.registrar.setCurrentHead(1, "0x11"));
+                    batchTx.push(api.tx.xcmPallet.forceSubscribeVersionNotify(versionedLocation));
+                } else if (runtimeName != "flasbox") {
+                    batchTx.push(api.tx.polkadotXcm.forceSubscribeVersionNotify(versionedLocation));
+                }
+
+                let tries = 0;
+
+                while (tries < MAX_BALANCE_TRANSFER_TRIES) {
```
girazoki, line 76: this should be reworked to something like `chopsticks_wait_till_included`

test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -11,17 +19,77 @@ describeSuite({
...
+                          parents: 1,
+                          interior: "Here",
+                      };
+
+                // fetch on-chain later
+                // TODO:Once we update the next runtime, remove this and take it form onchain
+                const previousXcmVersion = 5;
+                const latestVersion = "V" + previousXcmVersion.toString();
+
+                const versionedLocation = new Object();
```
evgeny-s, line 57: It's a bit outdated way to create objects in JS. 
You can change it to something like this:
```suggestion
  const versionedLocation = {
  [latestVersion]: queryLocation
};
```

evgeny-s, line 57: It's a bit outdated way to create objects in JS.
You can change it to something like this:
```suggestion
const versionedLocation = {
  [latestVersion]: queryLocation
};
```

test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -11,17 +19,77 @@ describeSuite({
...
+                versionedLocation[latestVersion] = queryLocation;
```
evgeny-s, line 58: It's a bit outdated way to create objects in JS.
You can change it to something like this:

```suggestion
const versionedLocation = {
  [latestVersion]: queryLocation
};
```

test/configs/dancebox.yml
```diff
@@ -12,3 +12,5 @@ import-storage:
                     free: "100000000000000000000000"
     Sudo:
         Key: "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY"
+    PolkadotXcm:
```
evgeny-s, line 15: Where do we use these constants?

test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
girazoki: thx

test/configs/dancebox.yml
girazoki: these are used so that we reset the query counter. it's something very much xcm specific:

- the xcm migration mostly migrate queries and locked assets. the latter are a bit difficult to create, so I chose to go with the first.
- in order to create a xcm query, you need a valid chain to which you can talk to. in the case of parachains, this is usually the relay chain. However some of them **might already have a query to the relay chain**, hence the reason for us to reset the counter to 0, so that we start fresh.
- in the case of the solo chain, we simply create a fake query to a fake para-id. we cannot do this in the parachain context as it is a more complex and convoluted thing to mock

test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -3,25 +3,77 @@ import { MoonwallContext, beforeAll, describeSuite, expect } from "@moonwall/cli
...
+                    ? await api.query.xcmPallet.queryCounter()
+                    : await api.query.polkadotXcm.queryCounter();
+
+                const batchTx = [];
+                if (runtimeName.includes("light")) {
+                    // We first inject a random paras head in parachain 1. this is necessary as we need a chain to which
+                    // we send queries
+                    batchTx.push(api.tx.registrar.setCurrentHead(1, "0x11"));
+                    batchTx.push(api.tx.xcmPallet.forceSubscribeVersionNotify(versionedLocation));
+                } else if (runtimeName !== "flasbox") {
```
tmpolaczyk, line 65:
```suggestion
                } else if (runtimeName !== "flashbox") {
```

test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -100,6 +136,27 @@ describeSuite({
...
+            title: "Xcm migrations have runned, if any",
+            test: async ({ skip }) => {
+                if (runtimeName === "flashbox") {
+                    skip();
+                }
+                // Regardless of migrations, query should be in the latest version
+                // Unfortunately the api is not very friendly
+                const currentXcmVersion = runtimeName.includes("light")
+                    ? (await api.consts.xcmPallet.advertisedXcmVersion).toNumber()
+                    : (await api.consts.polkadotXcm.advertisedXcmVersion).toNumber();
```
tmpolaczyk, line 148: I can't find this in polkadot js, where is it?

girazoki: haha not deployed yet. but I cherry-picked. its the companion PR

tmpolaczyk (approved):

PR #973, opened 2025-04-08
increase ts api to 1300

girazoki:


typescript-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@tanssi/api-augment",
-    "version": "0.800.1",
+    "version": "0.1300.0",
```
Agusrodri, line 3: Shouldn't this be 1200? Or maybe change the PR title :)

.github/workflow-templates/cargo-build/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
   using: "composite"
   steps:
     - name: Run sccache-cache
-      uses: mozilla-actions/sccache-action@v0.0.4
+      uses: mozilla-actions/sccache-action@v0.0.9
```
tmpolaczyk, line 15: This is to fix this CI error:

```
sccache: error: Server startup failed: cache storage failed to read: Unexpected (permanent) at read => {"$id":"1","innerException":null,"message":"This legacy service is shutting down, effective April 15, 2025. Migrate to the new service ASAP. For more information: https://gh.io/gha-cache-sunset",
```

tmpolaczyk (approved):

typescript-api/package.json
evgeny-s: Could you explain why do we add the next major runtime version here?
If the dev wants to use some API, and he see that node has 1200, how he should know that he need to have 1300 version?
Probably the JS package versioning should be detached from the runtime. 
Sometimes we might have a runtime release without API change, right? In this case we will have gaps in versions. 

Polkadot seems like has their own versioning for JS packages:
https://github.com/polkadot-js/api/releases

evgeny-s (approved):

typescript-api/package.json
girazoki: I am going to publish from perm-runtime-1200 and add there the 1200 version as a commit.

This one is so that the next time we create the perm-runtime-1300, we get the right version.

Note that current master is not probably aligned already with runtime 1200, hence it would not make sense put 1200

PR #975, opened 2025-04-08
Add xcm fungible benchmarks for dancebox

dimartiro:
# Description

Add benchmarks for xcm fungibles in dancebox

- [x] Add `pallet_xcm_benchmarks::fungible` in `dancebox`
- [x] Add and use benchmarks

Benchmarks where generated using:

```
BINARY=target/release/tanssi-node TEMPLATE_PATH=benchmarking/frame-weight-runtime-template-xcm.hbs OUTPUT_PATH=chains/orchestrator-paras/runtime/dancebox/src/weights/xcm ./tools/benchmarking.sh "pallet_xcm_benchmarks::fungible" "*"
```


chains/orchestrator-paras/runtime/dancebox/src/lib.rs
```diff
@@ -2038,11 +2041,35 @@ impl_runtime_apis! {
 
             use xcm::latest::prelude::*;
             use crate::xcm_config::SelfReserve;
+
             parameter_types! {
                 pub ExistentialDepositAsset: Option<Asset> = Some((
                     SelfReserve::get(),
                     ExistentialDeposit::get()
                 ).into());
+                pub TrustedReserve: Option<(Location, Asset)> = Some(
```
girazoki, line 2050: hmm I am not sure if this what we would consider a trusted reserve in dancebox, we need to revisit this, will let u know

Agusrodri: From my understanding, this should be configured with the relay location/asset, to (explicitly) let dancebox's benchmarks know that the relay is the chain that should be recognized as a reserve for the relay token.

chains/orchestrator-paras/runtime/dancebox/src/lib.rs
```diff
@@ -2038,11 +2041,35 @@ impl_runtime_apis! {
...
+            impl pallet_xcm_benchmarks::fungible::Config for Runtime {
+                type TransactAsset = Balances;
+                type CheckedAccount = ();
+                type TrustedTeleporter = ();
+                type TrustedReserve = TrustedReserve;
+
+                fn get_asset() -> Asset {
+                    Asset {
+                        id: AssetId(SelfReserve::get()),
+                        fun: Fungible(1 * UNIT),
```
Agusrodri, line 2070: I suppose this small amount doesn't affect too much in benchmarks, but I would try putting more just to ensure we don't underestimate anything.

chains/orchestrator-paras/runtime/dancebox/src/lib.rs
girazoki: yeah this looks better!

chains/orchestrator-paras/runtime/dancebox/src/lib.rs
girazoki: yeah let's put 100 existential deposit, sounds good?

girazoki (approved):
looks good!

PR #976, opened 2025-04-09
Test pallet versions after a migration

tmpolaczyk:
The pallet version is a storage item defined using the `pallet(storage_version)` macro. It is a number that is increased every time there is a migration that needs to be run. When the migration completes, it updates the on chain value.

This PR extends our upgrade tests with a check that compares:
* The pallet version after running a runtime upgrade
* With the pallet version that is generated if a new chain is started with the new runtime

Since the new chain will always start at the latest version, this is a hacky way of getting the latest storage version, which is not part of the metadata.


test/suites/rt-upgrade-chopsticks-orchestrator/test-upgrade-chain.ts
```diff
@@ -219,6 +219,17 @@ describeSuite({
...
         });
 
+        it({
+            id: "T6",
+            title: "Test pallet versions for missed migrations",
+            test: async () => {
+                const command = "../target/release/tanssi-node";
+                const args = ["build-spec", "--chain=dancebox-local", "--raw"];
+
+                await testPalletVersions(api, command, args);
```
tmpolaczyk, line 229: This fails with this message:

```
Error: ❌ 2 pallet version mismatches out of 38 pallets. Missing migrations?
  xcmpQueue   : 4 != 5
  asyncBacking: 0 != 1
```

The xcmpQueue is a known issue, but for asyncBacking I don't see any migration in moonkit. Moonbeam has version 1 on chain so maybe they did have some migration.

https://github.com/moondance-labs/moonkit/blob/002c1a5ae3ce82e1136d5321bff124ab93ed050f/pallets/async-backing/src/lib.rs#L82

girazoki: we might want to put version 5 in xcmpqueue manually

tmpolaczyk: Ok the async backing is not 0 but empty. In previous versions, the pallet version had to be manually initialized. But now it is automatic, so if we start a new chain this will have a 1, not empty. Polkadot has a migration to set the correct version for all the pallets that do not have one, I don't think we need that because we only have this one pallet. We probably have more pallets with no version but they are defined to be version 0 so it passes the test.

So we might want to set this version to 1 manually, I will add this to the xcmpqueue ticket

test/utils/migrations.ts
```diff
@@ -0,0 +1,169 @@
...
+    for (const moduleName in api.query) {
+        // Ensure that the module has a 'palletVersion' query method.
+        const moduleQuery = api.query[moduleName];
+        if (typeof moduleQuery.palletVersion !== "function") {
+            continue;
+        }
+
+        // Call the query and convert the result to a number.
+        const palletVer = await moduleQuery.palletVersion();
+        versions[moduleName] = palletVer.toNumber();
```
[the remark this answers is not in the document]
tmpolaczyk: How do I do that? It works now

evgeny-s (approved):

girazoki (approved):

PR #980, opened 2025-04-11
Benchmark missing pallets

nanocryk:
`pallet_session` and `parachains_slashing` currently use test weights and are not benchmarks. They are not trivial to bench as their original benchmarking implementation requires `pallet_staking` which we don't use.

- `pallet_session`: a simpler benchmark is implemented in `cumulus_pallet_session_benchmarking` and used in other runtimes. It actually doesn't depend on cumulus and have pretty straitforward implementation.
- `parachains_slashing`: this one doesn't seem to have an alternative bench so I had to write it. Currently it doesn't work as the bench runs for a long time and crash. Still need to figure out why.


pallets/alt-benchmarks/src/bench_parachains_slashing.rs
```diff
@@ -0,0 +1,186 @@
...
+        log::info!("setup_dispute");
+        // submit a single `ForInvalid` dispute for a past session.
+        let dispute_proof = setup_dispute::<T>(session_index, validator_id);
+
+        log::info!("extrinsic call");
+        #[extrinsic_call]
+        _(RawOrigin::None, Box::new(dispute_proof), key_owner_proof);
+
+        // // TODO: Storage is pub(crate) so cannot be accessed here
+        // let unapplied = <UnappliedSlashes<T>>::get(session_index, CANDIDATE_HASH);
```
girazoki, line 183: which storage is this? is it ours? if so just make it public

girazoki (commented):
@nanocryk we are missing another one besides pallet_session righyt?

pallets/alt-benchmarks/src/bench_parachains_slashing.rs
nanocryk: It is part of `parachains_slashing` which is part of PolkadotSDK.

nanocryk:
The 2 ones I see in Dancelight that are actually benchmarked in Kusama are `pallet_session` and `parachain_slashing`.

The following ones are not benchmarked in Kusama either: `pallet_babe`, `pallet_beefy` and `pallet_grandpa`.

girazoki (approved):

chexware (approved):

PR #982, opened 2025-04-14
Enable eth foreign asset registration in dancelight

dimartiro:
# Description

Enable ethereum foreign assets registration in dancelight through `pallet_foreign_asset_creator` ensuring only root can create, modify or destroy these tokens

## Summary

- [x] Add required pallets `pallet_assets` and `pallet_foreign_asset_creator`
- [x] Add and update benchmarks
- [x] Integration tests


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -2289,7 +2294,9 @@ mod benches {
         [pallet_pooled_staking, PooledStaking]
         [pallet_configuration, CollatorConfiguration]
         [pallet_stream_payment, StreamPayment]
-
+        // Foreign Assets
+        //[pallet_foreign_asset_creator, ForeignAssetsCreator]
```
girazoki, line 2298: is there a reason to comment this?

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -3124,11 +3131,27 @@ sp_api::impl_runtime_apis! {
             }
 
             parameter_types! {
-                pub TrustedTeleporter: Option<(Location, Asset)> = Some((
-                    AssetHub::get(),
-                    Asset { fun: Fungible(1 * UNITS), id: AssetId(TokenLocation::get()) },
-                ));
-                pub TrustedReserve: Option<(Location, Asset)> = None;
+                pub TrustedTeleporter: Option<(Location, Asset)> = None;
+                pub TrustedReserve: Option<(Location, Asset)> = Some(
```
girazoki, line 3135: We should look into this, as probably we want to make this be Ethereum-based, but we will let u know

chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
```diff
@@ -291,3 +292,61 @@ impl pallet_xcm::Config for Runtime {
...
     type AdminOrigin = EnsureRoot<AccountId>;
 }
+
+parameter_types! {
+    pub const ForeignAssetsAssetDeposit: Balance = 0;
+    pub const ForeignAssetsAssetAccountDeposit: Balance = 0;
+    pub const ForeignAssetsApprovalDeposit: Balance = 0;
+    pub const ForeignAssetsAssetsStringLimit: u32 = 50;
+    pub const ForeignAssetsMetadataDepositBase: Balance = 0;
+    pub const ForeignAssetsMetadataDepositPerByte: Balance = 0;
```
Agusrodri, line 302: We should also revisit these values in the future.

Agusrodri (approved):

chains/orchestrator-relays/runtime/dancelight/src/tests/integration_test.rs
```diff
@@ -625,3 +626,132 @@ fn stream_payment_stored_profile_correct_size() {
...
+fn test_register_same_eth_foreign_asset_twice_should_fail() {
+    ExtBuilder::default()
+        .with_balances(vec![
+            (AccountId::from(ALICE), 210_000 * UNIT),
+            (AccountId::from(BOB), 100_000 * UNIT),
+        ])
+        .build()
+        .execute_with(|| {
+            let asset_location = Location {
+                parents: 0,
```
Agusrodri, line 711: Just noticed that this should be `parents: 1`. It will not affect on tests but just a reminder that to represent an Ethereum asset, we need to go one step above.

chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
girazoki: yers, let's add a TODO

chains/orchestrator-relays/runtime/dancelight/src/tests/integration_test.rs
```diff
@@ -625,3 +626,132 @@ fn stream_payment_stored_profile_correct_size() {
...
+            let asset_id = 42u16;
+
+            assert_ok!(ForeignAssetsCreator::create_foreign_asset(
+                root_origin(),
+                asset_location,
+                asset_id,
+                AccountId::from(BOB),
+                true,
+                1
+            ));
```
girazoki, line 660: after this let's test that the asset-id ad ML are correct in the pallet

test/suites/dev-tanssi-relay/foreign-assets/test-foreign-assets.ts
```diff
@@ -0,0 +1,131 @@
...
+            id: "T01",
+            title: "Should succeed calling foreign assets creation for ethereum token and sudo",
+            test: async () => {
+                if (shouldSkipStarlightForeignAssetsCreation) {
+                    console.log(`Skipping DEVT2001T01 test for Starlight version ${specVersion}`);
+                    return;
+                }
+                
+                const tokenLocation = {
+                    parents: 0,
```
girazoki, line 38: same here, the theory says this should be parents:1, ti should not affect the test though

chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
girazoki: in any case, since this is only sudo (for now) that si going to use it, I think it's fine

girazoki (approved):
Minor comments, looks good overall. Dont forget to add the new benchmarks (benchmark them yourself on your machine, and before the new release we will do it on the proper benchmarking server)

PR #984, opened 2025-04-15
Disable TrustedTeleporter in fungible benchmarks

dimartiro:
# Description

Fix fungible benchmarks in dancelight to do not support teleported assets


girazoki (approved):
thanks @dimartiro !

Agusrodri (approved):

PR #986, opened 2025-04-15
Limit params accepted by root_test_send_msg_to_eth

tmpolaczyk:


pallets/external-validator-slashes/src/lib.rs
```diff
@@ -329,6 +331,12 @@ pub mod pallet {
         ) -> DispatchResult {
             ensure_root(origin)?;
 
+            // Ensure we don't accidentally pass huge params that would stall the chain
+            ensure!(
+                num_msgs <= 32 && msg_size <= 900,
```
girazoki, line 336: shall we make them config items?

tmpolaczyk: Please no :joy:

tmpolaczyk: Tests are failing because we send a msg with size 1921, and in another test we send 100 messages, so I will increase the limits here and in the benchmark

Agusrodri (approved):

evgeny-s (approved):

PR #994, opened 2025-04-22
[MD-459] XCM core buyer improvements

evgeny-s:
**Description:**
- Fixed wrong parameter for `CoreBuyingXCMQueryTtl`.
- Forcing to set a value for `maxCorePrice`.

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
@@ -41,7 +41,7 @@ use {
     },
     pallet_xcm::EnsureXcm,
     parity_scale_codec::DecodeAll,
-    snowbridge_beacon_primitives::{Fork, ForkVersions},
+    snowbridge_beacon_primitives::ForkVersions,
     snowbridge_core::{gwei, inbound::Message, meth, Channel, PricingParameters, Rewards},
     snowbridge_pallet_inbound_queue::RewardProcessor,
     snowbridge_pallet_outbound_queue::OnNewCommitment,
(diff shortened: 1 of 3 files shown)


tmpolaczyk (approved):

tmpolaczyk:
Typescript API failing because the LookupXX name changed, change Lookup89 to Lookup88 in https://github.com/moondance-labs/tanssi/blob/master/typescript-api/scripts/runtime-upgrade.ts

nanocryk (approved):

evgeny-s:
> Typescript API failing because the LookupXX name changed, change Lookup89 to Lookup88 in https://github.com/moondance-labs/tanssi/blob/master/typescript-api/scripts/runtime-upgrade.ts

Why did it change BTW?
Is it because of my changes?

@tmpolaczyk

tmpolaczyk:
> > Typescript API failing because the LookupXX name changed, change Lookup89 to Lookup88 in https://github.com/moondance-labs/tanssi/blob/master/typescript-api/scripts/runtime-upgrade.ts
> 
> Why did it change BTW? Is it because of my changes?
> 
> @tmpolaczyk

Yes because of some change, not sure which, after we see the typescript-api diff maybe we can understand it

typescript-api/src/dancebox/interfaces/lookup.ts
```diff
@@ -804,7 +804,7 @@ export default {
         },
     },
     /**
-     * Lookup73: pallet_data_preservers::pallet::Event<T>
+     * Lookup72: pallet_data_preservers::pallet::Event<T>
```
tmpolaczyk, line 807: So the typescript-api hack failed because of this, because we didn't have a "Lookup72" before, and now for whatever reason we do. I guess its related to using an Option<u128> above and a bug in the type generation code, wild stuff

PR #998, opened 2025-04-25
[Starlight] Add dev-test suite and adapt existing TS tests

Agusrodri:
## What does it do?

This PR adds the proper configuration to have a dev TS test suite for Starlight runtime. It also adapts the existing tests to behave properly according (call) filters of different Starlight runtime versions.


chains/orchestrator-relays/client/cli/src/command.rs
```diff
@@ -517,6 +517,24 @@ fn load_spec(
                 )?,
             )
         }
+        #[cfg(feature = "starlight-native")]
+        "starlight-dev" => {
```
girazoki, line 521: we should also add starlight-local

test/helpers/starlightVersions.ts
```diff
@@ -0,0 +1,11 @@
+export const STARLIGHT_VERSIONS_TO_EXCLUDE_FROM_BALANCES = [1300];
```
girazoki, line 1: nice!

test/suites/common-all/proxy/test-proxy-cancel.ts
```diff
@@ -41,13 +48,18 @@ describeSuite({
...
                 const tx = polkadotJs.tx.proxy.addProxy(delegate, "Any", delay);
-                await context.createBlock([await tx.signAsync(alice)]);
 
+                if (shouldSkipStarlightProxy) {
+                    console.log(`Skipping E02 test for Starlight version ${specVersion}`);
+                    await checkCallIsFiltered(context, polkadotJs, await tx.signAsync(alice));
+                    return;
+                }
+
+                await context.createBlock();
```
girazoki, line 61: is this one needed?

test/utils/constants.ts
```diff
@@ -20,3 +20,4 @@ export const ASSET_HUB_PARA_ID = 0;
 
 export const HOLESKY_SOVEREIGN_ACCOUNT_ADDRESS = "0x34cdd3f84040fb44d70e83b892797846a8c0a556ce08cd470bf6d4cf7b94ff77";
 export const SEPOLIA_SOVEREIGN_ACCOUNT_ADDRESS = "0xce796ae65569a670d0c1cc1ac12515a3ce21b5fbf729d63d7b289baad070139d";
+export const ETHEREUM_SOVEREIGN_ACCOUNT_ADDRESS = "0x204dfe37731e8e2b4866ad0da9a17c49f434542c3477c5f914a3349acd88ba1a";
```
girazoki, line 23:
```suggestion
export const ETHEREUM_MAINNET_SOVEREIGN_ACCOUNT_ADDRESS = "0x204dfe37731e8e2b4866ad0da9a17c49f434542c3477c5f914a3349acd88ba1a";
```

girazoki (approved):
good job @Agusrodri ! small comments only

test/suites/common-all/proxy/test-proxy-cancel.ts
Agusrodri: Mmm It was there before, I'll check if it's still needed.

Agusrodri: Just checked and it's not needed, so I'll remove it.

test/utils/constants.ts
Agusrodri: Applied in 0b48af36459828fd95bf8c9810ca7c05d004f475.

test/suites/common-all/proxy/test-proxy-cancel.ts
Agusrodri: Applied in 0b48af36459828fd95bf8c9810ca7c05d004f475.

chains/orchestrator-relays/client/cli/src/command.rs
Agusrodri: `starlight-local` added in 0b48af36459828fd95bf8c9810ca7c05d004f475.

test/suites/common-all/proxy/test-proxy-cancel.ts
Agusrodri: At the end it seems to be needed for dancelight, flashbox and dancebox to properly include all the needed txs in the blocks that are being created. I just put it again in 0f7226bd0accde6f1f01988a8be3d8110937b545.

.github/workflows/release.yml
```diff
@@ -674,7 +674,45 @@ jobs:
                   chmod uog+x target/release/tanssi-relay-execute-worker
                   chmod uog+x target/release/tanssi-relay-prepare-worker
                   cd test
-                  pnpm moonwall test dev_tanssi_relay
+                  pnpm moonwall test dev_tanssi_relay_dancelight
+
+    typescript-tests-relay-starlight:
```
evgeny-s, line 679: Do we really want to replicate the same for the starlight? Why can't we just add another run for the starlight? It uses everything the same, no?
 Like 
 ```
 - name: "Run Dev tests (Dancelight)"
    run: |
       ...
       pnpm moonwall test dev_tanssi_relay_dancelight
    
  - name: "Run Dev tests (Starlight)"
    run: |
       ...
       pnpm moonwall test dev_tanssi_relay_starlight
 ```

girazoki: yes I think this is a good idea

Agusrodri: Applied in 79c20e31a534578a0dd3cfe6f18e781a1950afa9.

PR #1004, opened 2025-04-29
Fix on_offence weight when slashing is disabled

tmpolaczyk:
Previously we reported a weight of 0 when doing 1 storage read, in the case when slashing is disabled.


pallets/external-validator-slashes/src/mock.rs
```diff
@@ -58,7 +59,7 @@ impl system::Config for Test {
     type BaseCallFilter = frame_support::traits::Everything;
     type BlockWeights = ();
     type BlockLength = ();
-    type DbWeight = ();
+    type DbWeight = RocksDbWeight;
```
tmpolaczyk, line 62: Without this the cost of db read and writes is 0

girazoki: yep, and this is needed to test, good catch

Agusrodri (approved):

PR #1005, opened 2025-04-29
Pass fast-runtime feature to nextest

tmpolaczyk:
Previously we didn't pass it, but rust tests run very fast so the slowest one was 250 seconds. In starlight `EpochDurationInBlocks` is 6x that of dancelight, so that test should take 6x longer. However it takes ~20x longer. I don't know what is this extra factor, but anyway now with fast-runtime the slowest test takes less than 2 seconds.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,4 @@
+[profile.default]
+# Tests that take more than 60 seconds are considered slow
+# Tests that take 2 times that (2 minutes) are terminated making CI fail
+slow-timeout = { period = "60s", terminate-after = 2 }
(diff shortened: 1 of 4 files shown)


girazoki (approved):

tmpolaczyk:
More info: this is how many blocks we create in starlight vs dancelight, still doesn't explain the 20x factor:

```
dancelight
first era deferred 1
first_session_era_3_pruned 180
[chains/orchestrator-relays/runtime/dancelight/src/tests/common/mod.rs:133:18] session_to_block(n) = 108001


starlight
first era deferred 1
first_session_era_3_pruned 120
[chains/orchestrator-relays/runtime/starlight/src/tests/common/mod.rs:132:18] session_to_block(n) = 432001
```

PR #1009, opened 2025-05-05
Weights rt 1300

girazoki:
Weights generated for runtime 1300. Manually modified the enact_candidate ones as they seemed to be off again


tmpolaczyk:
> Found problematic weights in PR diff

I believe these are the ones from dancelight, and we now see them duplicated because of starlight as well. So nothing new

Agusrodri (approved):

tmpolaczyk (approved):

PR #1010, opened 2025-05-06
[Dancelight] Add maintenance mode pallet

Agusrodri:
## What does it do?

This PR adds the `pallet-maintenance-mode` to Dancelight runtime. It also includes the proper Normal and Maintenance filters which contain the calls to allow and forbid on normal operations and maintenance mode respectively.

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/lib.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/lib.rs
@@ -324,7 +324,7 @@ parameter_types! {
 
 #[derive_impl(frame_system::config_preludes::RelayChainDefaultConfig)]
 impl frame_system::Config for Runtime {
-    type BaseCallFilter = EverythingBut<(IsRelayRegister, IsParathreadRegistrar)>;
+    type BaseCallFilter = MaintenanceMode;
     type BlockWeights = BlockWeights;
     type BlockLength = BlockLength;
     type DbWeight = RocksDbWeight;
@@ -1089,7 +1089,7 @@ impl pallet_message_queue::Config for Runtime {
     type MessageProcessor =
         pallet_message_queue::mock_helpers::NoopMessageProcessor<AggregateMessageOrigin>;
     type QueueChangeHandler = ParaInclusion;
-    type QueuePausedQuery = ();
+    type QueuePausedQuery = MaintenanceMode;
     type WeightInfo = weights::pallet_message_queue::SubstrateWeight<Runtime>;
 }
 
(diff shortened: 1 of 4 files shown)


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1606,6 +1606,50 @@ impl pallet_multiblock_migrations::Config for Runtime {
...
+                | RuntimeCall::StreamPayment(..)
+        )
+    }
+}
+
+/// Normal Call Filter
+pub struct NormalFilter;
+impl Contains<RuntimeCall> for NormalFilter {
+    fn contains(c: &RuntimeCall) -> bool {
+        !IsRelayRegister::contains(c) || !IsParathreadRegistrar::contains(c)
```
tmpolaczyk, line 1641:
```suggestion
        !IsRelayRegister::contains(c) && !IsParathreadRegistrar::contains(c)
```

Don't we have tests?

Agusrodri: Yeah some of the tests are failing. I added the OR clause because the call cannot be a `RelayRegister` and a `ParathreadRegistrar` at the same time, but seems to be failing for some reason, so I will add the previous `EverythingBut` to see if the error is there.

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1606,6 +1606,45 @@ impl pallet_multiblock_migrations::Config for Runtime {
...
 
+/// Maintenance mode Call filter
+pub struct MaintenanceFilter;
+impl Contains<RuntimeCall> for MaintenanceFilter {
+    fn contains(c: &RuntimeCall) -> bool {
+        !matches!(
+            c,
+            RuntimeCall::Balances(..)
+                | RuntimeCall::Registrar(..)
+                | RuntimeCall::Session(..)
```
girazoki, line 1617: Add HRMP here, to prevent contaienr chains calling it

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1606,6 +1606,45 @@ impl pallet_multiblock_migrations::Config for Runtime {
...
 }
 
+/// Maintenance mode Call filter
+pub struct MaintenanceFilter;
+impl Contains<RuntimeCall> for MaintenanceFilter {
+    fn contains(c: &RuntimeCall) -> bool {
+        !matches!(
+            c,
+            RuntimeCall::Balances(..)
+                | RuntimeCall::Registrar(..)
```
girazoki, line 1616: Add identity here, ondemandAssignerProvider, remove multisig (as multisig might be needed for e.g., sudo), I would probably remove proxy too (as it might be needed to execute certain actions)

Agusrodri: Makes sense. Changes applied in 3a07acf72ccf530a74b5ff1d94d59dd2d05e4ec0.

test/suites/common-all/test-maintenance/test-maintenance-mode.ts
```diff
@@ -1,33 +1,44 @@
...
         });
 
         it({
             id: "E01",
             title: "No maintenance mode at genesis",
             test: async () => {
+                if (isStarlight) {
+                    console.log("Skipping E01 test for Starlight: maintenance mode is not available yet");
+                    expect(polkadotJs.query.maintenanceMode).to.be.undefined;
+                    return;
```
tmpolaczyk, line 39: I wish there was a way to automatically skip all tests in beforeAll, but there is not

tmpolaczyk (approved):

girazoki:
@Agusrodri I would say let's merge this

PR #1027, opened 2025-05-13
Add dev_flashbox to coverage

tmpolaczyk:


tmpolaczyk:
As expected coverage doesn't increase because flashbox is not using the native runtime

PR #1033, opened 2025-05-15
Inbound processor to receive ETH tokens in Tanssi

Agusrodri:
## Description

This PR introduces the capability for the Tanssi chain to receive ETH native token and ERC20 tokens from Ethereum.

### `EthTokensLocalProcessor` implementation:

- This processor handles inbound `SendToken` commands from Ethereum.

- It validates the incoming message against the configured token-transfers channel and gateway.

- It checks if the token (ETH native represented by a zero address, or an ERC20 token) has been registered via `pallet_foreign_asset_creator`.

- If valid, it constructs and executes an XCM message containing both `ReserveAssetDeposited` and `DepositAsset` instructions indicating the specified beneficiary account on the Tanssi chain.

- This processor is now part of the `MessageProcessor` tuple in the `snowbridge_pallet_inbound_queue::Config`.

### `EthereumAssetReserve` XCM Filter:

A new XCM asset filter, `EthereumAssetReserve`, has been introduced in `primitives/xcm-commons/src/lib.rs`.
This filter is designed to be used in XCM configurations to ensure that assets claiming to be from Ethereum (identified by `GlobalConsensus(EthereumNetwork)`) indeed originate from the trusted Ethereum location specified by `EthereumLocation::get()`.

### Testing:

Integration tests were added for `EthTokensLocalProcessor` to cover:

- Successful reception of ETH native token (zero address).

- Successful reception of ERC20 tokens (non-zero address).

- Failure cases where the ETH native token or ERC20 token is not registered in `ForeignAssetsCreator`.

### DIFF
--- a/test/scripts/download-ethereum-client-test-files.sh
+++ b/test/scripts/download-ethereum-client-test-files.sh
@@ -9,6 +9,16 @@ cd $(dirname $0)/..
 # Grab Polkadot version
 branch=$(egrep -o '/polkadot-sdk.*#([^\"]*)' ../Cargo.lock | head -1)
 polkadot_release=$(echo $branch | sed 's/.*branch=//' | sed 's/#.*//')
+
+if [ -z "$polkadot_release" ]; then
+    echo "Failed to read polkadot-sdk branch name from Cargo.lock"
+    echo "This happens when using diener or custom branch names"
+    echo "Edit script $0 as a workaround"
+    # Uncomment this to override version
+    #polkadot_release="stable2412"
+    exit 1;
+fi
+
 if [ -f tmp/ethereum_client_test/latest_version.txt ]; then
     stored_version=$(< tmp/ethereum_client_test/latest_version.txt)
     if [[ "$polkadot_release" == "$stored_version" ]]; then


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,179 @@ where
...
+                            AccountKey20 {
+                                network: Some(EthereumNetwork::get()),
+                                key: token_address.into(),
+                            },
+                        ]
+                        .into()),
+                    }
+                };
+
+                if let None = pallet_foreign_asset_creator::ForeignAssetToAssetId::<Runtime>::get(
```
girazoki, line 318: I do prefer if let Some here

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,179 @@ where
...
+            return false;
+        }
+
+        // Try decoding the message
+        match VersionedXcmMessage::decode_all(&mut envelope.payload.as_slice()) {
+            Ok(VersionedXcmMessage::V1(MessageV1 {
+                command:
+                    Command::SendToken {
+                        token: token_address,
+                        destination: Destination::AccountId32 { id: _ },
```
girazoki, line 292: Do I assume that destination will be different in the case of sending it to a para-id? does this mean that it could be expanded later on? if so I would try put common code between `can_process` and `process`.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,179 @@ where
...
+                                network: None,
+                                id: destination_account,
+                            }],
+                        ),
+                    },
+                ]);
+
+                let ethereum_location = EthereumLocation::get();
+
+                let weight = XcmWeigher::weight(&mut xcm)
```
girazoki, line 391: we need a way to charge for this weight to the relayer or at least it is accounted. How to make it be accounted is something that I am not sure how to do because it depends on the pallet, and we probably cannot benchmark it for this worst case scenario. We can assume the send() weight might be sufficient to cover for this though, but let's think about it

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -362,6 +541,13 @@ impl snowbridge_pallet_inbound_queue::Config for Runtime {
     type MessageProcessor = (
         SymbioticMessageProcessor<Self>,
```
girazoki, line 542: Are we sure symbiotic message processor does not pick this command before the `EthTokensLolcaProcessor` does?

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -518,6 +523,367 @@ fn receive_native_tokens_from_eth_happy_path() {
...
+                proof: mock_snowbridge_message_proof(),
+            };
+
+            let balance_before = ForeignAssets::balance(asset_id, AccountId::from(BOB));
+            assert_eq!(balance_before, 0u128);
+
+            assert_ok!(EthereumInboundQueue::submit(relayer, message));
+
+            let balance_after = ForeignAssets::balance(asset_id, AccountId::from(BOB));
+            assert_eq!(balance_after, amount_to_transfer);
```
girazoki, line 704: Is the relayer rewarded? if so can we add a test? (it might not be becausde there is no tanssi tokens in the feeAccount, but if you add some it might be)

girazoki (approved):
Overall looks good, but what I would do is anticipate ourselves with the future us and try to put everything into cleaner functions (e.g., `convert_eth_token_transfer_with_local_dest`)

However, very good job @Agusrodri

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,179 @@ where
...
 }
 
+/// `EthTokensLocalProcessor` is responsible for receiving and processing the ETH native
+/// token and ERC20s coming from Ethereum with Tanssi chain as final destination.
+pub struct EthTokensLocalProcessor<T, XcmProcessor, XcmWeigher, EthereumLocation, EthereumNetwork>(
+    PhantomData<T>,
+    PhantomData<XcmProcessor>,
+    PhantomData<XcmWeigher>,
+    PhantomData<EthereumLocation>,
+    PhantomData<EthereumNetwork>,
```
tmpolaczyk, line 252: You can use a single PhantomData<(A, B, C)>, if you want

chains/orchestrator-relays/runtime/dancelight/src/xcm_config.rs
```diff
@@ -38,22 +42,23 @@ use {
...
+        ChildParachainConvertsVia, ConvertedConcreteId, DescribeAllTerminal, DescribeFamily,
+        FixedWeightBounds, FrameTransactionalProcessor, FungibleAdapter, FungiblesAdapter,
+        HashedDescription, IsChildSystemParachain, IsConcrete, MintLocation, NoChecking,
+        OriginToPluralityVoice, SendXcmFeeToAccount, SignedAccountId32AsNative,
+        SignedToAccountId32, SovereignSignedViaLocation, TakeWeightCredit, TrailingSetTopicAsId,
+        UsingComponents, WeightInfoBounds, WithComputedOrigin, WithUniqueTopic,
+        XcmFeeManagerFromComponents,
     },
-    xcm_executor::XcmExecutor,
+    xcm_executor::{traits::JustTry, XcmExecutor},
```
tmpolaczyk, line 61: I believe JustTry has been renamed to TryConvertInto

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,179 @@ where
...
+        }
+    }
+
+    fn process_message(_channel: Channel, envelope: Envelope) -> DispatchResult {
+        // - Decode payload as SendToken
+        let message = VersionedXcmMessage::decode_all(&mut envelope.payload.as_slice())
+            .map_err(|_| DispatchError::Other("unable to parse the envelope payload"))?;
+
+        match message {
+            VersionedXcmMessage::V1(MessageV1 {
```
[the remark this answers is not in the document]
Agusrodri: Sure! What you have in mind is to also to manage the para transfers inside this processor right? In that case it makes sense to have several functions yes, initially I thought that we would have different processors for each kind of transfer (local and para), but it makes sense to generalize.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: At the end I refactored it to return the result of `is_some()` accordingly.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: Done :)

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
Agusrodri: Yes it's being rewarded! I added a few checks for this in 10b007709e52c848a57722ede088bedb49513d52.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: Mmm yes I agree, we should somehow ensure we are accounting for it. I'll be investigating different options.

Agusrodri:
@girazoki I just pushed an optimized version of the code, creating a common function `decode_message_for_eth_transfer` which is used at the moment of retrieving necessary info that is common to both `can_process_message` and `process_message`.

The purpose of the other `process_xcm_local_native_eth_transfer` function is to specifically process the local native transfer. In the future, once we implement transfers for container chains, we can easily add a new function and manage the proper `destination` branches inside the main `process_message` function of the `EthTokensLocalProcessor`.

Let me know if this looks better, otherwise I'm happy to continue improving it :)

primitives/xcm-commons/src/lib.rs
```diff
@@ -57,3 +58,37 @@ impl frame_support::traits::ContainsPair<Asset, Location> for NativeAssetReserve
...
+            None
+        };
+
+        // Origin must match the EthereumLocation reserve
+        if let Some(ref reserve) = reserve {
+            if reserve == origin {
+                return true;
+            }
+        }
+        false
```
tmpolaczyk, line 92: This function can be shorter if we check the origin first. Unless we need to support more origins in the future (then this refactor doesn't help)

```rust
if origin != EthereumLocation::get() {
    return false;
}
match (asset.id.0.parents, asset.id.0.first_interior()) {
    (1, Some(GlobalConsensus(network))) if network == EthereumNetwork::get() => true,
    _ => false,
}
```

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,192 @@ where
...
+    EthTokensLocalProcessor<T, XcmProcessor, XcmWeigher, EthereumLocation, EthereumNetwork>
+where
+    T: frame_system::Config,
+    XcmProcessor: ExecuteXcm<T::RuntimeCall>,
+    XcmWeigher: WeightBounds<T::RuntimeCall>,
+    EthereumLocation: Get<Location>,
+    EthereumNetwork: Get<NetworkId>,
+{
+    /// Retrieve the eth transfer data from the message payload.
+    fn decode_message_for_eth_transfer(payload: &mut &[u8]) -> Option<EthTransferData> {
```
tmpolaczyk, line 337: Does it work if you do this?

```suggestion
    fn decode_message_for_eth_transfer(mut payload: &[u8]) -> Option<EthTransferData> {
```

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -236,6 +242,192 @@ where
...
+        match VersionedXcmMessage::decode_all(payload) {
```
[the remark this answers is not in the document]
Agusrodri: From what I can see, currently magic bytes are only used inside the `can_process_message` check in the `SymbioticMessageProcessor`: https://github.com/moondance-labs/tanssi/blob/master/primitives/bridge/src/symbiotic_message_processor.rs#L73

They are not being used inside the (already existing) `NativeTokenTransferMessageProcessor`.

Should we add the check to all processors or is it something that only needs to be validated at the moment of receiving the validators?

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: Yes it does, applied in db34d4d7b71a1ac4d3df49ce526d71f0c61645b6.

primitives/xcm-commons/src/lib.rs
Agusrodri: Done in df03945e159739ddab5ee8794a139c94fe650790. Thanks for the suggestion :)

tmpolaczyk (approved):

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: I've been analyzing this and we could do the same as we do in other parts of our code to account for extra weight:

```rust
frame_system::Pallet::<T>::register_extra_weight_unchecked(weight, DispatchClass::Mandatory);
```
This will add the extra weight based on our benchmarked instructions. I tested it and with the current weight values I have in this branch, it gives a total extra weight of `Weight { ref_time: 161196000, proof_size: 3593 }`.

This value shouldn't change a lot overtime, as we always use the same two instructions (`ReserveAssetDeposited` and `DepositAsset`).

What do you think? (cc @tmpolaczyk)

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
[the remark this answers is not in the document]
Agusrodri: As discussed internally, the purpose of the magic bytes is to ensure we are properly decoding our command instead of a snowbridge one while receiving validators.

We don't need this in the other processors, as we use snowbridge's commands directly.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
tmpolaczyk: Not sure, can `register_extra_weight_unchecked` fail? Because we never check if we have enough weight. So this will prevent executing more extriniscs, but if it's run in on_initialize we need some other way to ensure we don't use too much weight.

Agusrodri: At the end we concluded it's fine to leave this `register_extra_weight_unchecked` function for now. Will open another PR to generalize the benchmarking of a `WorstCaseProcessor` helper to have the weight that is used only by the worst processor we have, and add it to the weight of `submit()` extrinsic.

PR #1035, opened 2025-05-15
[MD-1043] Benchmarks testing for all runtimes using CI server

evgeny-s:
**Description**:
This PR adds the new pipeline that runs on the weekend: builds the master branch, runs the script that checks the weights anomaly, and in case of any, creates a PR ([example](https://github.com/moondance-labs/tanssi/pull/1037)) and assigns the team to review it.


.github/workflows/test-all-benchmarks-schedule.yml
```diff
@@ -0,0 +1,91 @@
+name: Test all benchmarks on schedule
+
+on:
+  push:
+#  TODO: Remove
+    branches:
+      - MD-1043-evgeny-benchmarks
```
evgeny-s, line 7: remove before merge

.github/workflows/test-all-benchmarks-schedule.yml
```diff
@@ -0,0 +1,91 @@
...
+      RUSTC_WRAPPER: "sccache"
+      CARGO_INCREMENTAL: "0"
+      SCCACHE_CACHE_SIZE: "100GB"
+      SCCACHE_GHA_ENABLED: "true"
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: MD-1043-evgeny-benchmarks # TODO Change to master before merge
```
evgeny-s, line 27: replace to master before merge

.github/workflows/test-all-benchmarks-schedule.yml
```diff
@@ -0,0 +1,91 @@
...
+          git config user.email "github-actions@github.com"
+          git checkout -b "$BRANCH"
+          git add .
+          git commit -m "chore(weights): update weights after benchmark failure"
+          git push origin "$BRANCH"
+
+          ~/bin/bin/gh pr create \
+            --title "chore(weights): auto-update weights on benchmark failure" \
+            --body "Auto-generated PR with new weights after benchmark failure." \
+            --base MD-1043-evgeny-benchmarks \
```
evgeny-s, line 86: replace to master before merge

.github/workflows/test-all-benchmarks-schedule.yml
```diff
@@ -0,0 +1,92 @@
...
+    branches:
+      - MD-1043-evgeny-benchmarks
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 6"  # Runs every Saturday at 12 AM UTC
+
+jobs:
+  test-all-benchmarks:
+    runs-on: self-hosted
+    timeout-minutes: 720
```
evgeny-s, line 15: 12 hours not enough, try to increase to 18hours

tmpolaczyk: This is still 12 hours, we could double it to 24h

tools/benchmarking.sh
```diff
@@ -5,17 +5,19 @@
 # The tanssi binary is required to be compiled with --features=runtime-benchmarks
 # in release mode.
 
-set -e
+set -euo pipefail
 
 # By default we use the tanssi-node release binary
 # However we can use any binary by running the benchmark tool with
 # BINARY=./target/release/container-chain-simple-node ./tools/benchmarking.sh
+: "${BINARY:=}"
```
tmpolaczyk, line 13: What is this?

evgeny-s: Previously, if benchmarking sh script is failed, the script did not fail on CI. This is why I've changed the parameter for this sh file from: 
`set -e`
to 
`set -euo pipefail`
Since we run the command in the pipe, it means that if the pipeline fails, it will fail the whole script, which is expected.

We didn't properly handle the empty parameter for both BINARY and CHAIN options, and the script was not failing because it was just ignored. Now - it is handled, and in case any of the option is not provided, it will work properly.

tmpolaczyk (approved):
Approve, but is the 12 hours timeout enough?

evgeny-s:
> Approve, but is the 12 hours timeout enough?
Let's see. If we run it in parallel, it shouldn't take more. Let's see the results after the next weekend.

.github/workflows/test-all-benchmarks-schedule.yml
evgeny-s: It is an outdated comment.
We should be good if we run it in parallel.

PR #1036, opened 2025-05-16
add new smoke tests dancelight

girazoki:
Adds some new smoke tests that were present in dancebox and were not prsent in dancelight.

### DIFF
--- a/test/suites/smoke-test-dancelight/test-invulnerables-priority.ts
+++ b/test/suites/smoke-test-dancelight/test-invulnerables-priority.ts
@@ -28,12 +28,12 @@ describeSuite({
                 // there should be no collator in orchestrator
                 expect(
                     currentAssignment.orchestratorChain.toHuman(),
-                    `In tanssi-solo there should be no collator assigned to orchestrator`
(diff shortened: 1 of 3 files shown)


Agusrodri (approved):

evgeny-s (commented):
Could you please share:
- Why do we find collators differently in comparison with dancebox
- Also, I see that for example: `test-staking-consistency` can be fully re-used for both chains. Do we have a folder for both dancebox/dancelight, so we can avoid duplication here?

girazoki:
> Could you please share:
> 
> * Why do we find collators differently in comparison with dancebox
> * Also, I see that for example: `test-staking-consistency` can be fully re-used for both chains. Do we have a folder for both dancebox/dancelight, so we can avoid duplication here?

-> To your first question, in dancebox there are no validators and we store collators in session>validators. in tanssi we dancelight we cannot do that, as we need validators in session->validators. therefore we need to count collators from their assignation to chains.

-> I can build a everything-but-flashbox smoke test folder, but I am not sure if it's worth, what do u think

tmpolaczyk (approved):

PR #1039, opened 2025-05-19
[MD-1018] Deduplicate Dancebox runtime common helpers for integration tests

evgeny-s:
**Description**:
The goal of this PR is to deduplicate the Dancebox integration tests' common helpers, which we have inside the runtime crate and XCM integration tests.

**What's done:**
- Moved dancebox common helpers to a separate crate
- Moved dancebox integration tests outside the dancebox runtime
- Renamed integration tests


evgeny-s:
A lot of changes have been introduced after 2503 in moved files, so I'll replicate it from the current master.

PR #1048, opened 2025-05-22
Enter maintenance mode if multi block migration fails

tmpolaczyk:
And remove already executed multi block migrations


tmpolaczyk:
zombie upgrade jobs failing because the job uses the latest release instead of the on chain release, and in the latest release the pallet version is old. For some unknown reason, even if the on chain pallet version is 1, as stored in storage, it turns into 0 if the runtime says that the latest version for the pallet is 0.

evgeny-s (approved):

tmpolaczyk:
And benchmarks are failing because we added one new event when migrations fail, and the pallet benchmarks dont like it

```
2025-05-28 09:37:36 [ 75 % ] Starting benchmark: pallet_multiblock_migrations::exec_migration_fail    
2025-05-28 09:37:36 Migration Some(0) failed - activating maintenance mode and continuing    
2025-05-28 09:37:36 panicked at /home/gh-actions/.cargo/git/checkouts/polkadot-sdk-df3be1d6828443a1/91d714a/substrate/frame/system/src/lib.rs:2019:9:
assertion `left == right` failed: expected event RuntimeEvent::MultiBlockMigrations(<wasm:stripped>) is not equal to the last event RuntimeEvent::MaintenanceMode(<wasm:stripped>)
  left: RuntimeEvent::MaintenanceMode(<wasm:stripped>)
 right: RuntimeEvent::MultiBlockMigrations(<wasm:stripped>)    
2025-05-28 09:37:36 Error executing and verifying runtime benchmark: Invalid input: Could not call runtime API to dispatch a benchmark: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
```

PR #1049, opened 2025-05-23
Benchmarking CI script fix

evgeny-s:
**Description**:
Removed unneeded files from the commit list

### DIFF
--- a/.github/workflows/test-all-benchmarks-schedule.yml
+++ b/.github/workflows/test-all-benchmarks-schedule.yml
@@ -92,6 +92,11 @@ jobs:
             echo "No changes to commit, skipping PR creation."
             exit 0
           fi
+          
+          rm -rf binaries
+          rm -rf mold
+          rm -rf runtime_size_report.md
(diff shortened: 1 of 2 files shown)


tmpolaczyk (approved):

.github/workflows/test-all-benchmarks-schedule.yml
```diff
@@ -51,6 +51,12 @@ jobs:
         run: |
           ./tools/benchmark-all-runtimes.sh --parallel
 
+      - name: Upload benchmark logs
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmark-logs
+          path: tmp/benchmark_log_*.txt
```
tmpolaczyk, line 58: Does this work if the previous step fails? Or does it need if: always()

evgeny-s: Yes, you are correct, we will miss logs in case of failure.
Added if always. 
Thanks!

PR #1054, opened 2025-05-26
Send tokens back to Ethereum (dancelight)

nanocryk:
Re-implement code from Snowbridge compatible with Tanssi being a relaychain instead of a parachain, and add it in dancelight.


primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,557 @@
...
+				log::error!(target: "xcm::ethereum_blob_exporter", "could not get global consensus from universal source '{universal_source:?}'.");
+				SendError::NotApplicable
+			})?;
+
+        if Ok(local_net) != universal_location.global_consensus() {
+            log::trace!(target: "xcm::ethereum_blob_exporter", "skipped due to unmatched relay network {local_net:?}.");
+            return Err(SendError::NotApplicable);
+        }
+
+        let source_location = Location::new(1, local_sub.clone());
```
ParthDesai, line 127: I think we need to check value of local sub here. Like the original ethereum blob exporter does.  https://github.com/paritytech/polkadot-sdk/blob/36c3039007866480be258d9645626c56ab168d69/bridges/snowbridge/primitives/outbound-queue/src/v1/converter/mod.rs#L94

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,557 @@
...
+    }
+
+    /// Convert the xcm for Polkadot-native token from AH into the Command
+    /// To match transfers of Polkadot-native tokens, we expect an input of the form:
+    /// # ReserveAssetDeposited
+    /// # ClearOrigin
+    /// # BuyExecution
+    /// # DepositAsset
+    /// # SetTopic
+    fn make_mint_foreign_token_command(
```
Agusrodri, line 385: Just a note: this function will be used only when we start receiving tokens from containers. We can either add it now or later. In my opinion it's better to include it now and modify it in the future if necessary.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,557 @@
...
+
+    fn network_matches(&self, network: &Option<NetworkId>) -> bool {
+        if let Some(network) = network {
+            *network == self.ethereum_network
+        } else {
+            true
+        }
+    }
+
+    /// Convert the xcm for Polkadot-native token from AH into the Command
```
Agusrodri, line 378:
```suggestion
    /// Convert the xcm for Polkadot-native token from container chains into the Command
```

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,557 @@
...
+            .any(|asset| !deposit_assets.matches(asset))
+        {
+            return Err(FilterDoesNotConsumeAllAssets);
+        }
+
+        // We only support a single asset at a time.
+        ensure!(reserve_assets.len() == 1, TooManyAssets);
+        let reserve_asset = reserve_assets.get(0).ok_or(AssetResolutionFailed)?;
+
+        // Fees are collected on AH, up front and directly from the user, to cover the
```
Agusrodri, line 446:
```suggestion
        // Fees are collected on the origin chain (containers), up front and directly from the user, to cover the
```

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,557 @@
...
+
+        // // Ignore messages that don't send assets.
+        // // TODO: Is it necessary? Bridges don't filter?
+        // let &[Instruction::WithdrawAsset(_), Instruction::ClearOrigin, Instruction::DepositAsset { .. }, ..] =
+        //     &xcm.inner()
+        // else {
+        //     // We need to make sure that msg is not consumed in case of `NotApplicable`.
+        //     *msg = Some(xcm);
+        //     return Err(NotApplicable);
+        // };
```
Agusrodri, line 527: Probably this is not necessary, as we can verify the message's integrity using `XcmConverter` functions.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
Agusrodri, line 127: As we are sending directly from Tanssi, I believe we should use parents 0 here:

```suggestion
        let source_location = Location::new(0, local_sub.clone());
```

nanocryk: Do you have more context on what `source_location` and `agent_id` obtained from it mean?

Agusrodri: I just checked again and parents 1 is correct, because it's the Ethereum chain the one which will interpret the message, so you can ignore my previous comment.

From what I understand, the `source_location` is the chain from which we are sending the message and the `agent_id` is the corresponding identifier of the snowbridge agent on Ethereum for that chain.

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+                Box::new(beneficiary_location.into()),
+                Box::new(vec![eth_asset].into()),
+                0u32,
+                Unlimited,
+            ));
+
+            // Correct amount has been sent
+            assert_eq!(ForeignAssets::balance(asset_id, &AccountId::from(BOB)), 10);
+
+            // Check fees have been payed (TODO: expect specific amount?)
```
evgeny-s, line 1431: Do not forget to either update the comment or specify the fee amount in the test.

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+            };
+
+            let eth_asset = AssetId(erc20_asset_location.clone())
+                .into_asset(Fungibility::Fungible(amount_to_transfer));
+
+            let balance_before = Balances::balance(&AccountId::from(BOB));
+
+            assert_ok!(XcmPallet::transfer_assets(
+                RuntimeOrigin::signed(AccountId::from(BOB)),
+                Box::new(EthereumLocation::get().into()),
```
evgeny-s, line 1421: Shouldn't we use here `VersionedLocation` with V3 or something, also the `parents` should be 1, since we need 1 level up from the relay.

nanocryk: I think the `.into()` does the conversion to V3.
`EthereumLocation` is defined in the constants as `(1, [GlobalConsensus(NetworkId { chain_id: ... })])`, so already with parents 1. Are you speaking about another `parents` ?

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
nanocryk: Currently I just check the balance of native tokens decreased and thus some fees have been payed. I currently don't know how to get the expected fee.

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
evgeny-s: Nevermind, I see that this is a Location

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+                erc20_asset_location.clone(), // Use the ERC20 location
+                asset_id,
+                AccountId::from(ALICE),
+                true,
+                1
+            ));
+
+            // Give tokens to user as if tokens were bridged before + extra to check the correct
+            // amount is sent
+            ForeignAssets::mint_into(asset_id, &AccountId::from(BOB), amount_to_transfer + 10)
```
girazoki, line 1398: is minting by any other ways other than receiving the tokens prohibited? I am assuming so

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+                parents: 0,
+                interior: X1([AccountKey20 {
+                    network: Some(EthereumNetwork::get()),
+                    key: beneficiary_address.into(),
+                }]
+                .into()),
+            };
+
+            let eth_asset = AssetId(erc20_asset_location.clone())
+                .into_asset(Fungibility::Fungible(amount_to_transfer));
```
girazoki, line 1415: I had no idea `into_asset` existed!

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+                    index: 90,
+                    error: [24, 0, 0, 0],
+                    message: Some("LocalExecutionIncomplete")
+                })
+            );
+        })
+}
+
+#[test]
+fn cant_send_eth_native_token_more_than_owned() {
```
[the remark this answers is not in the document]
girazoki: And potentially, add also a test where we pass a location different to the ethereum locaiton, which should give you something like non-valid destination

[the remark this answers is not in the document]
girazoki: Also all these should leave a trace in the form of an event in EthereumOutboundQueue. please check it up

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1338,6 +1341,267 @@ fn test_pricing_parameters() {
...
+            (EthereumSovereignAccount::get(), 100_000 * UNIT),
+            (SnowbridgeFeesAccount::get(), 100_000 * UNIT),
+            (AccountId::from(ALICE), 100_000 * UNIT),
+            (AccountId::from(BOB), 100_000 * UNIT),
+        ])
+        .build()
+        .execute_with(|| {
+            // Setup bridge and token
+            let channel_id: ChannelId = ChannelId::new(hex!(
+                "00000000000000000000006e61746976655f746f6b656e5f7472616e73666572"
```
girazoki, line 1450: where does this channel-id come from? is it random?

nanocryk: Copied this value from other tests already present. Seems to come from a PR of yours @dimartiro , can you give us some insight?

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        let expected_asset_id = ConvertAssetId::convert(&token_id).ok_or(InvalidAsset)?;
+
+        ensure!(asset_id == expected_asset_id, InvalidAsset);
+
+        // Check if there is a SetTopic and skip over it if found.
+        let topic_id = match_expression!(self.next()?, SetTopic(id), id).ok_or(SetTopicExpected)?;
+
+        Ok((
+            Command::MintForeignToken {
```
evgeny-s, line 487: Shouldn't we call here [TransferNativeToken](https://github.com/moondance-labs/snowbridge/blob/7818f9cbf81b21ad1d8703570b816bbde64f760d/contracts/src/Gateway.sol#L491) instead?

evgeny-s: Seems like this command for foreign fungible tokens, not ETH native

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,554 @@
...
+        // transfer amount must be greater than 0.
+        ensure!(amount > 0, ZeroAssetTransfer);
+
+        // Check if there is a SetTopic and skip over it if found.
+        let topic_id = match_expression!(self.next()?, SetTopic(id), id).ok_or(SetTopicExpected)?;
+
+        Ok((
+            Command::AgentExecute {
+                agent_id: self.agent_id,
+                command: AgentExecuteCommand::TransferToken {
```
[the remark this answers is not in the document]
Agusrodri: Agree, we should use `TransferNativeToken` instead.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        // TODO: Support source being a parachain.
+        if !matches!(local_sub, Junctions::Here) {
+            log::trace!(target: "xcm::ethereum_blob_exporter", "skipped due to unmatched sub network {local_sub:?}.");
+            return Err(SendError::NotApplicable);
+        }
+
+        let source_location = Location::new(1, local_sub.clone());
+
+        let agent_id = match AgentHashedDescription::convert_location(&source_location) {
```
Agusrodri, line 135: I believe here we should use the same `agent_id` that we have stored for token transfers in `EthereumTokenTransfers` pallet. Let's see if we can fetch it from there.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
Agusrodri: I believe `MintForeignToken` is the correct one here, as we are dealing with container native assets coming directly from the source chain.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,554 @@
...
+                fun: Fungible(amount),
+            } => Some((inner_location.clone(), *amount)),
+            _ => None,
+        }
+        .ok_or(AssetResolutionFailed)?;
+
+        // transfer amount must be greater than 0.
+        ensure!(amount > 0, ZeroAssetTransfer);
+
+        let token_id = TokenIdOf::convert_location(&asset_id).ok_or(InvalidAsset)?;
```
Agusrodri, line 477: Instead of `TokenIdOf`, we could use `EthereumSystem` pallet to fetch the `token_id` directly from there, and that would imply the token needs to be properly registered on Ethereum before receiving it here.

We do this in our `EthereumTokenTransfers` pallet already for the Tanssi native token: https://github.com/moondance-labs/tanssi/blob/master/pallets/ethereum-token-transfers/src/lib.rs#L210

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,554 @@
...
+    fn validate(
+        dest: &mut Option<Location>,
+        msg: &mut Option<Xcm<()>>,
+    ) -> SendResult<Self::Ticket> {
+        let universal_source = UniversalLocation::get();
+
+        // TODO: Better value?
+        // This channel value is ignored anyway (like in original snowbridge code).
+        // Should we not ignore it and actually use this value instead?
+        let channel = 0;
```
Agusrodri, line 518: In my opinion it's cleaner to add the `BridgeChannelId` type on this router, and pass its value directly to `validate_export`, instead of fetching it inside the exporter.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
nanocryk: This is code regarding tokens from containers, which is outside of the scope of this PR. Code is copied from snowbridge and modified in some places, and in a latter PR this function will be modified too. Adding a comment about that.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
Agusrodri: Yeah I see in some places they just ignore the param and put 0, so let's do the same.

I would remove the initial comment and add another one explaining that we ignore the param as we later fetch it using the `BridgeChannelId` type instead.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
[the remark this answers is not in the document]
Agusrodri: This piece of the code is actually part of the scope of this PR, as this function is used to build the Snowbridge command when sending ETH tokens from Tanssi. For container tokens, the `make_mint_foreign_token_command` function is used instead.

So let's make sure we are using the right command here. In my opinion is better to introduce `TransferNativeToken` directly.

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
nanocryk: Does it now match your expectation?

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,558 @@
...
+            .any(|asset| !deposit_assets.matches(asset))
+        {
+            return Err(FilterDoesNotConsumeAllAssets);
+        }
+
+        // We only support a single asset at a time.
+        ensure!(reserve_assets.len() == 1, TooManyAssets);
+        let reserve_asset = reserve_assets.get(0).ok_or(AssetResolutionFailed)?;
+
+        // Fees are collected on the origin chain (container chain), up front and directly from the user, to cover the
```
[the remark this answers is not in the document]
nanocryk: Better?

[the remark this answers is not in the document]
Agusrodri: Perfect!

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1672,6 +1678,479 @@ fn test_pricing_parameters() {
...
+                ))
+                .count(),
+                0,
+                "MessageQueued event should NOT be emitted!"
+            );
+        })
+}
+
+#[test]
+fn cant_send_tokens_of_unknown_remote_network() {
```
girazoki, line 2059: Can we add a test in which we send a message not related to to ethereum token transfers (through pallet-xcm->send), I dont know something like a clearOrigin and see what happens?

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    /// Convert the xcm for Polkadot-native token from the origin chain (container chain) into the Command
+    /// To match transfers of Polkadot-native tokens, we expect an input of the form:
+    /// # ReserveAssetDeposited
+    /// # ClearOrigin
+    /// # BuyExecution
+    /// # DepositAsset
+    /// # SetTopic
+    fn make_mint_foreign_token_command(
```
girazoki, line 384: I would for now perhaps error here. FIrst because it's untested (we have another way of transferring the tanssi tokens), so if we are going to open another path for this, I would like to be tested first. what do youy think?

girazoki: well in any case this is only goint to dancelight so I am fine with it. but let's create another issue please for adding tests for this route

primitives/bridge/src/snowbridge_outbound_token_transfer.rs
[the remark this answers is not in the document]
Agusrodri: Mmm not sure if there is an easy way to test that through integration tests, it might be complex because we don't have an easy way to fetch the command sent through the outbound-queue. 

Although, if you want you can create a unit test in primitives testing the `XcmConverter` directly. I believe that should be enough.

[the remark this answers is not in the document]
nanocryk: Done with the expected XCM message. Do you want more of them?

girazoki (approved):

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1670,6 +1676,686 @@ fn test_pricing_parameters() {
...
+                para_id
+            ));
+
+            // Define a mock ERC20 token address
+            let token_address = H160(hex!("deadbeefdeadbeefdeadbeefdeadbeefdeadbeef"));
+
+            let erc20_asset_location = Location {
+                parents: 1,
+                interior: X2([
+                    GlobalConsensus(NetworkId::BitcoinCore),
```
ParthDesai, line 2184: The network should not be `BitcoinCore` here as we need to have everything correct except the condition we are expecting to be checked.

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -1670,6 +1676,686 @@ fn test_pricing_parameters() {
...
```
ParthDesai, line 2282: The network should not be `BitcoinCore` here as we need to have everything correct except the condition we are expecting to be checked.

PR #1056, opened 2025-05-26
Fix pallet_collator_assignment randomness in relay chain mode

tmpolaczyk:
should_end_session cannot be implemented in relay chain because it is not possible to know for sure if the next block will be in the same session as the current one, as it depends on slots and validators can skip slots.


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -3513,22 +3513,9 @@ impl Get<u32> for ConfigurationCollatorRotationSessionPeriod {
...
-        n != 1 && {
-            let diff = Babe::current_slot()
-                .saturating_add(1u64)
-                .saturating_sub(Babe::current_epoch_start());
-            *diff >= Babe::current_epoch().duration
-        }
-    }
-
-    fn get_randomness() -> [u8; 32] {
+pub struct BabeGetCollatorAssignmentRandomness;
```
girazoki, line 3516: let's try to put common structs in common places. in this case both starlight and dancelight use them, so likely we can use somewhere a common crate for these

pallets/collator-assignment/src/lib.rs
```diff
@@ -717,3 +711,88 @@ impl<BlockNumber> GetRandomnessForNextBlock<BlockNumber> for () {
...
+    fn take_randomness() -> [u8; 32] {
+        Randomness::<Runtime>::take()
+    }
+}
+
+/// Relaychain randomness.
+///
+/// Uses current block randomness. This randomness exists in `on_initialize` so we don't need to
+/// `prepare_randomness` in the previous block.
+pub struct RelaychainRandomness<T>(PhantomData<T>);
```
girazoki, line 776: I would prefer to name this solo-chain or something ssimilar

test/suites/smoke-test-common-solo/test-randomness-consistency.ts
```diff
@@ -48,9 +57,13 @@ describeSuite({
                 const blockToCheck = blockBabeEpochStart - 1;
                 const apiBeforeNewSession = await api.at(await api.rpc.chain.getBlockHash(blockToCheck));
 
-                // Just before, the randomness was not empty
-                const randomnessBeforeSession = await apiBeforeNewSession.query.tanssiCollatorAssignment.randomness();
-                expect(randomnessBeforeSession.isEmpty).to.not.be.true;
+                // After runtime upgrade, randomness in storage is always empty
```
girazoki, line 60: the comment is wrong

girazoki (approved):

PR #1066, opened 2025-05-30
Refactor: merge the 3 download-runtime scripts into one

tmpolaczyk:
Attempt to simplify the scripts because they share some logic, although the result is more complex than the original scripts.

And centralize chmod in all workflows to use a workflow template.

The donwload-artifact action uses zip files to support multiple files. Zip files do not store file permissions, so they are lost when dowloading. Therefore, always after downloading binaries, we must chmod +x them.

There are alternative actions that handle this process automatically, by using a tar file instead of zip, but I thought that's more complex to add.


.github/workflows/release.yml
```diff
@@ -756,7 +733,10 @@ jobs:
                   path: target/release
             - name: "Install and run upgrade test"
               run: |
-                  chmod uog+x target/release/*
+                  chmod uog+x target/release/tanssi-node
+                  chmod uog+x target/release/tanssi-relay
+                  chmod uog+x target/release/tanssi-relay-execute-worker
+                  chmod uog+x target/release/tanssi-relay-prepare-worker
```
girazoki, line 739: arent you missing  the container-chain binaries here?

tmpolaczyk: Ah, so that's the issue. Will just put `target/release/*` then

chexware (approved):

.github/workflows/release.yml
```diff
@@ -924,6 +907,7 @@ jobs:
               uses: actions/checkout@v4
               with:
                   ref: ${{ needs.set-tags.outputs.git_ref }}
+            - name: "Download binaries"
             - uses: actions/download-artifact@v4
```
tmpolaczyk, line 911:
```suggestion
            - name: "Download binaries"
              uses: actions/download-artifact@v4
```

.github/workflows/release.yml
```diff
@@ -892,17 +875,17 @@ jobs:
...
                   path: target/release
+            - name: Make binaries executable
+              uses: ./.github/workflow-templates/make-binaries-executable
             - name: "Download runtime"
               uses: actions/download-artifact@v4
               with:
                   name: runtimes
                   path: target/release/wbuild/${{ matrix.chains.runtime }}-runtime/
+            - name: Make binaries executable
+              uses: ./.github/workflow-templates/make-binaries-executable
```
tmpolaczyk, line 886:
```suggestion
```

test/moonwall.config.json
```diff
@@ -788,10 +788,10 @@
             "name": "zombie_frontier_template_upgrade_with_latest_releases",
             "testFileDir": ["suites/zombie_simple_template_upgrade"],
             "runScripts": [
-                "download-latest-rt-binaries.sh",
+                "download-latest-rt-binaries.sh tanssi-node container-chain-frontier-node",
                 "build-spec-single-container.sh tmp container-chain-frontier-node flashbox-local",
                 "downloadPolkadot.ts",
-                "download-latest-binaries.sh ../target/release",
+                "download-latest-rt-binaries.sh --latest-client tanssi-node container-chain-frontier-node",
```
tmpolaczyk, line 794: This needs to download to ../target/release

PR #1070, opened 2025-06-02
Disabled treasury burn for *light chains

evgeny-s:
**Description**
- This PR disables Burn for the Treasury pallet
- Added TS integration tests to cover this case

### DIFF
--- a/chains/orchestrator-relays/runtime/starlight/src/lib.rs
+++ b/chains/orchestrator-relays/runtime/starlight/src/lib.rs
@@ -624,7 +624,6 @@ parameter_types! {
     pub const ProposalBondMaximum: Balance = 1 * GRAND;
     // We allow it to be 1 minute in fast mode to be able to test it
     pub const SpendPeriod: BlockNumber = runtime_common::prod_or_fast!(6 * DAYS, 1 * MINUTES);
-    pub const Burn: Permill = Permill::from_perthousand(2);
     pub const TreasuryPalletId: PalletId = PalletId(*b"py/trsry");
     pub const PayoutSpendPeriod: BlockNumber = 30 * DAYS;
     // The asset's interior location for the paying account. This is the Treasury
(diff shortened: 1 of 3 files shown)


chains/orchestrator-relays/runtime/starlight/src/lib.rs
```diff
@@ -682,7 +681,7 @@ impl pallet_treasury::Config for Runtime {
     type RejectOrigin = EitherOfDiverse<EnsureRoot<AccountId>, Treasurer>;
     type RuntimeEvent = RuntimeEvent;
     type SpendPeriod = SpendPeriod;
-    type Burn = Burn;
+    type Burn = ();
```
girazoki, line 684: I would even put () in the spend period. In the test we can leave it hardcoded, but there is not reason to leave a spend period that does nothing right?

evgeny-s: Do you mean something like this (`6 days` replaced by `()`)?
```
pub const SpendPeriod: BlockNumber = runtime_common::prod_or_fast!((), 1 * MINUTES);
```
I don't think we need to keep it even in the tests.

evgeny-s: The thing is, since we don't have a spend period, how to test that it is disabled..

evgeny-s: @girazoki , just checked.
If we set SpendingPeriod to (), it will trigger `treasury.Spending` for each block, which is not good, right?
So the test will be working, but I think better to keep it as is. 
WDYT?

tmpolaczyk: I think its better to keep a period similar to polkadot, to avoid bugs in the pallet because nobody ever used a different period. So just leave the period like it was before.

tmpolaczyk (approved):

PR #1072, opened 2025-06-03
Fix docker publish runtime draft

girazoki:
Fixes docker publish runtime draft job and adds 2 things:

The first, starlight is now taken into consideration depending on the name of the branch. if the branch contains the name starlight, then it will only build and push to docker tanssi-solochain executables. if the branch contains the name para, the opposite.

It also allows now to query this job for a particular branch manually, with a workflow dispatch


.github/workflows/docker-runtime-draft.yml
```diff
@@ -30,38 +53,77 @@ jobs:
...
+      - name: Save binary
         if: ${{ matrix.cpu == '' }}
         run: |
           mkdir -p binaries
-          cp target/production/tanssi-node binaries/tanssi-node
-          cp target/production/container-chain-frontier-node binaries/container-chain-frontier-node
-          cp target/production/container-chain-simple-node binaries/container-chain-simple-node
-      - name: Save parachain custom binary
+          if [[ "${{ needs.set-tags.outputs.git_branch }}" == *"para"* ]]; then
+            cp target/production/tanssi-node binaries/tanssi-node
```
evgeny-s, line 70: This can be combined to something like:
```
        suffix=""
        if [ "${{ inputs.cpu }}" != "" ]; then
          suffix="-${{ inputs.cpu }}"
        fi
        cp target/${{ inputs.profile }}/tanssi-node binaries/tanssi-node$suffix
```

.github/workflows/docker-runtime-draft.yml
```diff
@@ -74,7 +136,7 @@ jobs:
         id: prep
         run: |
           SHA8=$(git log -1 --format='%H' | cut -c1-8)
-          DOCKER_IMAGE=moondancelabs/${{matrix.image}}
+          DOCKER_IMAGE=moondancelabs/${{matrix.image.image}}
```
evgeny-s, line 130: And remove `image` from here as well.

PR #1078, opened 2025-06-06
Allow parachain manager to interact with services payment

nanocryk:
Currently only root was able to interact with sensitive calls of ServicesPayment. This PR allows the parachain manager to do so too.


chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -1675,6 +1675,37 @@ parameter_types! {
     pub const FreeCollatorAssignmentCredits: u32 = FreeBlockProductionCredits::get()/EpochDurationInBlocks::get();
 }
 
+pub struct EnsureParachainManager;
```
girazoki, line 1678: Very likely this should be both on dancelight and stagelight. And being that the case, I would probably this somewhere in common. Also tests

tmpolaczyk:
Merge master to fix ci

tmpolaczyk (approved):

PR #1085, opened 2025-06-10
Add trace logs to help debug snowbridge messages

tmpolaczyk:


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -496,7 +501,7 @@ impl<T> RewardProcessor<T> for RewardThroughFeesAccount<T>
 where
     T: snowbridge_pallet_inbound_queue::Config + pallet_ethereum_token_transfers::Config,
     T::AccountId: From<sp_runtime::AccountId32>,
-    <T::Token as Inspect<T::AccountId>>::Balance: From<u128>,
+    <T::Token as Inspect<T::AccountId>>::Balance: core::fmt::Display,
```
ParthDesai, line 504: Would it be better if this was Debug? As far more types implement Debug than Display and it seems appropriate for logging purpose.

PR #1106, opened 2025-06-25
Refactor common types within Dancelight and Starlight

nanocryk:
Move them in `tanssi-runtime-common` with feature `relay` so that other runtimes don't pull unnecessary dependencies.

Do you see other common types that are worth moving too?


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -538,43 +403,6 @@ mod test_helpers {
     }
 }
 
-/// Rewards the relayer that processed a native token transfer message
```
girazoki, line 541: this might conflict with @Agusrodri PR https://github.com/moondance-labs/tanssi/pull/1105

nanocryk: Merge conflict should be fixed :)

ParthDesai (approved):
LGTM

PR #1109, opened 2025-06-27
fix test ethereum token trannsfer

girazoki:
We always tried to decode inbound dispatches as token trasnfers, but its not correct as they might be validator receptions.


test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -171,7 +171,23 @@ describeSuite({
                                 eventLog.topics
                             );
 
-                            const versioned = api.registry.createType("VersionedXcmMessage", decodedEvent.payload);
+                            let versioned;
+                            try {
+                                versioned = api.registry.createType("VersionedXcmMessage", decodedEvent.payload);
```
tmpolaczyk, line 176: This way we always hide errors decoding message. I would prefer to check for the magic bytes of the message, and decide if its a validator update or a VersionedXcmMessage based on that.

```rust
pub const MAGIC_BYTES: [u8; 4] = [112, 21, 0, 56];
```

girazoki: me too, but that requiers a bunch of work that can be converted into another issue. I dont want alerts right now to mess up with us

girazoki: or if you want I can test only the first 4 bytes, if that sounds good. sorry I though you meant decoding the entire message

evgeny-s: I can add a quick thing that @tmpolaczyk mentioned, because I like it better than just try/catch thing and we can test it on both blocks.

test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -54,6 +54,7 @@ const customTypes = {
     },
     TokenId: "H256",
 };
+const MAGIC_BYTES = "0x70150038";
```
tmpolaczyk, line 57: Some docs for what is this

PR #1120, opened 2025-07-07
remove some todos throught the code, either because they are not relevant or are already done

girazoki:


chains/orchestrator-paras/runtime/dancebox/src/lib.rs
```diff
@@ -564,8 +564,6 @@ impl BabeCurrentBlockRandomnessGetter {
 
     /// Return the block randomness from the relay mixed with the provided subject.
     /// This ensures that the randomness will be different on different pallets, as long as the subject is different.
-    // TODO: audit usage of randomness API
```
girazoki, line 567: already aduited

evgeny-s (approved):

PR #1121, opened 2025-07-07
Add testing support for sending WETH to Tanssi

Agusrodri:
## Description

This PR adds support for testing the sending of WETH token from Ethereum to Tanssi via e2e tests.


test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -213,6 +242,26 @@ describeSuite({
                 ).stdout
             );
 
+            const ethereumNetwork = { Ethereum: { chainId: TESTNET_ETHEREUM_NETWORK_ID } };
+
+            wETHAssetId = 42;
```
evgeny-s, line 247: Maybe we can put it to const with a proper name? I saw we use it in several places in tests.

PR #1124, opened 2025-07-08
Use relaychain keystore for collators

tmpolaczyk:
Panic if user passes `--keystore-path` before the first `--`


evgeny-s (approved):
Maybe we can add a quick test for CLI run to make sure it works as expected.

PR #1126, opened 2025-07-08
Fix xcm foreign tokens config and use foreign token for benchmarks

dimartiro:
# Description

- Match dancelight config regarding xcm foreign tokens in starlight
- Use foreign token for xcm fungibles benchmarks

Inspired in [assets-hub benchmarks](https://github.com/paritytech/polkadot-sdk/commit/c4b8ec123afcef596fbc4ea3239ff9e392bcaf36)


chains/orchestrator-relays/runtime/starlight/src/lib.rs
```diff
@@ -3256,24 +3260,62 @@ sp_api::impl_runtime_apis! {
...
-                    Asset { fun: Fungible(1 * UNITS), id: AssetId(TokenLocation::get()) },
-                ));
-                pub TrustedReserve: Option<(Location, Asset)> = None;
+                pub TrustedReserve: Option<(Location, Asset)> = Some(
+                    (
+                        EthereumLocation::get(),
+                        Asset {
+                            id: AssetId(Location {
+                                parents: 1,
+                                interior: X1([GlobalConsensus(EthereumNetwork::get())].into()),
```
tmpolaczyk, line 3269: This is copied from dancelight, but wouldn't it be better to use EthereumLocation::get() directly?

Agusrodri: done :)

girazoki:
ok I tested and the test cannot work without us enabling the ethTokenPRocessor, which comes in https://github.com/moondance-labs/tanssi/pull/1162. so for me this is an approval

PR #1130, opened 2025-07-09
Fix rust version in srtool publish runtime job

tmpolaczyk:


tmpolaczyk:
Running again on a2a38bc https://github.com/moondance-labs/tanssi/actions/runs/16166789561

PR #1135, opened 2025-07-11
Ability to run publish-binary with a docker-uploaded image

girazoki:
For specific cases we might want the ability to publish a public release from a docker generated image. this new job does that by calling it like, e.g.,:

`gh workflow run .github/workflows/publish-binary-from-docker.yml -r=download-runtime-with-steroids -f binaries="tanssi-node tanssi-relay" -f tag_name=myrelease -f release_name=myrelease -f docker_tag=v0.13.0`

the binaries have to be provided with a white-space between them.

I had to modify the download-runtimes.sh to accept a couple of things:

- include-all-args to download all binaries for all architectures
- tag to be able to download from a specific docker tag

An example of this can be seen in draft releases

### DIFF
--- a/test/scripts/download-latest-rt-binaries.sh
+++ b/test/scripts/download-latest-rt-binaries.sh
@@ -260,7 +260,7 @@ if $DOWNLOAD_FRONTIER_NODE; then
     fi
     docker run --rm \
     --entrypoint tar \
-    "$SIMPLE_IMAGE" \
+    "$FRONTIER_IMAGE" \
     -C /container-chain-template-evm  -cf - \
       "${BINARIES[@]}" | tar -C $OUTPUT_PATH -xf -
   else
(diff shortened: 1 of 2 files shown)


test/scripts/download-latest-rt-binaries.sh
```diff
@@ -167,11 +212,25 @@ mkdir -p $OUTPUT_PATH
...
-      "$TANSSI_IMAGE" \
-      -C /tanssi -cf - tanssi-node \
-    | tar -C $OUTPUT_PATH -xf -
+    --entrypoint tar \
+    "$TANSSI_IMAGE" \
+    -C /tanssi -cf - \
+      "${BINARIES[@]}" | tar -C $OUTPUT_PATH -xf -
+
   chmod +x $OUTPUT_PATH/tanssi-node
   echo "→ $OUTPUT_PATH/tanssi-node"
```
tmpolaczyk, line 235: This could chmod and echo the real $BINARIES, although chmod is only needed for tests so it shouldn't matter

girazoki: that is why I left it like this...not important

tmpolaczyk (approved):

PR #1138, opened 2025-07-11
Add zombie_tanssi_relay_keystore test suite

tmpolaczyk:
Tests for #1124 

Do not run in CI, need to be run manually with

```
pnpm moonwall test zombie_tanssi_relay_keystore
```

### DIFF
--- a/test/suites/zombie_tanssi_relay_keystore/test_zombie_tanssi_relay_keystore.ts
+++ b/test/suites/zombie_tanssi_relay_keystore/test_zombie_tanssi_relay_keystore.ts
@@ -1,0 +1,286 @@
+import { beforeAll, describeSuite, expect } from "@moonwall/cli";
+import { generateKeyringPair } from "@moonwall/util";
+import { ApiPromise, Keyring, WsProvider} from "@polkadot/api";
+import type { Signer } from "ethers";
+import fs from "node:fs/promises";
+import {
+    chainSpecToContainerChainGenesisData,
+    checkLogsNotExist,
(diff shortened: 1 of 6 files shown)


test/suites/zombie_tanssi_relay_keystore/test_zombie_tanssi_relay_keystore.ts
```diff
@@ -0,0 +1,411 @@
...
+                const txBatch = relayApi.tx.utility.batchAll([tx1, tx2]);
+                await signAndSendAndInclude(relayApi.tx.sudo.sudo(txBatch), alice);
+            },
+        });
+
+        it({
+            id: "T04",
+            title: "Test assignation did not change",
+            test: async () => {
+                const currentSession = (await relayApi.query.session.currentIndex()).toNumber();
```
evgeny-s, line 180: We had already something similar in the previous test.
Maybe we want to extract these first 10 steps to the separate suite and run it for this and for the previous like a set of generic tests that we want to run for some zombienet tests?
Let's try to prevent duplication.

tmpolaczyk: How do I do that? The tests all have different timeouts, I can't just put a function call in beforeAll or in T01...

evgeny-s: Very quick example, something like this: https://github.com/moondance-labs/tanssi/pull/1197

PR #1166, opened 2025-07-18
Script that cancels other PR-s besides the specific one

evgeny-s:
**Description**:
Sometimes we need to speed up the specific PR merge. 
Using this script, we can cancel all the other PR jobs except the specified one. 
image

### DIFF
--- a/scripts/cancel_all_but.sh
+++ b/scripts/cancel_all_but.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+set -euo pipefail
+
+REPO="moondance-labs/tanssi"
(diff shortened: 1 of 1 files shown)


scripts/cancel_all_but.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
ParthDesai, line 1: Hmm. So, we need gh binary for this to work right?

evgeny-s: Yes, locally.
You also need to be logged in on Github via `gh cli`

girazoki (approved):

PR #1169, opened 2025-07-18
[MD-1197] Improve smoke ETH token transfer for ethereumInboundQueue

evgeny-s:
**Description**
Improve the error processing in case if the event payload is unknown


girazoki (approved):

tmpolaczyk:
Has this already happened in smoke tests? Or do we expect it to never happen?

PR #1173, opened 2025-07-22
Latest tanssi symbiotic commit

girazoki:
https://github.com/moondance-labs/tanssi-symbiotic/commit/1aef8e7c3b204dea46dc99b8656d67a1a4a67cee


test/scripts/bridge/build-relayer.sh
```diff
@@ -18,6 +18,7 @@ else
   git clone --recurse-submodules https://github.com/moondance-labs/tanssi-bridge-relayer $relayer_root_dir
   pushd $relayer_root_dir
   git fetch && git checkout $RELAYER_COMMIT
+  git submodule update --init --recursive
```
evgeny-s, line 21: Shouldn't be the case, there is git submodule command in place already for Makefile [here](https://github.com/moondance-labs/tanssi-symbiotic/blob/staging/Makefile#L30).

Agusrodri: From what I see, that file is related to the `tanssi-symbiotic` repo. We need this change to specifically update the submodules of `tanssi-bridge-relayer` repo while cloning it.

Agusrodri (approved):

PR #1175, opened 2025-07-23
Add ERC20 send receive smoke tests

dimartiro:
# Descriptions

Smoke tests for receiving and sending ERC20 tokens

### Inbound ERC20

We check that the command is `sendToken` then we check that we are minting the right amount of tokens in the destination account

### Outbound ERC20

If an `xcmPallet.transferAssets` extrinsic is executed we check that the destination is an ethereum account and then check that the tokens are burned from the signer account

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8972,7 +8972,7 @@ checksum = "43794a0ace135be66a25d3ae77d41b91615fb68ae937f904090203e81f755b65"
 
 [[package]]
 name = "node-common"
-version = "0.1.0"
+version = "0.15.0"
 dependencies = [
  "async-io 1.13.0",
  "async-trait",
@@ -9032,6 +9032,7 @@ dependencies = [
  "sp-session",
  "sp-timestamp",
  "sp-transaction-pool",
+ "substrate-build-script-utils",
  "substrate-frame-rpc-system",
  "substrate-prometheus-endpoint",
  "tp-traits",
--- a/chains/container-chains/nodes/frontier/src/command.rs
+++ b/chains/container-chains/nodes/frontier/src/command.rs
@@ -561,7 +561,10 @@ mod tests {
         // This is to verify we didn't forget to change one of them
         let v1 = ContainerChainCli::impl_version();
         let v2 = Cli::impl_version();
+        // Container chain nodes report the same version for relay chain side as the parachain side
+        let v3 = RelayChainCli::<NodeName>::impl_version();
 
         assert_eq!(v1, v2);
+        assert_eq!(v1, v3);
     }
 }
--- a/chains/container-chains/nodes/simple/src/command.rs
+++ b/chains/container-chains/nodes/simple/src/command.rs
@@ -50,7 +50,7 @@ pub struct NodeName;
 
 impl Get<&'static str> for NodeName {
     fn get() -> &'static str {
-        "Frontier"
+        "Simple"
     }
 }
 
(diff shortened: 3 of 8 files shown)


test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -225,6 +228,164 @@ describeSuite({
...
+                let currentBlock = (await api.rpc.chain.getBlock()).block.header.number.toNumber();
+
+                if (BLOCK_NUMBER_TO_DEBUG !== undefined) {
+                    BLOCKS_AMOUNT_TO_CHECK = 1;
+                    currentBlock = BLOCK_NUMBER_TO_DEBUG + 1;
+                }
+
+                for (let i = 1; i <= BLOCKS_AMOUNT_TO_CHECK; i++) {
+                    const blockNumber = currentBlock - i;
+                    process.stdout.write(`\rProcessing block [${blockNumber}]: ${i}/${BLOCKS_AMOUNT_TO_CHECK}`);
```
ParthDesai, line 248: Why not `console.log`?

test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -225,6 +228,164 @@ describeSuite({
                 }
             },
         });
+
+        it({
+            id: "C04",
```
evgeny-s, line 233: This test is very similar to C03, maybe we can just add another assertion there?
```
                            const matched = relatedEvents.some(({ event }) => {
                                const { section, method, data } = event;
                                if (section !== "balances" || method !== "Transfer") return false;

                                const [from, to, value] = data;

                                return (
                                    from.toString() === sovereignAccount &&
                                    to.toString() === destination.accountId32 &&
                                    value.toString() === amount.toString()
                                );
                            });

                            expect(
                                matched,
                                `Expected Transfer of ${amount.toString()} from sovereign to ${destination.accountId32} in block ${blockNumber}`
                            ).toBe(true);
```

test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -225,6 +228,164 @@ describeSuite({
...
+                            const signer = extrinsic.signer.toString();
+
+                            const [dest, _, assets] = extrinsic.args;
+
+                            const jsonDest = dest.toJSON();
+                            const versionKey = Object.keys(jsonDest)[0]; // e.g., "v3"
+                            const location = jsonDest[versionKey];
+
+                            // Is not a message to ethereum
+                            if (!location?.interior?.x1?.globalConsensus?.ethereum) continue;
```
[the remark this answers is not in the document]
evgeny-s: That's fine, I think moving forward better to avoid using `?.`. For now - let's keep it.

test/suites/smoke-test-common-solo/test-ethereum-token-transfers.ts
```diff
@@ -225,9 +225,173 @@ describeSuite({
...
+                            const { eventLog } = message.toJSON();
+
+                            const decodedEvent = iface.decodeEventLog(
+                                "OutboundMessageAccepted",
+                                eventLog.data,
+                                eventLog.topics
+                            );
+
+                            if (decodedEvent.payload.startsWith(MAGIC_BYTES)) {
+                                const channelId = "0x0000000000000000000000000000000000000000000000000000000000000001";
```
ParthDesai, line 266: Is this channel id correct?

girazoki: NO, this is not the channe for erc20

tmpolaczyk (approved):

PR #1176, opened 2025-07-23
collator-assignment: remove is_empty checks by making field private

tmpolaczyk:
Dancekit companion: https://github.com/moondance-labs/dancekit/pull/43/

The collator assignment pallet keeps a `BTreeMap<ParaId, Vec<Collator>>` in storage. This is a map of container chain to list of assigned collators. 

Before this PR, a para id could be present in the map but have 0 assigned collators. This could happen only in the case of not having enough collators. The idea was to be able to quickly identify the case of "we need to add more collators", by looking at this mapping and checking if any chains have 0 collators there.

But that feature wasn't used very often, and it complicates the logic of other users of this pallet. Every time we read the assignment of a chain we had to remember to check if the vector length is 0. So this PR removes this feature. Now if a chain has 0 assigned collators, it will never be in the mapping. There are a few tests that break because of this, but the fix is easy, example: [test/suites/common-tanssi-parachain/author-slot-prediction/test_author_slot_prediction.ts](https://github.com/moondance-labs/tanssi/pull/1176/files#diff-23eac733964cf8410c310f7bbad2e86c7b4f27971138f5e31946e37432b433af)

This requires making the `container_chains` field private, because otherwise anyone can mutate it and add an empty vec. See the dancekit PR description for more info.

Main files to review (the rest are just refactors and tests):

[primitives/traits/src/lib.rs](https://github.com/moondance-labs/tanssi/pull/1176/files#diff-c35139710c5462fa6b2f0c0376c749266f0edfc94da2acf58ddecab8c39de37d)
[chains/orchestrator-relays/runtime/dancelight/src/lib.rs](https://github.com/moondance-labs/tanssi/pull/1176/files#diff-f85584f2d1c4659987dfb22a549d704a4beb54b36c1cba7d8de8af6c6457b290)
[pallets/collator-assignment/src/lib.rs](https://github.com/moondance-labs/tanssi/pull/1176/files#diff-8686f26c396cd2f0d9520d4e109f7f2f163dd0635de8cd2c37c84e4b98e53976)
[pallets/collator-assignment/src/assignment.rs](https://github.com/moondance-labs/tanssi/pull/1176/files#diff-bda72f9e9448884dff5aa950cb1513e223746ece7aaadd370327b0c0f8791a17)


tmpolaczyk:
To try again in the future, maybe split some stuff into a smaller PR

PR #1178, opened 2025-07-24
Remove unused cli arg tanssi_experimental_max_pov_percentage

tmpolaczyk:
It's not used and causes some very weird error messages:

```
$ ./target/release/tanssi-node a
error: invalid value 'a' for '[tanssi_experimental_max_pov_percentage]': invalid digit found in string

For more information, try '--help'.
```

After this PR:

```
$ ./target/release/tanssi-node a
error: unexpected argument 'a' found

Usage: tanssi-node [OPTIONS] [-- <EXTRA_ARGS>...]
       tanssi-node <COMMAND>

For more information, try '--help'.
```


chains/orchestrator-paras/node/src/cli.rs
```diff
@@ -144,14 +144,6 @@ pub struct RunCmd {
     /// Id of the parachain this collator collates for.
     #[arg(long)]
     pub parachain_id: Option<u32>,
-
-    /// EXPERIMENTAL: This is meant to be used only if collator is overshooting the PoV size, and
-    /// building blocks that do not fit in the max_pov_size. It is a percentage of the max_pov_size
-    /// configuration of the relay-chain.
-    ///
-    /// It will be removed once <https://github.com/paritytech/polkadot-sdk/issues/6020> is fixed.
```
evgeny-s, line 152: We did this in 2503 update:
image


Also, this CLI argument can be helpful for [this](https://github.com/paritytech/polkadot-sdk/pull/8040#issuecomment-2754467977).

We might want to add a test for this argument, but not sure if it's a good idea to remove it, this is a fallback argument.

tmpolaczyk: If it compiles it means it was not used

ParthDesai: I added the `id` as otherwise the argument id is same as the variable name and it was conflicting with another argument's id.

Feel free to remove it if not needed.

tmpolaczyk:
Tested this manually using `zombie_tanssi` suite with these collator args:

```json
{
    "name": "Collator1000-01",
    "args": [
        "--no-hardware-benchmarks",
        "--database=paritydb",
        "--wasmtime-precompiled=wasm",
        "--experimental-max-pov-percentage=0",
        "--pool-type=fork-aware",
        "-- --experimental-max-pov-percentage=0"
    ]
},
```

By setting the max pov percentage to 0, collators cannot include any transactions so they only produce empty blocks. To verify, I used polkadot js to try to send a `system.remarkWithEvent`, and it is stuck in pending. Did the test in dancebox and also in the containers.

`tanssi-relay` doesn't support this arg, and upstream polkadot relay chain also doesn't support it.

Since this arg is experimental I'm not adding a test suite, as it will break when this arg is renamed or removed.

PR #1190, opened 2025-07-29
Updated para deposit for dancelight

evgeny-s:
**Description**:
This PR updates para deposit for dancelight


girazoki:
Is there a reason for this? I am not sure we want the same deposit in a test network than in mainnet

evgeny-s:
> Is there a reason for this? I am not sure we want the same deposit in a test network than in mainnet
Yes, we have a [ticket](https://opslayer.atlassian.net/browse/MD-1065) for this.
Let me know if it should be closed. 
@girazoki

PR #1191, opened 2025-07-29
Improve error handling for ETH tokens processor

Agusrodri:
## Description

Improves the way in which we handle validation and execution errors inside `EthTokensLocalProcessor`, ensuring that we log the errors instead of returning them directly. 

Also, two rust integration tests were added to ensure that:

- Transferring a specific asset amount below ED does not error.
- Transferring an amount to an account that does not have ED and being the asset non-sufficent does not error (assets are trapped in this case).


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -222,9 +222,10 @@ where
                 Self::process_xcm_local_native_eth_transfer(eth_transfer_data)
             }
             // TODO: Add support for container transfers here
-            _ => Err(DispatchError::Other(
-                "container transfers not supported yet",
-            )),
+            _ => {
+                log::error!("EthTokensLocalProcessor: container transfers not supported yet");
```
ParthDesai, line 226: Wait, so this means we will not return error here? Can you describe reasoning behind it?

girazoki (approved):
there is the discussion pending, but for me it's an approval. I would like to hear other opinions on the opened discussion

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: I'm also in favor of not erroring in that case, and only return error when the message cannot be decoded by this processor.

ParthDesai (approved):

PR #1194, opened 2025-07-30
feat: permit container chain registration and XCM extrinsics in Starlight

chexware:
This PR enables the extrinsics for container chains registration and XCM for the Starlight runtime,


girazoki (approved):
minor comment

PR #1197, opened 2025-07-31
Re-use sample for zombie tests

evgeny-s:


evgeny-s:
@moondance-labs/coredev  test

PR #1198, opened 2025-07-31
Run specific zombienet tests on schedule

evgeny-s:
**Description**:
This test will run specific zombienet tests on the weekend.


.github/workflows/run-zombienet-tests.yml
```diff
@@ -168,3 +175,13 @@ jobs:
         uses: ./.github/workflow-templates/zombienet-tests
         with:
           test_name: ${{ matrix.test_name }}
+
+      - name: Create issue with gh CLI
+        if: failure()
+        run: |
+          gh issue create \
+            --title "🚨 Critical Test Failure in ${{ github.workflow }}" \
```
girazoki, line 183: I am not sure I want to label this as critical. we know how zombienet tests are, they are not as reliable. I am not sure we win much from this basically, considering most of the zombienet tests run alreayd in PRs no?

evgeny-s: We want to exclude some zombienet tests from PR CI and put it to here, like for example warp sync.

.github/workflows/run-zombienet-tests.yml
```diff
@@ -133,23 +161,24 @@ jobs:
...
-  # Run each test as a separate job using a matrix strategy
   run-tests:
     runs-on: self-hosted
     needs: [ "get-tests", "build" ]
     env:
       CARGO_TARGET_DIR: "target"
     strategy:
       fail-fast: false
-      matrix: ${{ fromJSON(needs.get-tests.outputs.matrix) }}
+      matrix: ${{ fromJson(needs.get-tests.outputs.matrix) }}
```
tmpolaczyk, line 171: I only see `fromJSON` in docs? Why this change?

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#fromjson

evgeny-s: It seems like both options are possible. I've just made it consistent with other similar cases in our repo.

.github/workflows/run-zombienet-tests.yml
```diff
@@ -80,7 +107,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          ref: ${{ needs.set-tags.outputs.git_ref }}
+          ref: ${{ github.event.inputs.sha || 'master' }}
```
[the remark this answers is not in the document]
girazoki: makes sense thanks

PR #1201, opened 2025-07-31
Fix smoke test for inflation rewards

evgeny-s:
**Description**:
Fix smoke test for inflation rewards


test/suites/smoke-test-common-solo/test-inflation-rewards.ts
```diff
@@ -77,8 +87,12 @@ describeSuite({
 
                 const issuance = fetchIssuance(events).amount.toBigInt();
 
+                const blockInflation = isStarlightRuntime(api)
+                    ? STARLIGHT_BLOCK_INFLATION_PERBILL
+                    : DANCELIGHT_BLOCK_INFLATION_PERBILL;
+
                 // expected issuance block increment in prod
-                const expectedIssuanceIncrement = (supplyBefore * DANCELIGHT_BLOCK_INFLATION_PERBILL) / 1_000_000_000n;
+                const expectedIssuanceIncrement = (supplyBefore * blockInflation) / 1_000_000_000n;
```
tmpolaczyk, line 95: This is also the `PER_BILL_RATIO`, I believe this line had the same issue with rounding before and we also had to fix it.

Maybe in the future we should add a function to utils

```ts
function multiplyPerbill(x: bigint, perbill: bigint): bigint {
    return (x * perbill) / 1_000_000_000n;
}
```

evgeny-s: and also helper for ratio PerBill because sometime we take the value using `Perbill::from_rational`

PR #1203, opened 2025-08-01
Add litep2p support to all binaries

tmpolaczyk:
Currently we only support litep2p in relaychain (tanssi-relay binary, and the relay chain part of the other binaries). The other binaries have a CLI flag to select network backend, but it is silently ignored.

This PR adds litep2p support to all binaries, including: tanssi-node, container chain collators, container templates.

The default is still libp2p, this only adds the possibility of using litep2p.

Can be tested adding this CLI arg to the node, and looking for this log:

```
--network-backend=litep2p
```

```
[Relaychain] Running litep2p network backend
[Container-2000] Running litep2p network backend
```

The CLI flag can be set separately for each network, by adding it before or after the `--`.

### DIFF
--- a/scripts/run-release-workflows.sh
+++ b/scripts/run-release-workflows.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+# Help
+usage() {
+  echo "Usage: $0 --prev-runtime=1300 --next-runtime=1400 --prev-client=v0.13.0 --next-client=v0.14.0 --sha=COMMIT_SHA"
+  echo ""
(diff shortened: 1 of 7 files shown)


ParthDesai (approved):

evgeny-s:
@tmpolaczyk , where do we use lite p2p lib? Do we need to update deployment scripts to use this CLI?

tmpolaczyk:
> @tmpolaczyk , where do we use lite p2p lib? Do we need to update deployment scripts to use this CLI?

We don't use it yet. Parity is planning to make litep2p the default in some future release, so the purpose of this PR is to allow us to test it if we want to.

PR #1206, opened 2025-08-06
Refactor nodes to allow embeded Tanssi node in RPC Providers

nanocryk:
This PR refactors nodes service and cli so that they can be used as embeded nodes inside container chain drpc providers nodes.

Data preserver logic is separated into 2 crates:
  - `tc-service-container-chain-spawner` which contains the spawning logic, which is used both by rpc providers AND orchestrator collators
  - `tc-service-container-chain-rpc-provider` which stuff specific for rpc provider.

> This separation is necessary as rpc provider code will need to depend on orchestrator node logic which itself depends on spawner logic. This thus avoid a cyclic dependency.

Orchestrator service is refactored into new crate `tc-service-orchestrator-chain`. It is slightly modified to be usable inside rpc provider mode :
- Modified parameters to disable ContainerChainSpawner that is internally used only for collation. It should not be spawned are rpc provider mode manage it instead.
- Modify a bit some parameters to require less information that is not needed for rpc provider and is hard to generate in its alternate flow.
- Modify function outputs to return a struct with multiple useful objects needed for rpc provider.

> [!WARNING]
> RPC provider mode is moved to its own subcommand instead of being a flag. **This is a breaking change for nodes running rpc provider mode**

Zombienet tests are modified to test rpc providers in the 3 possible modes : connected to remote Tanssi node through RPC, embeded tanssi para, embeded tanssi relay. RPC and embeded para share the same typescript test while embeded relay needs a different test due to obvious differences in network structure and runtime pallets. As zombienet automatically appends arguments at the start for orchestrator and at the end for relay, it doesn't work correctly by default for rpc providers since the first arguments are for the container. For that reason, a script is used to reorder the group of arguments. It also automatically add a `--base-path` argument for the container chain based on the ones from the relay or orchestrator, since it is not performed by zombienet and is required to have working tests.

## Follow-up work:
- Fix issues related to clap and how subcommands work (see https://github.com/paritytech/polkadot-sdk/issues/9356). The new `rpc-provider` subcommand is designed with this in mind and will not need fixing.
- Deduplicate types `dc_orchestrator_chain_interface::DataPreserverAssignment` and `pallet_data_preservers_runtime_api::Assignment ` ?

### DIFF
--- a/.github/workflows/merge-when-ready.yml
+++ b/.github/workflows/merge-when-ready.yml
@@ -51,7 +51,9 @@ jobs:
             fail-fast: false
             matrix:
                 test_name:
-                    - zombie_data_preservers
+                    - zombie_data_preservers_remote_dancebox
+                    - zombie_data_preservers_embedded_dancebox
+                    - zombie_data_preservers_embedded_dancelight
                     - zombie_tanssi_rotation
         steps:
             - name: Checkout


client/service-container-chain-rpc-provider/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+                    tokio_handle.clone(),
+                )
+                .map_err(|err| format!("Orchestrator chain argument error: {}", err))?;
+
+                let polkadot_config = SubstrateCli::create_configuration(
+                    &self.polkadot_cli,
+                    &self.polkadot_cli,
+                    tokio_handle,
+                )
+                .map_err(|err| format!("Relay chain argument error: {}", err))?;
```
Agusrodri, line 156: Probably worth differentiating the error log here between the different modes we have (parachain, solochain and connection through RPC) like this or with something similar that you consider appropriate.
```suggestion
                .map_err(|err| format!("RpcProvider::parachain_mode: Relay chain argument error: {}", err))?;
```

client/service-container-chain-rpc-provider/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+                // Solochain orchestrator
+
+                let tokio_handle = self.config.tokio_handle.clone();
+
+                let polkadot_config = SubstrateCli::create_configuration(
+                    &self.polkadot_cli,
+                    &self.polkadot_cli,
+                    tokio_handle,
+                )
+                .map_err(|err| format!("Relay chain argument error: {}", err))?;
```
Agusrodri, line 215:
```suggestion
                .map_err(|err| format!("RpcProvider::solochain_mode: Relay chain argument error: {}", err))?;
```

client/service-container-chain-rpc-provider/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            // Relay
+            let collator_options = self.collator_options;
+
+            let tokio_handle = self.config.tokio_handle.clone();
+            let polkadot_config = sc_cli::SubstrateCli::create_configuration(
+                &self.polkadot_cli,
+                &self.polkadot_cli,
+                tokio_handle,
+            )
+            .map_err(|err| format!("Relay chain argument error: {}", err))?;
```
Agusrodri, line 256:
```suggestion
            .map_err(|err| format!("RpcProvider::rpc_mode: Relay chain argument error: {}", err))?;
```

client/service-container-chain-rpc-provider/src/lib.rs
ParthDesai: That's good idea. Let's implement both, if possible.

chains/orchestrator-paras/node/src/cli.rs
```diff
@@ -68,7 +68,9 @@ pub enum Subcommand {
     SoloChain(SoloChainCmd),
```
ParthDesai, line 68: What does this mean? Why would orchestrator para start solochain?

ParthDesai: Or is this embedded node?

nanocryk: Yeah as discussed on Slack I found this terminology confusing. "orchestrator paras" is run by collators while "orchestrator relay" is run by validators. So even with dancelight runtime we run "orchestrator paras" node. Note that it is already like that on master.

test/configs/zombie_data_preservers_embeded_dancebox.json
```diff
@@ -125,26 +125,28 @@
                 {
                     "name": "DataPreserver-2000",
                     "validator": false,
-                    "command": "../target/release/container-chain-simple-node",
+                    "command": "./scripts/data-preservers-run-hack.sh ../target/release/container-chain-simple-node rpc-provider --",
```
evgeny-s, line 128: Why can't we just specify the exact command without args, it will execute it as is, from what I see [here](https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/providers/native/nativeClient.ts#L488).

So, theoretically, we make args an empty array and put all the needed arguments to the command. Would it work?

client/node-common/src/cli.rs
```diff
@@ -50,7 +50,7 @@ where
 }
 
 #[derive(Debug)]
-pub struct RelayChainCli<N: Get<&'static str>> {
+pub struct ContainerNodeRelayChainCli<N: Get<&'static str>> {
```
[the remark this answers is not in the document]
ParthDesai: @nanocryk I get it now, so the `RelayChainCli` is mainly used by collators and validators while `ContainerNodeRelayChainCli` is used by full nodes.

client/service-container-chain-rpc-provider/src/cli.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn description() -> String {
+        "Orchestrator embeded node".into()
+    }
+
+    fn author() -> String {
+        env!("CARGO_PKG_AUTHORS").into()
+    }
+
+    fn support_url() -> String {
+        "https://github.com/paritytech/cumulus/issues/new".into()
```
ParthDesai, line 49: Should we change this to our repo?

test/scripts/data-preservers-run-hack.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
ParthDesai, line 1: Let's track the issue. So, once this hack is fixed by zombienet, we can remove this script.

test/moonwall.config.json
```diff
@@ -1086,6 +1086,88 @@
                 }
             ]
         },
+        {
+            "name": "zombie_data_preservers_embeded_dancebox",
```
tmpolaczyk, line 1090:
```suggestion
            "name": "zombie_data_preservers_embedded_dancebox",
```

Also the filenames

chains/container-chains/nodes/frontier/src/command.rs
```diff
@@ -60,6 +54,19 @@ fn load_spec(id: &str, para_id: ParaId) -> std::result::Result<Box<dyn ChainSpec
         "dev" => Box::new(chain_spec::development_config(para_id, vec![])),
         "template-rococo" => Box::new(chain_spec::local_testnet_config(para_id, vec![])),
         "" | "local" => Box::new(chain_spec::local_testnet_config(para_id, vec![])),
+
+        // dummy container chain spec, it will not be used to actually spawn a chain
+        "container-chain-unknown" => Box::new(
```
tmpolaczyk, line 59: I believe this is needed because of how we create the runner later:

```rust
let runner = cli.create_runner(&cmd.base.container_run.normalize())?;

runner.run_node_until_exit(|config| async move {
```

So the dummy spec could be avoided if we rewrite a bit the runner creation code, but I tried and it's not easy.

So it's fine

client/service-container-chain-rpc-provider/src/cli.rs
nanocryk: Created ticket.

PR #1207, opened 2025-08-07
[MD-1263] Validate and optimize the caching for e2e bridge test

evgeny-s:
**Description**:
This PR adds the optimization for the bridge e2e zombienet test.

**What is done:**
- Optimize caching for e2e bridge test (moved caching steps outside composite to enable them)
- Added logging + measurements for the test
- Adding the ci-fast foundry profile to make the CI build as fast as possible. 

**Result**:
Workflow time decreased from [37.54 min](https://github.com/moondance-labs/tanssi/actions/runs/16719724022/job/47326576157?pr=1100) to [29.54 min](https://github.com/moondance-labs/tanssi/actions/runs/16809072317/job/47609858564?pr=1207) .

### DIFF
--- a/test/scripts/bridge/build-snowbridge-contracts.sh
+++ b/test/scripts/bridge/build-snowbridge-contracts.sh
@@ -11,5 +11,6 @@ echo "Building snowbridge contracts"
 
 pushd $contract_dir
 $scripts_root_dir/add-ci-fast-foundry-profile.sh
+export FOUNDRY_PROFILE=ci-fast
 forge build
 popd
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
Re-running the test to see how much we save thanks to the cache

https://github.com/moondance-labs/tanssi/actions/runs/16809072317/job/47663107750

evgeny-s:
> Re-running the test to see how much we save thanks to the cache
> 
> https://github.com/moondance-labs/tanssi/actions/runs/16809072317/job/47663107750

Just FYI, the cache key depends on the hash sum of set-env.sh, because we store commit for tanssi-bridge-relayer and tanssi-simbiotic there.
So, it is cached already and the cache will be used across different PR-s until we change commits.

PR #1208, opened 2025-08-08
Split tp-author-noting-inherent into 2 crates

tmpolaczyk:
To avoid bringing client-side dependencies in tests.

When running `cargo test -p pallet-author-noting`, the std feature is enabled. `tp-author-noting-inherent` defines some client-side methods gated under the `std` feature, assuming that the std feature will never be enabled in a runtime environment. That's true but only when building the wasm runtime, for native runtime and tests the std feature is still enabled.

This PR uses the simple solution of splitting the crate into 2:
* tp-author-noting-inherent: client-side + runtime-side
* tp-author-noting-inherent-core: runtime-side

Then, all runtimes and pallets can depend on `-core`, and only the node depends on the client side features.

Also same change in dancekit `ccp-authorities-noting-inherent`: https://github.com/moondance-labs/dancekit/pull/47

In short, this will make unit tests faster to compile sometimes. It doesn't affect the case of running all tests so I don't expect to see any improvements in CI.

Edit: unfortunately there are still some polkadot crates that bring client dependencies in runtime integration tests, such as `emulated-integration-tests-common`. So this PR only improves the units tests of `pallet-author-noting`, not the `dancelight-runtime` integration tests that I wanted to improve.

To check if the runtime crate depends on libp2p:

```
cargo tree -p dancelight-runtime -e normal | grep libp2p
```


chains/orchestrator-paras/node/Cargo.toml
```diff
@@ -31,11 +31,11 @@ tokio = { workspace = true }
 tokio-util = { workspace = true }
 
 # Local
-ccp-authorities-noting-inherent = { workspace = true, features = [ "std" ] }
+#ccp-authorities-noting-inherent = { workspace = true, features = [ "std" ] }
```
tmpolaczyk, line 34:
```suggestion
```

PR #1213, opened 2025-08-12
revert back changes from gap sync

girazoki:
We revert back changes from gap sync until we find the reason on why it was non-succesful. I asked parity about it:

https://github.com/paritytech/polkadot-sdk/issues/9467


girazoki:
Closing this as it has proven to work with another bootnode

PR #1225, opened 2025-08-22
[ETH -> Tanssi -> Containers] Add support for container native token transfers

Agusrodri:
## Overview
This PR introduces the `NativeContainerTokensProcessor`, a new inbound message processor that enables native container chain tokens to be transferred from Ethereum to their respective container chains via Tanssi through XCM. This extends the bridge functionality to support container chain native tokens alongside the existing support for Tanssi native tokens and foreign (ETH/ERC20) tokens.

### `NativeContainerTokensProcessor`
Processes `SendNativeToken` commands where the destination is a container chain account.

### Key Features

- Validates that the destination parachain owns the token being transferred
- Creates and sends XCM messages to forward tokens to the appropriate container chain
- Handles proper asset reanchoring and fee management
- Supports both `AccountId32` and `AccountKey20` destination types

### Fee Handling

Fees are paid on the container chain side in **Container native token**. The `fee` field from `ForeignAccountId32` (or `ForeignAccountId20`) destination type composes the final fee amount that will be charged on the container side.

For the execution to suceed, it is important that the Ethereum sovereign account has enough container native tokens to cover both the deposit and fees amounts on destination.

### XCM processing
Inside `process_message` function, the new processor takes care of preparing and building an XCM message with the specified container chain as destination, applying the following structure:

- `DescendOrigin`: mutates the origin to the Tanssi's inbound queue pallet instance
- `UniversalOrigin`: after checking that the container chain trust Tanssi's inbound queue pallet as an aliaser for Ethereum, it mutates the origin for it to be the Ethereum network itself
- `WithdrawAsset`: takes total container native token (amount to deposit + fee)
- `BuyExecution`: buy execution using container native asset
- `DepositAsset`: deposit the container native token into the final destination account

### DIFF
--- a/chains/runtime-common/src/relay.rs
+++ b/chains/runtime-common/src/relay.rs
@@ -31,7 +31,7 @@ use frame_system::pallet_prelude::BlockNumberFor;
 use parity_scale_codec::{DecodeAll, Encode, EncodeLike};
 use snowbridge_core::Channel;
 use snowbridge_pallet_inbound_queue::RewardProcessor;
-use sp_core::{Get, H256, H160};
+use sp_core::{Get, H160, H256};
 use sp_runtime::{
     traits::{Hash as _, MaybeEquivalence},
     DispatchError, DispatchResult,


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,239 @@ where
...
+                );
+                false
+            }
+        } else {
+            false
+        }
+    }
+
+    fn process_message(_channel: Channel, envelope: Envelope) -> DispatchResult {
+        if let Some(token_data) =
```
evgeny-s, line 398: We might want to throw an explicit error here in case the payload has not been decoded properly; right now we consider it as a success. Or If we want to avoid  stalled pipe, maybe add a similar error log at least before `Ok(())`.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,239 @@ where
...
+    fn validate_destination_owns_token(
+        token_location: &Location,
+        destination: &Destination,
+    ) -> bool {
+        let chain_part = token_location.interior().clone().split_global().ok();
+
+        let expected_para_id = match destination {
+            Destination::ForeignAccountId32 { para_id, .. } => *para_id,
+            Destination::ForeignAccountId20 { para_id, .. } => *para_id,
+            _ => 0u32,
```
evgeny-s, line 444: We might want to explicitly specify the default `para_id` to None or false, instead of defaulting it to a valid u32. And maybe process it accordingly.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,239 @@ where
...
+        if !GatewayAndChannelValidator::<T>::validate_gateway_and_channel(channel, envelope) {
+            log::warn!(
+                "NativeContainerTokensProcessor::can_process_message: invalid gateway or channel"
+            );
+            return false;
+        }
+
+        // Try decoding the message and check if the destination owns the token being transferred
+        if let Some(token_data) =
+            NativeTokenTransferData::decode_native_token_message(&envelope.payload)
```
evgeny-s, line 380: We do similar stuff in `can_process_message` and `process_message`.
Consider moving the logic to the common helper that will decode and extract the token location.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: If the payload is not decoded properly, `process_message` will be unreachable, as the payload it's also checked inside `can_process_message`, but I can add an explicit error here as well just in case.

Agusrodri: Also done in 97f01bb6bb4e1cd3fc59d70dd5f5170f3c25d0d7.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,239 @@ where
     }
 }
 
+/// `NativeContainerTokensProcessor` is responsible for receiving and processing native container
+/// chain tokens coming from Ethereum and forwarding them to the container chain via Tanssi through XCM.
+pub struct NativeContainerTokensProcessor<
```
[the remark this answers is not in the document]
Agusrodri: More tests added in 2860558e8069b70b0e13507d79a3c34d5496de64.

chains/runtime-common/src/universal_aliases.rs
```diff
@@ -0,0 +1,72 @@
...
+extern crate alloc;
+
+use {
+    alloc::collections::btree_set::BTreeSet,
+    frame_support::{parameter_types, traits::Contains},
+    xcm::latest::prelude::*,
+};
+
+/// The pallet index of the Ethereum inbound queue pallet in the Tanssi runtime.
+pub const PARENT_INBOUND_QUEUE_PALLET_INDEX: u8 = 24;
```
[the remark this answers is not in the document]
ParthDesai: Okay. I agree. 

If the pallet index is not going to change, let's go with that. 

I was suggesting to chose a pallet index that is sentinel, i.e will not change ever.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,269 @@ where
...
+                            fees: tanssi_asset_fee,
+                            weight_limit: Unlimited,
+                        },
+                        DepositAsset {
+                            assets: Definite(container_asset.into()),
+                            beneficiary,
+                        },
+                        // When the execution finishes deposit any leftover assets to the ETH
+                        // sovereign account on destination.
+                        SetAppendix(Xcm(vec![DepositAsset {
```
ParthDesai, line 560: Maybe we need to leave the user tokens trapped, so that it can be claimed later on by the user in case of failed execution of xcm.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,269 @@ where
...
+            return false;
+        }
+
+        // Try decoding the message and check if the destination owns the token being transferred
+        match Self::get_token_data_and_location(&envelope.payload) {
+            TokenDataResult::Success(token_data, token_location) => {
+                Self::validate_destination_owns_token(&token_location, &token_data.destination)
+            }
+            TokenDataResult::DecodeFailure => {
+                log::error!(
```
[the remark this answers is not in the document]
Agusrodri: What about a `warn` level?

chains/runtime-common/src/relay.rs
```diff
@@ -80,96 +75,136 @@ where
...
             return false;
         }
+        true
+    }
+}
 
-        // Try decode the message and check the token id is the expected one
-        match VersionedXcmMessage::decode_all(&mut envelope.payload.as_slice()) {
+/// Information needed to process a native token transfer message from ethereum.
+pub struct NativeTokenTransferData {
```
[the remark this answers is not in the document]
Agusrodri: `NativeTokenTransferData` (more specifically `decode_native_token_message` function) will be used only in processors which purpose is to decode a `SendNativeToken` command. 

Previously we already had one processor to process native Tanssi token transfers (`NativeTokenTransferMessageProcessor`). In this processor we were not using any magic bytes to decode `SendNativeToken`, why would we need it now? 

Let me know if I'm missing something :)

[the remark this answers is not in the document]
Agusrodri: `NativeTokenTransferData` is just a helper struct that we use after decoding a `SendNativeToken` command, and which purpose is to extract the fields from it. We never encode/decode `NativeTokenTransferData`. Is that your concern?

Also, if one processor fails to decode the message (or decodes garbage), it will pass the execution to the next one right? And that one will try to decode it. 

If none of the processors is able to decode/process the message, it makes sense that everything fails no?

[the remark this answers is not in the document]
ParthDesai: > Also, if one processor fails to decode the message (or decodes garbage), it will pass the execution to the next one right? And that one will try to decode it.

Not true. Depending upon what we validate, it can pass `can_process_message`. 

But, if you are decoding `SendNativeToken` command, then we cannot add the discriminator as we do not want to modify original snowbridge command.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
ParthDesai: @Agusrodri Yeah. Not forever. But, will require manual intervention to check the asset amount.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,272 @@ where
     }
 }
 
+/// `NativeContainerTokensProcessor` is responsible for receiving and processing native container
+/// chain tokens coming from Ethereum and forwarding them to the container chain via Tanssi through XCM.
+pub struct NativeContainerTokensProcessor<
```
girazoki, line 338: why is this in the runtime itself if the other processor is in common? We should anticipate we will need it in starlight and therefore put it in common if possible

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,272 @@ where
...
+        // Validate channel and gateway
+        if !GatewayAndChannelValidator::<T>::validate_gateway_and_channel(channel, envelope) {
+            log::warn!(
+                "NativeContainerTokensProcessor::can_process_message: invalid gateway or channel"
+            );
+            return false;
+        }
+
+        // Try decoding the message and check if the destination owns the token being transferred
+        match Self::get_token_data_and_location(&envelope.payload) {
```
girazoki, line 379: can this be succesful for the native token? I know the processor is after the native one but I would prefer to have more barriers

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -345,6 +333,272 @@ where
...
+                );
+                false
+            }
+        }
+    }
+
+    fn process_message(_channel: Channel, envelope: Envelope) -> DispatchResult {
+        match Self::get_token_data_and_location(&envelope.payload) {
+            TokenDataResult::Success(token_data, token_location) => {
+                Self::process_native_token_transfer(token_data, token_location);
```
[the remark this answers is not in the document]
Agusrodri: done in 5f5539a41e0e20bdc8198410e89dcdbe861af12e

chains/container-chains/runtime-templates/frontier/src/xcm_config.rs
```diff
@@ -76,7 +83,7 @@ parameter_types! {
     pub UnitWeightCost: Weight = Weight::from_parts(1_000_000_000, 64 * 1024);
 
     // TODO: revisit
-    pub const RelayNetwork: NetworkId = NetworkId::Polkadot;
+    pub const RelayNetwork: NetworkId = NetworkId::ByGenesis(DANCELIGHT_GENESIS_HASH);
```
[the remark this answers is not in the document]
Agusrodri: At the end I tested again with the `Polkadot` variant and the execution is still successful, so I reverted the change in 5bde61b9752dc3054d1a393f5f17808eae9d275f.

In my opinion it's better to handle this separately in another PR, we already have a ticket for that: https://opslayer.atlassian.net/browse/MD-1223?atlOrigin=eyJpIjoiMzE4ZmVkMDgxMDVjNGUyYmI5ODhkOWM3MGExOWMyODgiLCJwIjoiaiJ9

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
Agusrodri: Added an extra safety check in 50160da4bfe1a2ab8fd84225f58291e80772f3f5.

Let me know if this works :)

test/suites/dev-tanssi-relay/ethereum-token-transfers/test_receive_container_native_tokens_from_eth.ts
```diff
@@ -0,0 +1,143 @@
...
+import { generateEventLog, generateUpdate, mockAndInsertHeadData } from "utils";
+import { expectEventCount, STARLIGHT_VERSIONS_TO_EXCLUDE_FROM_CONTAINER_TRANSFERS } from "helpers";
+import type { KeyringPair } from "@moonwall/util";
+
+describeSuite({
+    id: "DTR1802",
+    title: "NativeContainerTokensProcessor: receive and forward container native tokens from Ethereum",
+    foundationMethods: "dev",
+
+    testCases: ({ it, context }) => {
```
girazoki, line 14: how hard is to generate failure tests here?

- Example: we recieve a non-registered token. 

If so how can we verify it is working as expected? I am assuming we can check for a non-sent event

chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
```diff
@@ -2695,3 +2698,806 @@ fn receive_erc20_tokens_does_not_fail_if_not_sufficient_and_random_address() {
...
+                    topics: event
+                        .encode_topics()
+                        .into_iter()
+                        .map(|word| H256::from(word.0 .0))
+                        .collect(),
+                    data: event.encode_data(),
+                },
+                proof: mock_snowbridge_message_proof(),
+            };
+
```
girazoki, line 3175: what about moving all these tests to `container_token_transfers`.rs?

chains/runtime-common/src/universal_aliases.rs
```diff
@@ -0,0 +1,72 @@
...
+
+parameter_types! {
+    /// Network and location for the Ethereum chain. On Starlight, the Ethereum chain bridged
+    /// to is the Ethereum mainnet, with chain ID 1.
+    /// <https://chainlist.org/chain/1>
+    /// <https://ethereum.org/en/developers/docs/apis/json-rpc/#net_version>
+    pub EthereumNetwork: NetworkId = NetworkId::Ethereum { chain_id: 11155111 };
```
girazoki, line 33: this is actually a bit missleading. I would prefer if we clearly state that this is the Sepolia network id. before I guess this was not in common so it really did not matter, but now being in common, I think it is important

chains/container-chains/runtime-templates/simple/src/xcm_config.rs
```diff
@@ -67,7 +69,7 @@ parameter_types! {
     pub UnitWeightCost: Weight = Weight::from_parts(1_000_000_000, 64 * 1024);
 
     // TODO: revisit
-    pub const RelayNetwork: NetworkId = NetworkId::ByGenesis(WESTEND_GENESIS_HASH);
+    pub const RelayNetwork: NetworkId = NetworkId::Polkadot;
```
girazoki, line 72: I am now having doubts about this again. it probably should be one of dancelight or starlight. But we are again at the same situation. let me research a bit

girazoki: I'd say we put the starlight one. I think it has no implications

chains/container-chains/runtime-templates/frontier/src/xcm_config.rs
```diff
@@ -64,6 +65,10 @@ use {
     xcm_executor::XcmExecutor,
     xcm_primitives::AccountIdAssetIdConversion,
 };
+
+pub const TANSSI_GENESIS_HASH: [u8; 32] =
+    hex_literal::hex!["dd6d086f75ec041b66e20c4186d327b23c8af244c534a2418de6574e8c041a60"];
```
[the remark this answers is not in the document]
Agusrodri: Yeah but as this config doesn't affect the execution in containers, we decided to include the Tanssi one instead of Polkadot, just for consistency purposes in both templates. 

And the reason why it doesn't affect the execution, is because the container token location reanchoring (relative to Ethereum) is done using the `UniversalLocation` each runtime has (Dancelight or Starlight), and this UniversalLocation param already has the proper hash for each chain, so no problem there.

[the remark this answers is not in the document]
Agusrodri: As discussed internally, added a comment for more clarity in 745e790d344321f7e4ed701a7f7c9593bba7c69e.

PR #1229, opened 2025-09-01
Improve e2e tests for ETH bridge

evgeny-s:
**Description**:
This PR improves a bit the e2e test for the ETH bridge.

The scope:

- [x]  Add 2 more external validators that produce blocks
- [ ] Test both veto and instant slashers. For now - blocked by [this](https://moondance-labs.slack.com/archives/C063T9F091P/p1756734699599569) conversation. Solidity team should add veto slasher to the demo setup (Moved to a separate ticket)
- [x] Rewards with proof claiming with more than one operator
- [x] Check what happens when an operator has more than one vault when distributing rewards


test/suites/zombie_tanssi_relay_eth_bridge/test_zombie_tanssi_relay_eth_bridge.ts
```diff
@@ -491,12 +503,23 @@ describeSuite({
...
                     const author = (await relayApi.derive.chain.getHeader(latestBlockHash)).author;
-                    if (author?.toString() === operatorAccount.address) {
-                        return;
+                    if (author) {
+                        if (
+                            [operatorAccount.address, operatorAccount2.address, operatorAccount3.address].includes(
+                                author.toString()
+                            )
+                        ) {
+                            countedSet.add(author.toString());
```
tmpolaczyk, line 516: This could be cleaner if you initialize the set to the 3 operators and then remove new authors from that set until its size is 0. This way when the test fails we can see which operator failed to produce a block.

PR #1236, opened 2025-09-03
Remove the option_env from node version, we always know the version

tmpolaczyk:
Fixes some nodes reporting "Unknown version" at the p2p level on the relay side. Also fixed simple template node reporting "Frontier" as version, also at the p2p level.

Added a missing build.rs script, without it the env variable is never present and it always returns "Unknown version". Note that this is not the version we see in RPC system.version, or in `command --version`, it is the version broadcasted to other p2p peers. It can be seen using the RPC method `system_unstable_networkState`:

```
curl -sH 'content-type: application/json'   -d '{"jsonrpc":"2.0","id":1,"method":"system_unstable_networkState","params":[]}'   http://127.0.0.1:9947/ | jq -r '.result.connectedPeers[]?.versionString' | sort
```

The option_env! was introduced in #911 so this should have been broken in all releases since then.


girazoki (approved):
thank you tomasz!

PR #1238, opened 2025-09-04
Give more priority to collators based on stake

tmpolaczyk:
The collator assignment algorithm tries to keep old collators assigned to the same chain. But that means that when a new collators with more priority joins, they must wait for a new slot to become available. So this PR makes collator priority to always be respected, so when new collators with more stake join, they will always be assigned in the next session. This is implemented by counting the number of collators needed and removing all but the first top N.

In practice this only affects cases such as decreasing the number of collators per chain, because then we keep a random subset of collators assigned, so collators with more priority may be unassigned. With this PR that won't happen anymore, we first unassign all the collators with least stake, and then move the other collators around, so the top N collators will always be assigned.


tmpolaczyk:
Not planned

PR #1240, opened 2025-09-04
[MD-1304] Fix benchmarks for the export_message function for the Relay exporter

evgeny-s:
**Description**:
This PR updates the configuration to enable the `export_message` benchmark to be generated properly ([context](https://github.com/moondance-labs/tanssi/pull/1100#discussion_r2169420906)).

**What was done:**
- Added configuration for `fn export_message_origin_and_destination` that enables the benchmark generation for `export_message` function.
- Added `ToEthDeliveryHelper` that will send required amount of tokens to the holding registry.
- Mock implementation for the exporter converter.


tools/benchmarking.sh
```diff
@@ -108,7 +109,6 @@ function bench {
             fi
             touch "$OUTPUT"
             WASMTIME_BACKTRACE_DETAILS=1 ${BINARY} benchmark pallet \
-            --execution=wasm \
```
Agusrodri, line 111: We might want to revert the changes in this file no?

evgeny-s: No, it seems like this option is deprecated, I would remove it:
```
2025-09-04 16:18:52.430 ERROR main polkadot_sdk_frame::benchmark::pallet: ⚠️  Argument `--execution` is deprecated. Its value of `wasm` has on effect.    

```

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -3133,6 +3134,8 @@ sp_api::impl_runtime_apis! {
                 ).into());
                 pub AssetHubParaId: ParaId = dancelight_runtime_constants::system_parachain::ASSET_HUB_ID.into();
                 pub const RandomParaId: ParaId = ParaId::new(43211234);
+                // Minimum fee amount for exporter: 2680020006582, rounding to a bit more
+                pub const ContainerToEthTransferFee: u128 = 2_700_000_000_000u128;
             }
```
girazoki, line 3139: how can this change break the relay tests?

evgeny-s: The test has failed because of updated benchmarks weights. Increased refTime for tests.

PR #1243, opened 2025-09-05
Removed hardcoded pallet_instance_id from container exporter

evgeny-s:
**Description**:
This PR extracts the pallet instance ID from the reserved asset instead of hardcoding the value in the exporter

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,7 +783,7 @@ dependencies = [
 [[package]]
 name = "asset-test-utils"
 version = "23.0.0"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#0e0d8ef22ea3d9a8eb65bce50fb22591e1d6bdb0"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#d78fcadbaca093cd09d440f43fad47db07454a7b"
 dependencies = [
  "cumulus-pallet-parachain-system",
  "cumulus-pallet-xcmp-queue",
@@ -1166,7 +1166,7 @@ checksum = "89e25b6adfb930f02d1981565a6e5d9c547ac15a96606256d3b59040e5cd4ca3"
 [[package]]
 name = "binary-merkle-tree"
 version = "16.0.0"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#0e0d8ef22ea3d9a8eb65bce50fb22591e1d6bdb0"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#d78fcadbaca093cd09d440f43fad47db07454a7b"
 dependencies = [
  "hash-db",
  "log",
@@ -1425,7 +1425,7 @@ dependencies = [
 [[package]]
 name = "bp-header-chain"
 version = "0.20.1"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#0e0d8ef22ea3d9a8eb65bce50fb22591e1d6bdb0"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#d78fcadbaca093cd09d440f43fad47db07454a7b"
 dependencies = [
  "bp-runtime",
  "finality-grandpa",
@@ -1442,7 +1442,7 @@ dependencies = [
 [[package]]
 name = "bp-messages"
 version = "0.20.1"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#0e0d8ef22ea3d9a8eb65bce50fb22591e1d6bdb0"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#d78fcadbaca093cd09d440f43fad47db07454a7b"
 dependencies = [
  "bp-header-chain",
  "bp-runtime",
@@ -1458,7 +1458,7 @@ dependencies = [
 [[package]]
 name = "bp-parachains"
 version = "0.20.0"
-source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#0e0d8ef22ea3d9a8eb65bce50fb22591e1d6bdb0"
+source = "git+https://github.com/moondance-labs/polkadot-sdk?branch=tanssi-polkadot-stable2503#d78fcadbaca093cd09d440f43fad47db07454a7b"
 dependencies = [
  "bp-header-chain",
  "bp-polkadot-core",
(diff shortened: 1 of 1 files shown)


test/suites/dev-tanssi-relay/xcm/test-export-message-from-para.ts
```diff
@@ -190,5 +190,175 @@ describeSuite({
...
+                const tokenTransferNonceBefore =
+                    await polkadotJs.query.ethereumOutboundQueue.nonce(tokenTransferChannel);
+
+                // Wait until message is processed
+                await jumpToSession(context, 3);
+                await context.createBlock();
+                const tokenTransferNonceAfter =
+                    await polkadotJs.query.ethereumOutboundQueue.nonce(tokenTransferChannel);
+
+                expect(tokenTransferNonceAfter.toBigInt()).toBe(tokenTransferNonceBefore.toBigInt()); // No change in nonce
```
tmpolaczyk, line 275: Maybe we can also find the emitted error, if any?

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
```diff
@@ -313,6 +314,23 @@ where
...
+        }
+
+        let Junctions::X3(arc) = &location.interior else {
+            return Err(());
+        };
+
+        let [j1, j2, j3] = arc.as_ref();
+
+        match (j1, j2, j3) {
+            (GlobalConsensus(_), Parachain(_), PalletInstance(pallet_id)) => Ok(*pallet_id),
```
Agusrodri, line 329: I believe the goal of this PR is not only to extract the `pallet_id`, but instead to replace the entire incoming `GlobalConsensus` by ours, and build a new location combining both (our GlobalConsensus + incoming interior). This should be done after validating that the parachain owns the token being sent (which is already done in `check_reserve_asset_para_id`)

tmpolaczyk: Isn't the GlobalConsensus already replaced later in `asset_location.reanchored(...)`? Or do you mean that this needs to work without using reanchored

evgeny-s: You mean we should take the interior from the ReserveAssetDeposited, replace the GlobalConsensus with the correct relay's with byGenesis hash, and then pass it to the reanchored, and then use the reanchored_location for the token_id conversion. Does it sound right?

Agusrodri: > You mean we should take the interior from the ReserveAssetDeposited, replace the GlobalConsensus with the correct relay's with byGenesis hash

Yeah this is correct, and once you have that combined location, you should be able to use it directly and call `convert_back` with it.

Agusrodri: I took a look and the new approach looks better. What I was trying us to achieve was avoid creating any location manually (like we still do with `container_token_from_tanssi` in the new approach) and also avoid the usage of `reanchored` function and our config variables, but for now I believe this is cleaner enough until we solve this via `pallet-parameters`.

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
```diff
@@ -313,6 +319,15 @@ where
         }
     }
 
+    pub fn get_reanchored_location(&self, location: &Location) -> Result<Location, ()> {
+        let token_split = location.interior().clone().split_global().unwrap();
```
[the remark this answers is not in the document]
evgeny-s: https://github.com/moondance-labs/tanssi/pull/1243/commits/ef4889175fd7e501d3eca211c93dc1b6ad6ee54c

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
evgeny-s: I guess until we pass incorrect genesis in the byConsensus instruction (for now we pass hardcoded tanssi genesis, not dancelight), it is not possible. After we fix that, we no longer need reanchoring; we can just pass the asset as is and it should work, right?
So we can keep reanchoring and remove it later once we introduce pallet parameters, or update the [RelayNetwork](https://github.com/moondance-labs/tanssi/blob/55ca5803371eee507b82e1a7f82b6e74576ab766/chains/container-chains/runtime-templates/frontier/src/xcm_config.rs#L89) to dancelight and remove anchoring now.

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
```diff
@@ -313,6 +319,16 @@ where
         }
     }
 
+    pub fn get_reanchored_location(&self, location: &Location) -> Result<Location, ()> {
+        let token_split = location.interior().clone().split_global().map_err(|_| ())?;
+
+        let container_token_from_tanssi = Location::new(0, token_split.1);
```
girazoki, line 325: Add a comment here explaining that this only works for tokens that belong to the container-chain (and not when they are rtying to export something different)

test/suites/dev-tanssi-relay/xcm/test-export-message-from-para.ts
evgeny-s: https://github.com/moondance-labs/tanssi/pull/1243/commits/15bcea9ea96c8e93440bbf5ab72d2058fc44dae5

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
```diff
@@ -236,9 +236,15 @@ where
...
     }
 
+    pub fn get_reanchored_location(&self, _location: &Location) -> Result<Location, ()> {
+        return Ok(().into());
+    }
+
     fn convert(&mut self) -> Result<(Command, [u8; 32]), sp_runtime::DispatchError> {
         ensure!(self.iter.len() > 0, "Should have at least one instruction");
 
+        let _ = self.get_reanchored_location(&Location::new(0, []));
```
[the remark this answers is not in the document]
evgeny-s: Yes, this is much better: https://github.com/moondance-labs/tanssi/pull/1243/commits/96d0a95520e20037921fff65c7c7922b7876975c
Thanks!

test/suites/dev-tanssi-relay/xcm/test-export-message-from-para.ts
```diff
@@ -190,5 +191,195 @@ describeSuite({
...
+
+                const ethereumNetwork = { Ethereum: { chainId: TESTNET_ETHEREUM_NETWORK_ID } };
+
+                const incorrectContainerAsset: any = {
+                    parents: 1,
+                    interior: {
+                        X3: [
+                            {
+                                GlobalConsensus: {
+                                    ByGenesis: "0x983a1a72503d6cc3636776747ec627172b51272bf45e50a355348facb67a820a",
```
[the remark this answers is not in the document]
evgeny-s: It won't fail, because for the Relay chain, there is no incorrect genesis hash, as long as it is passed as AssetId. For now, we have the hardcoded genesis hash in the container chain side.

primitives/bridge/src/container_token_to_ethereum_message_exporter.rs
```diff
@@ -320,11 +319,13 @@ where
 
         match &location.interior {
             Junctions::X3(arc) => {
-                let [j1, j2, _] = arc.as_ref();
+                let [j1, j2, j3] = arc.as_ref();
                 if let GlobalConsensus(_) = j1 {
                     if let Parachain(id) = j2 {
                         if *id == self.para_id {
-                            return Ok(());
+                            if let PalletInstance(_) = j3 {
```
Agusrodri, line 326: Not sure if we actually want this change, as a container asset could also be represented with `GeneralKey` (or `AccountKey20` if it's a contract in an evm chain for instance). 

I believe we are fine by just having the `GlobalConsensus` and `Parachain` checks.

Agusrodri (approved):

PR #1249, opened 2025-09-09
Avoid recompiling all binaries on git commit hash change

tmpolaczyk:
I noticed that when switching to a different git branch all binaries need to be recompiled even if their code has not changed. This adds several minutes of compilation that could be avoided.

That happens because the `build.rs` scripts are configured to re-run when the git hash changes. That's needed because the git hash is part of the version string, as in `0.14.0-77200a65234`.

So this PR adds a new feature flag to disable this by default, and instead show a version `0.14.0-dev`. This avoids the recompilations. Production builds must pass this new feature flag if they want to show the version with the git hash.

I wanted to do `if env!("PROFILE") == "production"` to detect when to use the git hash, but that doesn't work because the profile can only be "debug" or "release", so we cannot distinguish "production" from "release".

How to test:

```
cargo build --release --features fast-runtime,force-debug
git commit --allow-empty -m 'empty-commit'
cargo build --release --features fast-runtime,force-debug
# the second command should be instant, without this PR it takes a few minutes
```

Also includes a smoke test to ensure we don't deploy the `-dev` version.

### DIFF
--- a/.github/workflows/docker-runtime-draft.yml
+++ b/.github/workflows/docker-runtime-draft.yml
@@ -59,10 +59,10 @@ jobs:
         run: rustup show
       - name: Build Node tanssi-relay
         if:  contains(needs.set-tags.outputs.git_branch, 'starlight')
-        run: cargo build -p tanssi-relay --profile=production
+        run: cargo build -p tanssi-relay --profile=production --features=include-git-hash-in-version
       - name: Build Node parachains
         if:  contains(needs.set-tags.outputs.git_branch, 'para')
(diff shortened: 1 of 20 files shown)


tmpolaczyk:
> Do you think it would be better if we keep the default behaviour as is? But if you want to avoid rebuilding, you add the extra flag that you introduced, and it will override the git commit with dev. In this case, we don't need to touch the existing setup and CI builds, but locally, if you want to avoid rebuilding, just add this flag?

I tried that first but then I realized that my IDE will try to compile without that flag, so I would need to manually add it to whatever config file sets the default features. And also any devs that don't know about this flag won't get the benefit. But I'm open to alternatives.

PR #1260, opened 2025-09-11
Fixed smoke test for the container assignment

evgeny-s:
**Description**:
This PR fixes the smoke test for the container assignment in case if we don't have enough cores available to cover a certain number of collators. 
Added a possibility to debug a specific block.
In case the container still doesn't have collators, but it should, we throw an exception.

image

### DIFF
--- a/.github/workflow-templates/cargo-build/action.yml
+++ b/.github/workflow-templates/cargo-build/action.yml
@@ -107,10 +107,11 @@ runs:
     - name: Verify binary version
       shell: bash
       run: |
-        GIT_COMMIT=`git log -1 --format="%H" | cut -c1-7`
+        #GIT_COMMIT=`git log -1 --format="%H" | cut -c1-7`
+        GIT_COMMIT="-dev"
         TS_VERSION=`./target/release/tanssi-node --version`
         echo "Checking $TS_VERSION contains $GIT_COMMIT"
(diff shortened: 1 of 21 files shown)


test/suites/smoke-test-common-solo/test-configuration-consistency.ts
```diff
@@ -42,27 +59,28 @@ describeSuite({
...
-                const currentAuthorities = await api.query.session.validators();
+                const currentAuthorities = await apiAtBlock.query.session.validators();
 
                 const currentCollatorNumber = Math.min(currentAuthorities.length, config.maxCollators.toNumber());
 
                 const maxParas = Math.trunc(
                     (currentCollatorNumber - config.minOrchestratorCollators.toNumber()) /
                         config.collatorsPerContainer.toNumber()
                 );
+                const maxPossibleAssignmentsBecauseOfCores = await getMaxPossibleAssignmentsBecauseOfCores(apiAtBlock);
```
girazoki, line 75: a few takes on this.

- pendingAssignment is done with the configuration present at the begining of this session
- liveAssingment is done with the configuration present at the beginning of the previous session.

therefore we should compare apples with apples in my opinion. I have 0 doubts this is already more robust than what we have but since we are changing this, better change it for good I guess

PR #1264, opened 2025-09-11
Add trace logs to collator_assignment

tmpolaczyk:
Sometimes it is not clear why some chain doesn't get collators assigned. To help debug that case, this PR adds some trace logs.

Also move the "Collator assignment: keep old assigned" log from info to debug, as we don't need to see it by default.

Example output with dancelight block 2730922:

Captura desde 2025-09-11 16-18-31


pallets/collator-assignment/src/lib.rs
```diff
@@ -401,18 +411,24 @@ pub mod pallet {
                     )
                 };
 
+            log::trace!(
```
girazoki, line 414: I guess it would be ebtter if we haved a target for this logs, otherwise we always have to enable all debug no??

PR #1268, opened 2025-09-12
feat: technical committee and whitelist proposals support for governance

chexware:
This PR adds technical committee and allows it to submit governance proposals through the whitelisted governance track:

TO DO:
- [x] Moonwall tests verifying  that technical committee can enter  maintenance mode  and resume normal operations
- [x] Moonwall tests verifying  that non-committee origin can not create tech committee proposals and vote on them
- [x] Moonwall test verifying  that technical committee  cannot execute a proposal unless the voting threshold is passed
- [x] Moonwall tests verifying  that non-whitelist origin cannot whitelist calls and dispatch whitelisted calls
- [x] Moonwall test verifying  that non-whitelisted call cannot be executed via whitelist track proposal
- [x] Moonwall test veifying that a whitelisted track is able to accept proposed whitelisted calls from a whitelisted origin

### DIFF
--- a/test/configs/frontierContainer.yml
+++ b/test/configs/frontierContainer.yml
@@ -1,4 +1,4 @@
-endpoint: wss://services.tanssi-testnet.network/tanssi-2000
+endpoint: wss://services.tanssi-testnet.network/dancelight-2001
 block: ${env.FRONTIER_TEMPLATE_BLOCK_NUMBER}
 mock-signature-host: true
 db: ./tmp/db_ftrcon.sqlite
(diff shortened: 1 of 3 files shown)


girazoki:
This requires a lot of more tests. you are changing the following things:

- a technical committee that can enter maintenance mode AND propose whitelisted calls. this should be properly tested

- a whitelisted track that acquires a whitelisted origin and is able to accept proposed whitelisted calls. this should be properly tested

- test that any other origin is not able to propose and/or accept

- test that below the required proportion the tech committee cannot take actions

PR #1277, opened 2025-09-16
Added logs to zombienet test for debug

evgeny-s:


evgeny-s:
The issue is with the container chain. It stops producing blocks and the test if failing with the timeout.

PR #1278, opened 2025-09-17
update panic message for reading author data

ParthDesai:
# Description
This PR updates panic message for reading author data and now panics on additional error type called `ReadEntryErr::Decode` signifying that while value was read it was not able to decode.

All other error types results in error being returned.

# Note
Since this changes the runtime behavior it is a breaking change.


girazoki (approved):

pallets/author-noting/src/lib.rs
```diff
@@ -331,7 +331,10 @@ impl<T: Config> Pallet<T> {
         let head_data = relay_state_proof
             .read_entry::<HeadData>(key.as_slice(), None)
             .map_err(|e| match e {
-                ReadEntryErr::Proof => panic!("Invalid proof provided for para head key"),
+                ReadEntryErr::Proof => {
+                    panic!("Value for para-head key cannot be extracted from proof")
+                }
+                ReadEntryErr::Decode => panic!("Value for para-head key cannot be decoded"),
```
tmpolaczyk, line 337: This wasn't a panic before, wouldn't it be better to just log the error? Then its not a breaking change

tmpolaczyk (changes requested):
Revert panic

girazoki:
Yes @ParthDesai this panics on situations we did not have before, e.g., changes the behavior. I think we either put a more descriptive panic where we had it or we leave as it was, but this one changes the behavior and I think for bad.

PR #1286, opened 2025-09-19
Enable Typescript check

evgeny-s:
**Description**:
This PR enables TypeScript checking CI for the tests and disable it for all the files with errors. 
The plan is eventually to fix all the tests and remove @ts-nocheck everywhere in the tests.


.github/workflows/release.yml
```diff
@@ -174,6 +174,9 @@ jobs:
               run: |
                   pnpm install
                   pnpm run fmt
+            - name: "Types check"
+              run: |
+                  pnpm run typecheck
```
tmpolaczyk, line 179: Can you add this to

```
pnpm run lint
```
instead of being a new check?

Then we don't need to modify CI and devs already run `pnpm run lint`

evgeny-s: sure

tmpolaczyk (approved):

PR #1290, opened 2025-09-19
[MD-1339] Zombienet tests improvements

evgeny-s:
**Description**:
This PR improves the zombienet tests developer experience. What is added:
- Added block production monitoring for zombienet
- Added stack trace for the error

### DIFF
--- a/test/suites/zombie_tanssi_relay/test_zombie_tanssi_relay.ts
+++ b/test/suites/zombie_tanssi_relay/test_zombie_tanssi_relay.ts
@@ -1,4 +1,4 @@
-import { beforeAll, describeSuite, expect } from "@moonwall/cli";
+import { afterAll, beforeAll, describeSuite, expect } from "@moonwall/cli";
 import { generateKeyringPair } from "@moonwall/util";
 import { type ApiPromise, Keyring } from "@polkadot/api";
 import type { Signer } from "ethers";
@@ -23,6 +23,7 @@ describeSuite({
         let container2001Api: ApiPromise;
         let container2002Api: ApiPromise;
         let ethersSigner: Signer;
+        let cleanMonitorsFn: () => void;
 
         beforeAll(async () => {
             relayApi = context.polkadotJs("Tanssi-relay");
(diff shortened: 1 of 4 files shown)


test/suites/zombie_tanssi_relay/test_zombie_tanssi_relay.ts
```diff
@@ -47,6 +48,8 @@ describeSuite({
             const paraId2002 = (await container2002Api.query.parachainInfo.parachainId()).toString();
             expect(container2002Network, "Container2002 API incorrect").to.contain("container-chain-template");
             expect(paraId2002, "Container2002 API incorrect").to.be.equal("2002");
+
+            await monitorBlockProduction([relayApi, container2000Api, container2001Api, container2002Api]);
```
evgeny-s, line 52: I don't really like it, will try to find another way how we can incapsulate it

test/utils/zombie.ts
```diff
@@ -321,3 +321,52 @@ export async function directoryExists(directoryPath: PathLike) {
         return false;
     }
 }
+
+export async function monitorBlockProduction(apis: ApiPromise[], blockProductionTimeout = 15000): Promise<() => void> {
```
tmpolaczyk, line 325: Maybe this is too hacky but let's try if it works. Add a comment of what it does. If I understand right:

This will log an error if no blocks are produced in 15 seconds for any chain.

It could cause problems with container 2002 because it starts as not registered, so it may error at the beginning. And also we deregister it later so it may also error at the end of the test.

evgeny-s: image

test/suites/zombie_tanssi_relay/test_zombie_tanssi_relay.ts
```diff
@@ -13,17 +10,24 @@ import {
...
     foundationMethods: "zombie",
     testCases: ({ it, context }) => {
+        addApiMonitoringToContext(context);
+
         let relayApi: ApiPromise;
         let container2000Api: ApiPromise;
         let container2001Api: ApiPromise;
         let container2002Api: ApiPromise;
-        let ethersSigner: Signer;
+        let ethersSigner: Wallet;
```
evgeny-s, line 30: All changes below in this file are related to the fixing types, after `@ts-nocheck` is removed.

PR #1295, opened 2025-09-24
Prioritize assigning chains that were assigned before

nanocryk:
Alter the ordering logic of assignments to prioritize chains that was assigned before.

### DIFF
--- a/pallets/collator-assignment/src/lib.rs
+++ b/pallets/collator-assignment/src/lib.rs
@@ -232,6 +232,8 @@ pub mod pallet {
                     )
                 });
 
+                // Parathreads are not long-lived chains and block production is much more sporadic,
+                // so for them it makes sense to "battle for existing spots".
                 pool_paras.sort_by(|a, b| {
                     T::CollatorAssignmentTip::get_para_max_tip(b.para_id)
                         .cmp(&T::CollatorAssignmentTip::get_para_max_tip(a.para_id))


pallets/collator-assignment/src/lib.rs
```diff
@@ -222,17 +225,21 @@ pub mod pallet {
...
+                        b.para_id,
+                        old_assigned_para_ids,
+                    )
                 });
 
                 pool_paras.sort_by(|a, b| {
-                    T::CollatorAssignmentTip::get_para_tip(b.para_id)
-                        .cmp(&T::CollatorAssignmentTip::get_para_tip(a.para_id))
+                    T::CollatorAssignmentTip::get_para_max_tip(b.para_id)
+                        .cmp(&T::CollatorAssignmentTip::get_para_max_tip(a.para_id))
```
tmpolaczyk: Why don't parathreads use the same logic as parachains for priority? Change it or add a comment explaining why

pallets/collator-assignment/src/tests.rs
```diff
@@ -845,7 +845,7 @@ fn assign_collators_rotation_parathreads_are_shuffled() {
         // Random assignment depends on the seed, shouldn't change unless the algorithm changes
         // Test that container chains are shuffled because 1001 does not have priority
         let shuffled_assignment =
-            BTreeMap::from_iter(vec![(1, 3002), (2, 1000), (3, 1000), (4, 3002)]);
+            BTreeMap::from_iter(vec![(1, 3001), (2, 1000), (3, 1000), (4, 3001)]);
```
tmpolaczyk: Same as above

```suggestion
            BTreeMap::from_iter(vec![(1, 3003), (2, 1000), (3, 1000), (4, 3003)]);
```

pallets/collator-assignment/src/lib.rs
girazoki: reason: they are not long-lived chains, their block production is much more sporadic and between them it makes sense that the outcome is "battle for the existing spots"

pallets/collator-assignment/src/lib.rs
```diff
@@ -224,17 +227,27 @@ pub mod pallet {
             // which means even when we have some parathread cores empty we cannot schedule parachain there.
             if should_charge_tip {
```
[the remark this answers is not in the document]
tmpolaczyk: Nothing, we ignore that case. If there is a new slot we use the fee to sort the new chains and select the first one, but then we don't necessarily charge it the full fee. We charge the min of all the fees of the already assigned chains, as we did before.

So previously assigned chains also pay the fee on every session. If they can't afford the max fee they have set, they will be removed in pre-assignment hook and get no collators. Also after being assigned, they can set the fee to None and then they won't be charged anything on the next sessions.

PR #1266, opened 2025-09-11
Support to send erc20 tokens to container chains

dimartiro:
# Description

This PR updates the `EthTokensLocalProcessor` to support sending ERC-20 tokens (foreign tokens) to container chains via XCM messages, with fees charged in the relay chain’s native token.

## Notes

- Tokens must be registered in `pallet_foreign_asset_creator` before they can be received.  
- Since this feature should only be enabled in Dancelight but we are reusing the same processor, I introduced a `ContainerTransfersEnabled` config flag to control whether container transfers are active for a given runtime.  

## XCM Emulated tests to add:

Transfer native ETH token to the container:
- [x] Frontier  @evgeny-s 
- [x] Simple @evgeny-s 

Transfer ERC20 token from ETH to the container
- [x] Simple @dimartiro 
- [x] Frontier @dimartiro 

~~Transfer ERC20 token from container to ETH~~ (will be added here: https://github.com/moondance-labs/tanssi/pull/1276)

PR #1299, opened 2025-09-25
[MD-1336] Combine both exporters to the single ETH exporter

evgeny-s:
### Description
This PR combines both ETH exporters: the container chain ETH exporter and the relay chain ETH exporter into the single exporter.

For now, it **BLOCKED** by the ERC20 token transfer PRs:
- #1266 
- #1276


evgeny-s:
The approach has changed; we reused the original snowbridge exporter. So we don't need it.

PR #1308, opened 2025-10-01
XCM token transfer payload generator tool

evgeny-s:
### Description

This PR adds a new tool - XCM token transfer payload generator, to be able to generate payloads for XCM emulates tests and Typescript tests for different use cases.

### DIFF
--- a/bin/utils/tanssi-utils/src/xcm_payload.rs
+++ b/bin/utils/tanssi-utils/src/xcm_payload.rs
@@ -1,0 +1,295 @@
+use {
+    ethabi::{encode, Token},
+    hex,
+    parity_scale_codec::Encode,
+    snowbridge_core::TokenIdOf,
+    snowbridge_inbound_queue_primitives::v1::{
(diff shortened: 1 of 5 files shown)


bin/utils/tanssi-utils/src/xcm_payload.rs
```diff
@@ -0,0 +1,311 @@
...
+            let bytes = hex::decode(hex_trimmed).expect("invalid hex for genesis_hash");
+            let mut genesis_array = [0u8; 32];
+            genesis_array.copy_from_slice(&bytes);
+            NetworkId::ByGenesis(genesis_array)
+        } else {
+            NetworkId::ByGenesis(DANCELIGHT_GENESIS_HASH)
+        }
+    }
+
+    fn parse_token_location(&self) -> Location {
```
tmpolaczyk, line 232: If `Location` implements the serde traits, I would simply use `serde_json::from_str` here, this way we support arbitrarily complex locations and don't need custom parsing.

PR #1335, opened 2025-10-20
Add poke_deposit to data-preservers pallet

dimartiro:
# Description

Add the `poke_deposit` extrinsic to the `data-preservers` pallet.
Also include new tests and benchmarks

The goal of this extrinsic is to allow the profile owner to manually refresh the required deposit according to the latest `ProfileDeposit` logic, in case the profile configuration has changed and the deposit amount needs to be updated.

### DIFF
--- a/test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
+++ b/test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
@@ -17,6 +17,7 @@ describeSuite({
             title: "Invulnerables have priority over staking candidates",
             test: async () => {
                 const currentBlock = await api.rpc.chain.getBlock();
+                const currentBlockNumber = currentBlock.block.header.number.toNumber();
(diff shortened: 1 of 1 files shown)


pallets/data-preservers/src/lib.rs
```diff
@@ -470,6 +470,56 @@ pub mod pallet {
...
+            let required = T::ProfileDeposit::compute_deposit(&reg.profile)?;
+            let current = reg.deposit;
+
+            // If the deposit is already correct, do nothing
+            if required == current {
+                return Ok(().into());
+            }
+
+            // Adjust the hold as necessary
+            if let Some(delta) = required.checked_sub(&current) {
```
chexware: Hmm is there a raeson for implementing like this? I feel it will easier to read if it is implemented like:
```
if required > current {
	delta =  required.saturating_sub(current);
	 ...
}
else {
	 delta = current.sturating_sub(required);
	 ...
}
```

PR #1343, opened 2025-10-27
Ethereum outbound queue integration v2

girazoki:
It brings outbound queue v2 changes, in particular:

- outbound queue v2 is attached and selectable to use by user in those extrinsics involving token transfers.
- for system messages, v2 is selectable by setting a storage item UseV2 to true.

Additional changes:
- rewards are paid in tanssi. reward/fee handling is done in the validate_message function of the "exporter" as it is very difficult (or even impossible) to do it the way we want inside the xcm executor.
- the account that collects rewards is the classical snowbridgeFeeAccount
- this account then, upon succesful claim, rewards relayers.
- there is a minimum reward set by the system.

### DIFF
--- a/client/service-container-chain-data-preserver/Cargo.toml
+++ b/client/service-container-chain-data-preserver/Cargo.toml
@@ -52,7 +52,7 @@ polkadot-overseer = { workspace = true }
 sc-client-api = { workspace = true }
 
 [build-dependencies]
-substrate-build-script-utils = { workspace = true }
+substrate-build-script-utils = { workspace = true, optional = true }
 
 [features]
-include-git-hash-in-version = [ "node-common/include-git-hash-in-version", "tc-service-container-chain-spawner/include-git-hash-in-version", "tc-service-orchestrator-chain/include-git-hash-in-version" ]
+include-git-hash-in-version = [ "node-common/include-git-hash-in-version", "substrate-build-script-utils", "tc-service-container-chain-spawner/include-git-hash-in-version", "tc-service-orchestrator-chain/include-git-hash-in-version" ]
--- a/client/service-container-chain-data-preserver/build.rs
+++ b/client/service-container-chain-data-preserver/build.rs
@@ -12,12 +12,35 @@
 // GNU General Public License for more details.
 
 // You should have received a copy of the GNU General Public License
-// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
-
-use substrate_build_script_utils::{generate_cargo_keys, rerun_if_git_head_changed};
+// along with Tanssi.  If not, see <http://www.gnu.org/licenses/>.
 
 fn main() {
-    generate_cargo_keys();
+    #[cfg(feature = "include-git-hash-in-version")]
+    {
+        substrate_build_script_utils::generate_cargo_keys();
+        substrate_build_script_utils::rerun_if_git_head_changed();
+    }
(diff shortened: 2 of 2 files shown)


test/suites/dev-tanssi-relay/ethereum-token-transfers-v2/test-send-ethereum-token-from-para.ts
```diff
@@ -0,0 +1,296 @@
...
+            const newParaId = 500;
+
+            const setChannelTx = await polkadotJs.tx.sudo.sudo(
+                polkadotJs.tx.ethereumTokenTransfers.setTokenTransferChannel(
+                    tokenTransferChannel,
+                    newAgentId,
+                    newParaId
+                )
+            );
+            await context.createBlock(await setChannelTx.signAsync(alice), {
```
Agusrodri: I believe you can remove these lines as they should not be needed for the test.

primitives/bridge/src/custom_exporters/snowbridge_outbound_token_transfer_v2.rs
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    /// Convert the xcm for Polkadot-native token from the origin chain (container chain) into the Command
+    /// To match transfers of Polkadot-native tokens, we expect an input of the form:
+    /// # ReserveAssetDeposited
+    /// # ClearOrigin
+    /// # BuyExecution
+    /// # DepositAsset
+    /// # SetTopic
+    fn make_mint_foreign_token_command(&mut self) -> Result<Message, XcmConverterError> {
```
Agusrodri: Do you think we could remove this function? At the end this use case will be handled by the container exporter.

In the future we could analyze about combining them, but for now it should be safe to remove it.

girazoki: possibly yes. for now let's leave it but think about removing it

pallets/ethereum-token-transfers/src/lib.rs
```diff
@@ -166,6 +195,12 @@ pub mod pallet {
         InvalidMessage(SendError),
         /// The outbound message could not be sent.
         TransferMessageNotSent(SendError),
+        V2SendingIsNotAllowed,
+        TooManyCommands,
+        OriginConversionFailed,
+        LocationToOriginConversionFailed,
+        LocationReanchorFailed,
+        MinV2RewardNotAchieved,
```
girazoki: Reorder this

Agusrodri:
>  So I maintained the channel info set needed for ethereum-token-transfers-pallet-sendV2. I can remove it, but I dont know it feels a bit weird, I will leave that to you guys to decide. @Agusrodri @tmpolaczyk

In my opinion it's fine, you can leave it. That info is not being used so it has no effect at all.

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -503,6 +513,25 @@ mod benchmark_helper {
             P::process_message(channel, envelope)
         }
     }
+
+    impl<P> ProcessorV2<AccountId> for WorstCaseMessageProcessor<P>
```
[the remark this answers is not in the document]
tmpolaczyk: Doesn't matter, it's only for benchmarks

PR #1346, opened 2025-10-29
Fix for smoke test in case if the chain is de-registered and invulnerable not in the assignment list

evgeny-s:
### Description
Sometimes it can be a situation, when the chain has been de-registered and invulnerable, that has been assigned not in the collator assignment list. We check if it was in the list right after the full rotation - `tanssiCollatorAssignment`.`NewPendingAssignment` - `full_rotation` : `true`.


ParthDesai:
@evgeny-s why so many merge commits? And looks like changes from some other work is also present.

PR #1348, opened 2025-10-29
Fix for smoke test in case if the chain is de-registered and invulnerable not in the assignment list

evgeny-s:
### Description
Sometimes it can be a situation, when the chain has been de-registered and invulnerable, that has been assigned not in the collator assignment list. We check if it was in the list right after the full rotation - `tanssiCollatorAssignment`.`NewPendingAssignment` - `full_rotation` : `true`.

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/tests/ethereum_token_transfers.rs
@@ -32,6 +32,7 @@ use {
     },
     hex_literal::hex,
     parity_scale_codec::Encode,
+    snowbridge_core::TokenIdOf,
     snowbridge_core::{AgentId, Channel, ChannelId, ParaId},
     snowbridge_inbound_queue_primitives::v1::{
         Command, Destination, Envelope, MessageProcessor, MessageV1, OutboundMessageAccepted,
@@ -48,6 +49,7 @@ use {
         },
         VersionedAssets, VersionedLocation, VersionedXcm,
     },
+    xcm_executor::traits::ConvertLocation,
 };
 
 #[test]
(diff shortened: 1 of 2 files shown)


test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
```diff
@@ -47,28 +48,88 @@ describeSuite({
...
+    api: ApiPromise,
+    invulnerable: string
+): Promise<void> => {
+    const epochDuration = api.consts.babe.epochDuration.toNumber();
+    const fullRotationSessions = 24; // for both Dancelight and Starlight
+
+    const epochStart = (await api.query.babe.epochStart())[0].toNumber();
+
+    for (let i = 1; i <= fullRotationSessions; i++) {
+        const sessionFirstBlock = epochStart - i * epochDuration;
```
girazoki: change this please. you can always fetch the beginning block of the session by querying 

`(await api.query.babe.epochStart())[0].toNumber();`

like you do at the beginning. therefore you go to the first block of the session, and if its not the full-rotation, then you should go back one block and query the same storage item

test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
```diff
@@ -47,28 +48,88 @@ describeSuite({
...
         });
     },
 });
+
+export const checkIfInvulnerableWasAssignedAfterFullRotation = async (
+    api: ApiPromise,
+    invulnerable: string
+): Promise<void> => {
+    const epochDuration = api.consts.babe.epochDuration.toNumber();
+    const fullRotationSessions = 24; // for both Dancelight and Starlight
```
girazoki: fetch from config

test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
evgeny-s: Fixed:
https://github.com/moondance-labs/tanssi/pull/1348/commits/2c08c4e29b899d799be7320b49e5720176785c1c

test/suites/smoke-test-common-solo/test-invulnerables-priority.ts
```diff
@@ -47,28 +48,92 @@ describeSuite({
...
+    api: ApiPromise,
+    invulnerable: string
+): Promise<void> => {
+    const config = await api.query.collatorConfiguration.activeConfig();
+    const fullRotationSessions = config.fullRotationPeriod.toNumber();
+
+    let epochStartBlock = (await api.query.babe.epochStart())[0].toNumber();
+
+    for (let i = 1; i <= fullRotationSessions; i++) {
+        const prevBlock = epochStartBlock - 1;
```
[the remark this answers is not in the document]
evgeny-s: Added additional corner case check: https://github.com/moondance-labs/tanssi/pull/1348/commits/d2bee31d65a9a5428961da11d36141134aa8ee4f

PR #1363, opened 2025-11-03
Set min_peers_to_start_warp_sync instead of using a cherry-pick

tmpolaczyk:
Draft, first to test that indeed our tests fail if we don't set it.


tmpolaczyk:
All tests pass so either changing min_num_peers is no longer needed, or we need more tests

tmpolaczyk:
Closing in favor of https://github.com/moondance-labs/tanssi/pull/1385

PR #1367, opened 2025-11-07
[DO NOT MERGE] Weights for Dancelight using frame-omni-bencher

evgeny-s:
### Description
Weights for Dancelight using frame-omni-bencher


chains/orchestrator-relays/runtime/dancelight/src/weights/frame_benchmarking::baseline.rs
```diff
@@ -17,31 +17,29 @@
...
 // --template=benchmarking/frame-weight-runtime-template.hbs
-// --json-file
-// raw.json
-// --output
-// tmp/dancelight_weights/frame_benchmarking::baseline.rs
+// --output=tmp/dancelight_weights/frame_benchmarking::baseline.rs
+// --heap-pages=4096
+// --no-storage-info
+// --no-min-squares
+// --no-median-slopes
```
tmpolaczyk: What are these new flags? Are they needed?

PR #1389, opened 2025-11-20
Copy container token transfer feature from dancelight to starlight

tmpolaczyk:
Enable container <> ethrereum token transfers feature in starlight.

xcm tests were failing because we now need to manually call `set_templates_relay_param_to_starlight` to change the ethereum id from sepolia to mainnet. When we support setting params in genesis, this will be simplified.


girazoki:
Can we add tests here?

tmpolaczyk:
> at least copy-pasting those we have in the runtime section of dancelight

link? I already copied some

Agusrodri (approved):

PR #1394, opened 2025-11-25
sync collator with delayed best block

girazoki:
Spawns the collator-process that works in parallel with the collation process with `delayed_best_block`. this makes the collator only trust those blocks that have been accepted by the relay and should avoid issues like https://github.com/paritytech/polkadot-sdk/issues/10398, at least partially.

Once the rotation event takes effect, the block import will keep functioning as it was


tmpolaczyk (changes requested):
I prefer not to do this because it was never needed.

girazoki:
I will close, will reopen if we need it

PR #1397, opened 2025-11-27
Fix missing build script optimization

tmpolaczyk:
This was being run on every commit change, add missing feature to do the same as all the other build scripts


evgeny-s (approved):
Make sure you have all the dependencies propagated this time for this crate as well)

PR #1400, opened 2025-11-28
[MD-1333] Parachain WeightToFee should charge for pov (templates, *box)

evgeny-s:
## Description
When the first tanssi parachains were created, WeightInfo did not have a proof size, and we never updated it. Relay runtimes are ok (dancelight, starlight), so this only affects parachain runtimes (dancebox, flashbox, templates).


chains/orchestrator-paras/runtime/flashbox/src/lib.rs
```diff
@@ -199,7 +199,21 @@ pub mod currency {
...
+    type Balance = Balance;
+
+    fn weight_to_fee(weight: &Weight) -> Self::Balance {
+        let time_poly: FeePolynomial<Balance> = RefTimeToFee::polynomial().into();
+        let proof_poly: FeePolynomial<Balance> = ProofSizeToFee::polynomial().into();
+
+        // Take the maximum instead of the sum to charge by the more scarce resource.
+        time_poly
+            .eval(weight.ref_time())
+            .max(proof_poly.eval(weight.proof_size()))
```
tmpolaczyk: Flashbox tests are failing because the new fee is lower, how is that possible if we take the max here?

```
 FAIL   dev_flashbox  suites/common-tanssi/fees/test_fee_burning.ts > 🗃️  COMMO1001 Fee  burning test suite > 📁  COMMO1001E01 0% of Fees are burned, they are sent to treasury
AssertionError: expected 165480145n to equal 328960145n

- Expected
+ Received

- 328960145n
+ 165480145n
```

chains/container-chains/runtime-templates/frontier/src/lib.rs
```diff
@@ -314,6 +328,24 @@ impl WeightToFeePolynomial for WeightToFee {
     }
 }
 
+/// Maps the proof size component of `Weight` to a fee.
```
girazoki: moonbeam uses a constant multiplier here. not sure if that has any impact..

chains/orchestrator-paras/runtime/flashbox/src/lib.rs
tmpolaczyk: So the test is wrong? That sounds bad, that test is supposed to replicate be the same behavior that RPCs use to calculate the fee. If it fails, it can mean we broke fee estimation...

test/suites/common-tanssi/fees/test_fee_burning.ts
```diff
@@ -48,7 +48,9 @@ describeSuite({
                 const baseWeightToFee = (
                     await polkadotJs.call.transactionPaymentApi.queryWeightToFee({
                         refTime: baseWeight,
-                        proofSize: feeInfo.weight.proofSize.toBigInt(),
+                        // Base weight represents the fixed computational cost, independent of proof size.
+                        // Setting proofSize to 0 ensures we only calculate the ref_time component of the base fee.
+                        proofSize: 0n,
```
evgeny-s: we should use proofSize: 0 here, because the original formula for base fee doesn't include proof size: https://github.com/moondance-labs/polkadot-sdk/blob/ae4b68b36d6da1dd0d6a78643b634296b3fa8039/substrate/frame/support/src/weights/extrinsic_weights.rs#L55

PR #1410, opened 2025-12-01
Use BTreeSet instead of Vec in pallet_inflation_rewards + docs

tmpolaczyk:
And improve struct docs a bit.

And add some logs to warn about edge cases that can never happen.

This only breaks the typescript-api: `paraIds` was a vec so you could do `paraIds.length`, but now it is a set so you need to use `paraIds.size`.

No storage migration needed because Vec and BTreeSet have the same encoding.

## Release Notes

* **Bug Fixes**
  * Improved error handling in the reward system to prevent crashes; errors are now logged instead of causing failures.
  * Enhanced reward accounting for more accurate distribution amounts reflected in reward events.

* **Documentation**
  * Expanded API documentation describing reward initialization and distribution workflow across multiple chains.

* **Tests**
  * Updated test suites to reflect changes in how container chains are tracked and counted.


pallets/inflation-rewards/src/lib.rs
```diff
@@ -371,3 +432,18 @@ impl<T: Config> AuthorNotingHook<T::AccountId> for Pallet<T> {
         T::StakingRewardsDistributor::bench_execute_pending()
     }
 }
+
+/// Convert a `BoundedVec` into a `BoundedBTreeSet` with the same bound.
+// TODO: upstream this into BoundedVec crate
+fn bounded_vec_into_bounded_btree_set<T: core::cmp::Ord + core::fmt::Debug, S: Get<u32>>(
```
Agusrodri: Would it worth to put this function in a common place for us in the meantime? Like runtime-common or something like that, in case we need it in the future.

nanocryk (approved):

pallets/inflation-rewards/src/lib.rs
tmpolaczyk: Sure, but when we need this somewhere else, or when we add more helpers. For now I think there is only this one.

PR #1412, opened 2025-12-02
rewards_per_chain should not depend on chains without collators

tmpolaczyk:
We have a small bug in pallet_inflation_rewards, where we calculate the reward per chain based on the total number of registered chains, regardless of whether they have collators or not. This case doesn't happen in any live runtime because we have enough collators for all chains (for now).

In the case of some chains without collators, there is a portion of the inflation that always goes to the parachain bond. This PR fixes this, increasing the rewards_per_chain in that case. It doesn't affect inflation, the total inflated value is still the same. (except in the edge case where we have 0 collators, because now we won't mint anything).


pallets/inflation-rewards/src/lib.rs
```diff
@@ -91,9 +92,26 @@ pub mod pallet {
...
+                    .collect();
+
+            // Convert to BoundedVec. If the number of chains is greater than the limit, truncate
+            // and emit a warning. This should never happen because we assume that
+            // MaxContainerChains has the same value in all the pallets, but that's not enforced.
+            // A better solution would be for container_chains_with_collators to return an already
+            // bounded data structure.
+            let unbounded_len = container_chains_to_check_unbounded.len();
+            let container_chains_to_check =
+                BoundedVec::truncate_from(container_chains_to_check_unbounded);
```
Agusrodri: Maybe it's a dumb question but how does this `BoundedVec` know about the `MaxContainerChains` limit at the moment of truncating?

test/suites/common-tanssi-parachain/issuance-rewards/test_invulnerable_rewards.ts
```diff
@@ -38,7 +73,7 @@ describeSuite({
                 const chainRewards = (issuance * 7n) / 10n;
                 const expectedOrchestratorReward = chainRewards / 3n;
                 const reward = await filterRewardFromOrchestrator(events, account);
-                expect(reward).to.equal(expectedOrchestratorReward);
+                expect(reward).to.equal(expectedOrchestratorReward + 1n);
```
Agusrodri: Why this +1? Is it related to rounding errors?

tmpolaczyk: I wish I knew, because if I put +1 starlight tests fail, and if I don't then dancelight tests fail :joy:

pallets/inflation-rewards/src/lib.rs
```diff
@@ -91,9 +92,26 @@ pub mod pallet {
...
+            let container_chains_to_check_unbounded: Vec<_> =
+                T::ContainerChains::container_chains_with_collators(ForSession::Current)
+                    .into_iter()
+                    .filter_map(|(para_id, collators)| (!collators.is_empty()).then_some(para_id))
+                    .collect();
+
+            // Convert to BoundedVec. If the number of chains is greater than the limit, truncate
+            // and emit a warning. This should never happen because we assume that
+            // MaxContainerChains has the same value in all the pallets, but that's not enforced.
+            // A better solution would be for container_chains_with_collators to return an already
```
girazoki: exactly!

pallets/inflation-rewards/src/mock.rs
```diff
@@ -14,6 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
+use std::collections::BTreeSet;
```
girazoki: I would prefer if we fmt this (put them inside global use)

pallets/inflation-rewards/src/mock.rs
```diff
@@ -161,18 +163,29 @@ impl Default for Mocks {
 
 pub struct MockContainerChainGetter;
 
-impl tp_traits::GetCurrentContainerChains for MockContainerChainGetter {
-    type MaxContainerChains = ConstU32<5>;
+impl tp_traits::GetContainerChainsWithCollators<AccountId> for MockContainerChainGetter {
```
girazoki: As I said it might make more sense to put the maxLengthParaIds inside a type `GetContainerChainsWithCollators`, like `GetContainerChainsWithCollators` had

test/suites/common-tanssi-parachain/issuance-rewards/test_invulnerable_rewards.ts
girazoki: then you will have to check its between -1 and +1 no?

pallets/inflation-rewards/src/lib.rs
tmpolaczyk: For the future because other pallets use this trait assuming it returns a `Vec`, so changing to `BoundedVec` will break all the tests that use `vec!` syntax.

test/suites/common-tanssi-parachain/issuance-rewards/test_invulnerable_rewards.ts
tmpolaczyk: Fixed, this one was because of perbill rounding, the dancelight/starlight was the other test, which I also fixed the same way. The rounding error is multiplied by number of chains, so -1/+1 works by coincidence because we only have 2 chains registered. I also fixed smoke tests.

PR #1413, opened 2025-12-02
Revisit maintenance mode

girazoki:
Revisits maintenance mod for both dancelight and starlight. the most notable changes are:

- anything related to snowbridge v2 that can be user triggered is moved to disabled in maintenance mode
- utility is enabled, as the dispatch filter applies to whatever thing one wants to do after that.
- we add foreign assets to disabled in maintenance mode, as otherwise people would be able to transfer tokens within the chain.
- added substantial tests for both runtimes


chains/orchestrator-relays/runtime/dancelight/src/tests/maintenance_mode.rs
```diff
@@ -0,0 +1,324 @@
...
+                },
+            ));
+            assert_call_filtered(RuntimeCall::BridgeRelayers(pallet_bridge_relayers::Call::<
+                Runtime,
+            >::deregister {}));
+        });
+}
+
+#[test]
+fn test_non_filtered_calls_maintenance_mode() {
```
Agusrodri: Does it worth adding a simple test for the utility pallet here as well? Maybe checking that forbidden calls are still filtered if we use `utility.batch`

PR #1414, opened 2025-12-03
Register assets in containers genesis

nanocryk:
Registers ETH and TANSSI assets in container chains genesis.

## Requirement
- Merge https://github.com/Moonsong-Labs/moonkit/pull/87 and cherry-pick in `tanssi-polkadot-stable2506`

## Summary by CodeRabbit

* **New Features**
  * Added genesis configuration support for foreign assets, enabling TANSSI and ETH token registration and initialization across container chain runtime templates.

* **Tests**
  * Added comprehensive test suite to verify genesis asset registrations, including asset-to-location mappings and configuration validation for foreign assets.


girazoki (approved):
there is still the open question of "how do we make sure they serve so as to pay for fees in xcm". But I dont have an answer because putting a rate (in genesis we cannot even do that) is somewhat arbitrary. We could potentially do it through foreign-assets too, but it's not like I like it too much

tmpolaczyk (approved):
I guess this needs some sort of manual "migration" for existing chains, or else they will not have those tokens registered.

Also this is breaking right?

girazoki:
> I guess this needs some sort of manual "migration" for existing chains, or else they will not have those tokens registered.
> 
> Also this is breaking right?

Not breaking, they simply need to register it manually with root. I mean we can include a migration, but I feel like that is babysitting. what do you guys think?

Cargo.toml
```diff
@@ -71,6 +71,7 @@ uninlined_format_args = { level = "allow", priority = 3 }
 [workspace.lints.rust]
 unsafe-code = { level = "deny", priority = 1 }
 
+
```
Agusrodri:
```suggestion
```

evgeny-s (approved):

PR #1417, opened 2025-12-04
chore: 📦️ Update Moonwall Deps

timbrinded:
Update to the latest moonwall which fixes run command.

- Fixed port issue
- Improve tailing screen


test/suites/common-container-chains/dynamic-params/test-relay-network.ts
```diff
@@ -1,7 +1,10 @@
+import "@moonbeam-network/api-augment";
```
girazoki: why not import the tanssi-api¿

timbrinded: Ive cast the types so we dont need this anymore anyway

PR #1426, opened 2025-12-09
chore: 📦️ Update moonwall lib version to support `rpc-timeout` for chopsticks tests

evgeny-s:
## Description
This PR updates Moonwall lib version to be able to use the `rpc-timeout` parameter for chopsticks upgrade tests.


girazoki:
I dont think the retry is going to help much. what we need is to discover why the rpc-timeout hits at 60s when we have it programmed to hit at 120s. that is what we should investigate

PR #1440, opened 2025-12-16
Test that relayer pays for snowbridge v2 xcm execution

tmpolaczyk:
* Small runtime change: emit an error event in case of overweight xcm
* Improve update validators test, to check that validators didn't change after wrong origin tries to update them
* Add typescript tests about who pays for xcm execution in snowbridge v2. We assert that the relayer balance cost depends on the complexity of xcm message being executed. This must be typescript because rust tests don't charge for weight so a relayer with 0 balance works.

### DIFF
--- a/test/moonwall.config.json
+++ b/test/moonwall.config.json
@@ -960,7 +960,11 @@
             "timeout": 600000,
             "testFileDir": ["suites/zombie_tanssi_test_build_spec"],
             "runScripts": [
-                "build-spec-dancelight.sh", "test-build-spec.sh"
+                "build-spec-dancelight.sh",
+                "test-build-spec.sh",
+                "compile-wasm.ts compile -b ../target/release/tanssi-relay -o wasm -c specs/tanssi-relay.json",
+                "compile-wasm.ts compile -b ../target/release/container-chain-simple-node -o wasm -c specs/single-container-template-container-2000.json",
+                "compile-wasm.ts compile -b ../target/release/container-chain-frontier-node -o wasm -c specs/single-container-template-container-2001.json"
             ],
             "foundation": {
                 "type": "zombie",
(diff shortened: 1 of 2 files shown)


test/utils/light-client.ts
```diff
@@ -54,7 +55,7 @@ function encodeAsForeignTokenERC20(tokenId: string, value: bigint): Uint8Array {
     return getBytes(encoded);
 }
 
-function encodeXcmMessage(api: ApiPromise, xcmInstructions): Uint8Array {
+function encodeXcmMessage(api: ApiPromise, xcmInstructions: StagingXcmV5Xcm): Uint8Array {
```
tmpolaczyk: This was to get a type error if I use `originType` instead of `originKind` in `Transact:`, but it doesn't work. I'm leaving it because at least it documents that the xcm must be v5, even if it doesn't enforce it...

test/suites/dev-tanssi-relay/ethereum-inbound-queue-v2/test_receive_xcm_transact.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    testCases: ({ it, context }) => {
+        let polkadotJs: ApiPromise;
+        let alice: KeyringPair;
+        let isStarlight: boolean;
+        let ethNetworkId: number;
+        let shouldSkipStarlightSnV2TT: boolean;
+        let specVersion: number;
+        let sovereignAccountAddress: string;
+        let balanceDiff1: number | undefined;
```
evgeny-s: Seems like the type is incorrect here, because below you assign the bigint value, so it should be `biding | undefined`

tmpolaczyk: Right, it should be `bigint`

tmpolaczyk:
```suggestion
        let balanceDiff1: bigint | undefined;
```

test/utils/light-client.ts
tmpolaczyk: Now it gives me a lint error because the type is not correct, so I removed this

PR #1454, opened 2026-01-05
Introduce .coderabbit.yaml configuration file

evgeny-s:
## Added configuration for coderabbit review tool.
Disabled:
- auto-generated PR summary
- poem

## Summary by CodeRabbit

* **Chores**
  * Updated internal development configuration.

### DIFF
--- a/.coderabbit.yaml 
+++ b/.coderabbit.yaml 
@@ -1,3 +1,3 @@
 reviews:
   high_level_summary: false
-  auto_assign_reviewers: false
+  poem: false


.coderabbit.yaml 
```diff
@@ -0,0 +1,203 @@
...
+      threshold: 80
+    title:
+      mode: warning
+      requirements: ''
+    description:
+      mode: warning
+    issue_assessment:
+      mode: warning
+    custom_checks: []
+  tools:
```
evgeny-s: Check the number of static analyzers it runs during the review, impressive.

tmpolaczyk: Do we need to list them all? I'm worried that they add a new one and we need to manually add it to config. Or can the config be empty and it will use the default? That would be prettier, if it was a small config file with only the things we disabled.

evgeny-s: Yes, makes sense. It seems like it might work; let's try this approach.

PR #1461, opened 2026-01-13
add max xcm weight config in message processors

ParthDesai:
# Description
This is second part of refactoring of message processor. This PR adds Max xcm weight config option for message processors dealing with xcm and also modifies `execute_xcm` to take max xcm weight as parameter and handle the xcm execution outcome to enable reporting of used weight.

# Note
This is refactoring as of now, there is no change in functionality.

### DIFF
--- a/chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
+++ b/chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
@@ -15,11 +15,15 @@
 // along with Tanssi.  If not, see <http://www.gnu.org/licenses/>
 
 use crate::tests::common::ExtBuilder;
-use crate::{xcm_config, Runtime};
+use crate::{xcm_config, Runtime, RuntimeCall, RuntimeEvent, System};
+use parity_scale_codec::Encode;
+use sp_core::Hasher;
 use tanssi_runtime_common::processors::v2::execute_xcm;
 use xcm::latest::Error::WeightLimitReached;
-use xcm::latest::Instruction::{RefundSurplus, WithdrawAsset};
-use xcm::latest::{Asset, AssetId, Fungibility, InstructionError, Location, Outcome, Weight};
+use xcm::latest::Instruction::{RefundSurplus, Transact, WithdrawAsset};
+use xcm::latest::{
+    Asset, AssetId, Fungibility, InstructionError, Junction, Location, OriginKind, Outcome, Weight,
+};
 
 #[test]
 fn test_execute_xcm() {
@@ -30,7 +34,10 @@ fn test_execute_xcm() {
         // detail of xcm executor, and we can't assume anything since we work with trait) before we pass
         // xcm to executor that logic will never be triggerred.
 
+        type Hashing = <Runtime as frame_system::Config>::Hashing;
+
         // Scenario 1: execute xcm that uses more weight than this limit
+        let remark = vec![1];
         assert!(matches!(
             execute_xcm::<
                 Runtime,
@@ -39,7 +46,16 @@ fn test_execute_xcm() {
             >(
                 Location::here(),
                 Weight::from_all(200),
-                vec![RefundSurplus].into(),
+                vec![Transact {
+                    origin_kind: OriginKind::Native,
+                    fallback_max_weight: None,
+                    call: RuntimeCall::System(frame_system::Call::remark_with_event {
+                        remark: remark.clone(),
+                    })
+                    .encode()
+                    .into(),
+                }]
+                .into(),
             ),
             Err(InstructionError {
                 index: 0,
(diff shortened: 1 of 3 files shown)


chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
```diff
@@ -83,6 +83,7 @@ pub const SLOTS_PER_EPOCH: u32 = snowbridge_pallet_ethereum_client::config::SLOT
 // Ethereum Bridge
 parameter_types! {
     pub storage EthereumGatewayAddress: H160 = H160(hex_literal::hex!("EDa338E4dC46038493b885327842fD3E301CaB39"));
+    pub const MaxXcmWeight: Weight = Weight::MAX; // TODO: Change this to appropriate value
```
tmpolaczyk: You can put something conservative like `Perbill::from_percent(20) * BlockWeights::get().max_block;`. I don't remember exactly what value we discussed.

Also I don't see `worst_case_message_processor_weight` in this PR and the default is `Weigth::zero()`, shouldn't the xcm processor return this `MaxXcmWeight` as `worst_case_message_processor_weight`?

ParthDesai: That will be in next PR. :) This PR again does not change any behavior.

chains/runtime-common/src/processors/v2/raw_message_processor.rs
```diff
@@ -251,17 +257,40 @@ where
...
+                } => {
+                    log::error!(
+                        "Error while executing xcm in raw message processor: {:?}",
+                        instruction_error
+                    );
+                    Ok(Some(used))
+                }
+                Outcome::Error(instruction_error) => {
+                    log::error!(
+                        "Error while staring xcm execution in raw message processor: {:?}",
```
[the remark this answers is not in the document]
ParthDesai: Fixed

chains/orchestrator-relays/runtime/dancelight/src/bridge_to_ethereum_config.rs
ParthDesai: make sense. Will do it in this PR.

chains/runtime-common/src/processors/v2/mod.rs
```diff
@@ -339,20 +339,25 @@ where
...
-    XcmProcessor::prepare_and_execute(origin, xcm, &mut message_id, weight, weight)
-        .ensure_complete()
+    Ok(XcmProcessor::prepare_and_execute(
+        origin,
+        xcm,
+        &mut message_id,
+        weight,
+        weight,
+    ))
 }
```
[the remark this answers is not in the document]
ParthDesai: Fixed this in https://github.com/moondance-labs/tanssi/pull/1461/commits/b2a9f15a6d23e04befca388f1b00862764ed7c0e

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
```diff
@@ -0,0 +1,94 @@
...
+use xcm::latest::{Asset, AssetId, Fungibility, InstructionError, Location, Outcome, Weight};
+
+#[test]
+fn test_execute_xcm() {
+    ExtBuilder::default().build().execute_with(|| {
+        // Note that there is no way to test Outcome::Error variant as the current xcm executor
+        // uses XcmWeighter to only test if the xcm has enough weight.
+        // Since we explicitly check the weight (reason being as the weight checking is implementation
+        // detail of xcm executor, and we can't assume anything since we work with trait) before we pass
+        // xcm to executor that logic will never be triggerred.
```
tmpolaczyk, line 35: Hmm and what if we send an xcm message that fails on the first instruction, will it return `Outcome::Incomplete` or `Outcome::Error`? If the `Outcome::Error` branch is unreachable add a comment there, and we should be able to see it in code coverage.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
```diff
@@ -0,0 +1,94 @@
...
+                ),
+                RefundSurplus,
+            ]
+            .into(),
+        );
+        assert!(matches!(
+            response,
+            Ok(Outcome::Incomplete {
+                used: Weight { .. },
+                error: InstructionError { index: 2, .. }
```
tmpolaczyk, line 176: In this case, can we also assert that execution is not reverted? For example if `RefundSurplus` emits an event, check that the event exists in `System::events()`. Or add a different instruction that does a balance transfer or a storage write, and check that when `execute_xcm` returns error but it has already executed that instruction, the side effect is preserved.

chains/runtime-common/src/processors/v2/symbiotic_message_processor.rs
```diff
@@ -244,7 +249,11 @@ where
...
         extracted_message: Self::ExtractedMessage,
     ) -> Result<Option<Weight>, MessageProcessorError> {
-        // TODO: Add proper consumed weight
-        process_message(extracted_message).map(|_| None)
+        process_message(extracted_message).map(|_| Some(SetExternalValidatorsWeight::get()))
+    }
+
+    fn worst_case_message_processor_weight() -> Weight {
+        // This ensures the worst case covers both the primary and fallback paths regardless of configuration.
+        MaxXcmWeight::get().max(SetExternalValidatorsWeight::get())
```
tmpolaczyk, line 257: This is nice. I would have put `SetExternalValidatorsWeight::get()` because `process_extracted_message` always returns the same weight, but that's wrong because the fallback is also executed as part of this processor, as you say in the comment.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
ParthDesai: Yes, It will fail with `Outcome::Incomplete`, `Outcome::Error` branch is unreachable. Let me add comment there.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
ParthDesai: cool. I will add instruction to execute System::Remark and check the storage.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
tmpolaczyk: I also checked with a fuzzer and it seems unreachable.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
ParthDesai: Done.

chains/orchestrator-relays/runtime/dancelight/src/tests/processors/v2/common.rs
ParthDesai: Added comment as well.

PR #1467, opened 2026-01-19
Add V2 Snowbridge Integration Tests for Ethereum ↔ Container Chain Token Transfers

Agusrodri:
## Description

This PR adds comprehensive XCM emulated integration tests for the Snowbridge V2 inbound queue, validating token transfers between Ethereum and Tanssi container chains. It also includes necessary XCM configuration changes to support container native token transfers via V2.

### Overview

The new tests exercise the V2 message format using `IGatewayV2::OutboundMessageAccepted` events with `RawPayload::Xcm` payload encoding. These tests validate the `RawMessageProcessor` and related V2 processors in `tanssi-runtime-common`.

### Test Coverage

**Ethereum → Container Chain Transfers** (`eth_to_container_transfers_v2.rs`)

Four token transfer scenarios are covered:

| Token Type | Asset Kind | Description |
|------------|-----------|-------------|
| ERC20 tokens | kind: 0 | Native Ethereum ERC20 tokens transferred to containers |
| Native ETH | `value` field | Ether transferred as a foreign asset |
| Tanssi tokens | kind: 1 | Relay chain tokens previously bridged to Ethereum, sent back to containers |
| Container native tokens | kind: 1 | Container chain tokens previously bridged to Ethereum, returned to origin chain |

Each scenario is tested against both container chain templates:
- **Frontier Template** (EVM-compatible, `AccountId20`)
- **Simple Template** (Substrate-native, `AccountId32`)

#### Dancelight Runtime

Added `EthereumAssetReserveForContainerAssets<EthereumLocation>` to `IsReserve` in XcmConfig. This is a special reserve filter needed for V2 inbound processing that recognizes Ethereum as the reserve for container chain native assets when they're being transferred back from Ethereum. This is necessary because container assets don't exist in Tanssi, so we can't extract them from Ethereum's sovereign account—instead, we use `ReserveAssetDeposited` to place them in holding.

#### Container Chain Templates (Frontier & Simple)

**New Aliaser**: Added `type Aliasers = ParentAsEthereumAliaser<EthereumLocation>` to XcmConfig
   - Allows the parent (relay chain) to alias as the Ethereum origin
   - Required for container native token transfers where the relay sends XCM with `preserve_origin: true`

#### New XCM Primitives (`tp-xcm-commons`)

- **`EthereumAssetReserveForContainerAssets`**: Filter to recognize Ethereum as a reserve for container chain native assets (special V2 case)

#### New Universal Alias (`universal_aliases.rs`)

- **`ParentAsEthereumAliaser`**: Implements `ContainsPair<Location, Location>` to allow the parent to alias as Ethereum's global consensus location


chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
```diff
@@ -0,0 +1,2227 @@
...
+            <<Dancelight as DancelightRelayPallet>::Balances as Mutate<_>>::mint_into(
+                &EthereumSovereignAccount::get(),
+                10_000_000_000_000_000_000u128
+            )
+        );
+
+        // Get Tanssi token ID for the V2 message
+        let token_location_reanchored = token_location
+            .clone()
+            .reanchored(&EthereumLocation::get(), &UniversalLocation::get())
```
evgeny-s, line 322: A question. I see we use it in some other places, but why do we try to reanchor the "local" location and put universal as a context. Shouldn't we specify the `token_location` as a location from global path, like GlobalConsensus -> Dancelight -> Here

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
```diff
@@ -0,0 +1,2227 @@
...
+            );
+
+        let receiver_tanssi_balance_after =
+            <FrontierTemplate as FrontierTemplateParaPallet>::ForeignAssets::balance(
+                RELAY_NATIVE_TOKEN_ASSET_ID,
+                &token_receiver,
+            );
+
+        assert!(receiver_tanssi_balance_after > 0);
+        assert!(receiver_tanssi_balance_after < dest_fee_amount);
```
evgeny-s, line 567: Maybe we also want to check the exact amount of fees we paid in Tanssi on the destination for the execution. I think we can get it from the event.

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
```diff
@@ -0,0 +1,2227 @@
...
+                    data: destination_xcm_bytes.encode().into(),
+                },
+                claimer: vec![].into(),
+                value: 0,
+                executionFee: 0,
+                relayerFee: 0,
+            },
+        };
+
+        assert_ok!(send_inbound_message_v2(event));
```
evgeny-s, line 836: We may also want to check the events to ensure we don't have any failures that are masking minor issues.

chains/container-chains/runtime-templates/frontier/src/xcm_config.rs
```diff
@@ -256,7 +256,7 @@ impl xcm_executor::Config for XcmConfig {
     type UniversalAliases = CommonUniversalAliases;
     type CallDispatcher = RuntimeCall;
     type SafeCallFilter = Everything;
-    type Aliasers = Nothing;
+    type Aliasers = ParentAsEthereumAliaser<EthereumLocation>;
```
tmpolaczyk, line 259: I can't find the `AliasOrigin(EthereumLocation)` instruction, where is it? Or what's the part of the code that requires this aliaser?

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
```diff
@@ -0,0 +1,2227 @@
...
+
+        let custom_xcm: Vec<Instruction<()>> = vec![
+            // Transfer to parachain
+            InitiateTransfer {
+                destination: Location::new(0, Parachain(container_para_id)),
+                remote_fees: Some(AssetTransferFilter::ReserveDeposit(
+                    fee_asset.clone().into(),
+                )),
+                // We clear the origin in destination.
+                preserve_origin: false,
```
tmpolaczyk, line 454: This custom_xcm is all user controlled, so what happens if the user puts `true` here? I guess the token transfer will fail? I see we have tests with both `preserve_origin: false` and `preserve_origin: true`.

Agusrodri: If a user puts `true` in preserve_origin, it means that when the whole XCM message reaches destination, the origin (before executing the custom_xcm) will be Ethereum, so all the `custom_xcm` is executed on behalf of it. We need to set it to `true` when we are transferring container native assets, as the real native assets in destination are placed in the `EthereumSovereignAccount`, and we need to extract them from there.

In that case, it's true that a user could try "draining" the Ethereum sovereign account on destination, and this would only work for container native assets (since all the other real ERC20 and Tanssi assets are not present in that account). To be honest I'm not sure if that's something we can control given the current configuration for inbound processing in V2. I can open a new ticket to investigate a way of preventing this.

And regarding this specific test, putting `true` or `false` here should be the same, as we are not dealing with the `EthereumSovereignAccount` on destination.

Agusrodri: At the end this problem is solved here: https://github.com/moondance-labs/tanssi/blob/ea54298b7959769a487dc3f5b09b3f6f2e8be567/chains/runtime-common/src/processors/v2/mod.rs#L319

With a `DescendOrigin`, if a user sends a `preserve_origin: true`, the execution will fail. I'll add a test for that.

chains/container-chains/runtime-templates/frontier/src/xcm_config.rs
Agusrodri: Here is the part of the code in which `InitiateTransfer` instruction injects `AliasOrigin`: https://github.com/moondance-labs/polkadot-sdk/blob/tanssi-polkadot-stable2506/polkadot/xcm/xcm-executor/src/lib.rs#L1379

And then `AliasOrigin` instruction checks the aliasers here: https://github.com/moondance-labs/polkadot-sdk/blob/tanssi-polkadot-stable2506/polkadot/xcm/xcm-executor/src/lib.rs#L1794

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
Agusrodri: This reanchoring is done because we need to extract the `token_id` in the next step via `NativeToForeignId`. This storage expects the location parameter in a reanchored way. 

We initially registered the token with `token_location` set to `Location::here()` for easier setup (which is correct since the reanchoring is later done internally in `EthereumSystemV2`) but we cannot use this location directly to fetch the `token_id` so that's why we need to convert it again using the universal context.

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
Agusrodri: Test added here: 143329757008b45bdedccd315f8c737f97f459cc

tmpolaczyk (approved):

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
Agusrodri: Added checks for events in 6f7e5a789d75bb94f12d2ee668b7a8144fc5e3e8.

chains/integration-tests/xcm/emulated/tests/assets/dancelight/src/tests/eth_to_container_transfers_v2.rs
Agusrodri: We cannot extract the exact amount that was used to pay for fees directly from the event, but I added an event check to ensure that the `ForeignAssets::Issued` event exists for the refunded amount of Tanssi assets (which should match the final `receiver_tanssi_balance_after`).

By having the checks above + the event check I believe we are fine enough, but let me know your thoughts :)

evgeny-s (approved):

PR #1472, opened 2026-01-20
[MD-490] Remove verify and hold directly

evgeny-s:
## Description
This PR simplifies the deposit handling logic in the registrar pallet by removing redundant checks and relying on the underlying hold mechanism to handle insufficient funds errors directly.


pallets/registrar/src/lib.rs
```diff
@@ -956,8 +952,8 @@ pub mod pallet {
                 return Err(Error::<T>::GenesisDataTooBig.into());
             }
 
-            // Hold the deposit, we verified we can do this
-            T::Currency::hold(&HoldReason::RegistrarDeposit.into(), &account, deposit)?;
+            T::Currency::hold(&HoldReason::RegistrarDeposit.into(), &account, deposit)
+                .map_err(|_| Error::<T>::NotSufficientDeposit)?;
```
tmpolaczyk, line 956: Nice! I coudn't find any docs that explicitly say this is fine (e.g:, if can_hold returns err then hold will also always return err), but I found this in the docs for [can_hold](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/hold/trait.Inspect.html#method.can_hold):

> NOTE: This does not take into account changes which could be made to the account of who (such as removing a provider reference) after this call is made. Any usage of this should therefore ensure the account is already in the appropriate state prior to calling it.

So I guess using `can_hold` is discouraged.

tmpolaczyk (approved):

PR #1473, opened 2026-01-20
[MD-1520] Handle commitment record separately for v1 and v2

evgeny-s:
## Description
This PR implements unified commitment root handling for Snowbridge outbound queues in the Dancelight runtime. The OutboundMessageCommitmentRecorder pallet now stores commitments from v1 and v2 queues separately and computes a Merkle tree root when both are present, ensuring the relayer receives a single, unified commitment for Ethereum smart contract validation. Changes include adding new storage for v2 commitments, updating the commitment logic, and adding comprehensive integration tests. The pallet version has been bumped to 1 to reflect the new storage addition.


evgeny-s:
@coderabbitai review

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +48,53 @@ pub mod pallet {
...
     #[pallet::storage]
     pub type RecordedCommitment<T: Config> = StorageValue<_, H256, OptionQuery>;
 
+    /// Message commitment from last block (v2).
+    /// This will be set only when there are messages to relay.
+    #[pallet::storage]
+    pub type RecordedCommitmentV2<T: Config> = StorageValue<_, H256, OptionQuery>;
+
     impl<T: Config> Pallet<T> {
         pub fn take_commitment_root() -> Option<H256> {
```
tmpolaczyk, line 82: Can you remind me who uses this function? The PR description is useless because it explains what are the changes, which I can already see in the diff, but it doesn't explain why.

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +48,53 @@ pub mod pallet {
...
-                Pallet::<T>::deposit_event(Event::<T>::CommitmentRootRead { commitment });
-                Some(commitment)
-            } else {
-                None
+            let v1 = RecordedCommitment::<T>::take();
+            let v2 = RecordedCommitmentV2::<T>::take();
+            match (v1, v2) {
+                (Some(v1_commit), Some(v2_commit)) => {
+                    // Both v1 and v2, compute unified merkle root
+                    let root = merkle_root::<Keccak256, _>([v1_commit, v2_commit].into_iter());
```
tmpolaczyk: So I guess we need a custom relayer?

pallets/outbound-message-commitment-recorder/src/lib.rs
evgeny-s: The relay is fetching commitment and proof and validating it.

pallets/outbound-message-commitment-recorder/src/lib.rs
evgeny-s: Based on the discussion with @ParthDesai , Relay doesn't care; we just pass the upper-level root that consists of v1 and v2, and the validation still should not break.

pallets/outbound-message-commitment-recorder/src/lib.rs
evgeny-s: https://github.com/moondance-labs/tanssi-bridge-relayer/blob/acab82745d9fb6f09f71e6c637617b61c550275a/relays/solochain/scanner.go#L312

chains/orchestrator-relays/runtime/dancelight/src/lib.rs
```diff
@@ -3158,14 +3158,20 @@ sp_api::impl_runtime_apis! {
...
         }
 
         fn calculate_fee(command: Command, parameters: Option<PricingParameters<Balance>>) -> Fee<Balance> {
             snowbridge_pallet_outbound_queue::api::calculate_fee::<Runtime>(command, parameters)
         }
     }
 
+    impl snowbridge_outbound_queue_v2_runtime_api::OutboundQueueV2Api<Block, Balance> for Runtime {
+        fn prove_message(leaf_index: u64) -> Option<MerkleProof> {
+            OutboundMessageCommitmentRecorder::prove_message_v2(leaf_index)
```
tmpolaczyk, line 3154: Ok so the relayer uses this runtime api to get the merkle proof, therefore it doesn't care about the structure of the merkle tree. Nice!

pallets/outbound-message-commitment-recorder/src/lib.rs
ParthDesai: No

pallets/outbound-message-commitment-recorder/src/lib.rs
ParthDesai: The `take_commitment_root` is used by beefy extra data provider. https://github.com/moondance-labs/tanssi/blob/8cbeb0f8191f6d3de5f0e42a30ab69f5dbb7c79d/chains/orchestrator-relays/runtime/dancelight/src/lib.rs#L1449

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +51,106 @@ pub mod pallet {
...
+                    // Only v1, return v1 root directly no need to recompute root
+                    Pallet::<T>::deposit_event(Event::<T>::CommitmentRootRead { commitment: root });
+                    Some(root)
+                }
+                (None, Some((root, _leaves))) => {
+                    // Only v2, return v2 root directly no need to recompute root
+                    Pallet::<T>::deposit_event(Event::<T>::CommitmentRootRead { commitment: root });
+                    Some(root)
+                }
+                (None, None) => None,
```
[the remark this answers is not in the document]
ParthDesai: That is intentional. The commitment data is removed in next block. If a relayer want to get the proof, it needs to be calling the runtime api `prove_message` at a specific block in order to get the proof.

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +51,106 @@ pub mod pallet {
...
+                    let combined_leaves = Self::get_combined_leaves(
+                        v1_data.map(|(_root, leaves)| leaves),
+                        v2_data.map(|(_root, leaves)| leaves),
+                    );
+
+                    if let Some(combined_leaves) = combined_leaves {
+                        let proof = merkle_proof::<<T as Config>::Hashing, _>(
+                            combined_leaves.into_iter(),
+                            combined_leaf_index,
+                        );
```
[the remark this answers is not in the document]
ParthDesai: For overflow:
I have added saturating_add. 

Regarding panic when leaf index is out of bounds, this function is strictly used in runtime api, so panic is not a problem. In fact that is the behavior in master branch.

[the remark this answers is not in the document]
tmpolaczyk: Actually it was better to panic on overflow here, because this code is only used offchain and panics are easier to debug, but both AI tools and auditors will flag this code if we add a panic, so...

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +51,112 @@ pub mod pallet {
...
-                None
+            let v1 = RecordedCommitment::<T>::take();
+            let v2 = RecordedCommitmentV2::<T>::take();
+            match (v1, v2) {
+                (Some((_root_v1, leaves_v1)), Some((_root_v2, leaves_v2))) => {
+                    let combined_leaves =
+                        Self::get_combined_leaves(Some(leaves_v1), Some(leaves_v2));
+                    let root = merkle_root::<<T as Config>::Hashing, _>(
+                        combined_leaves.unwrap_or_default().into_iter(),
+                    );
```
tmpolaczyk, line 91: I don't like this because it computes a merkle root on-chain in on_finalize. What was the problem with the old approach of storing only the hash(root_v1, root_v2)?

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -25,14 +25,20 @@ pub use pallet::*;
...
-    use {frame_support::pallet_prelude::*, sp_core::H256};
+    use snowbridge_merkle_tree::{merkle_proof, MerkleProof};
+    use {
+        frame_support::pallet_prelude::*, snowbridge_merkle_tree::merkle_root, sp_core::H256,
+        sp_runtime::traits::Hash, sp_runtime::Vec,
+    };
 
     /// The current storage version.
-    const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
+    const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```
tmpolaczyk: Chopsticks upgrade tests are failing because we didn't add a migration. But I think we don't need a migration because the storage items are never read on-chain.

```suggestion
    const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```

Alternatively, if we want to keep the 1 here, we need to a migration that bumps the storage version to 1 and deletes `RecordedCommitment`.

pallets/outbound-message-commitment-recorder/src/lib.rs
tmpolaczyk: Asked @ParthDesai about this and the problem is the smart contracts, they expect a merkle tree created from a single `Vec<H256>`, they don't like unbalanced merkle trees. So here we merge the 2 separate `Vec<H256>` into 1.

I took a deeper look into the pallets, and it turns out we already calculate the merkle root in `snowbridge_pallet_outbound_queue` v1 and v2, also in `on_finalize`. These roots are the inputs to `record_commitment_root_v2`. So in the worst case we will be doing 4x the work we used to do in v1 (because we now hash in v1, v2, and again both combined here).

Also I looked at the on-chain values and the number of leaves is almost always 0 or 1 because it depends on the number of snowbridge messages sent in that block. So in the average case we won't be hashing, because we need at least 1 message sent from outbound v1 and 1 message sent from outbound v2 in the same block. It's only the worst case that is the problem.

Then my only question is whether we are benchmarking this somehow. I guess we don't. But I say it doesn't matter because the worst case will be 4x the weight of `commit()`, which is negligible:

https://github.com/moondance-labs/tanssi/blob/cbd101459706f304853a66762d6b65add6500c46/chains/orchestrator-relays/runtime/dancelight/src/weights/snowbridge_pallet_outbound_queue_v2.rs#L82

We don't care about proof size and `ref_time` is 31_028_000 picoseconds, which is 0 ms. So it's fine.

And also, we will remove this when we remove support for v1 outbound queue, so it is only a temporary hack.

tmpolaczyk (approved):

pallets/outbound-message-commitment-recorder/src/lib.rs
evgeny-s: I would better avoid adding a new migration if we don't need it. Let's keep the same version in this case.

pallets/outbound-message-commitment-recorder/src/lib.rs
```diff
@@ -45,25 +51,112 @@ pub mod pallet {
...
+                let maybe_combined_index = (leaves.len() as u64).checked_add(leaf_index);
+                if let Some(combined_index) = maybe_combined_index {
+                    Self::prove_message(combined_index)
+                } else {
+                    // Overflow
+                    None
+                }
+            } else {
+                Self::prove_message(leaf_index)
+            }
```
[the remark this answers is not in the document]
ParthDesai: If v1 exists we will always take combined index of v1's number of leaves plus the leaf index. How would it be possible to return v1 leave proof in that case?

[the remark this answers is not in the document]
tmpolaczyk: The fix seems reasonable, we can include it even if this never happens in practice.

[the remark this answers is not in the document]
ParthDesai: fixed. @coderabbitai and @tmpolaczyk

tmpolaczyk (approved):

